PR #91, opened 2012-10-31
Update safeDiff and safeSum

dakotacody:
Graphite usually treats 'None' values as missing data which results in breaks in the lines of the graph. However, the sum and diff functions will basically treat them as a 0 value which can make the graph inaccurate when adding or diffing 2 series. This update checks to make sure that there are at least 2 values to be summed or diffed and if not will not do the diff/sum. This falls more in line with the way that graphite handles 'None' values IMHO.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -46,12 +46,12 @@ INF = float('inf')
 #Utility functions
 def safeSum(values):
   safeValues = [v for v in values if v is not None]
-  if safeValues:
+  if safeValues and len(safeValues) > 1:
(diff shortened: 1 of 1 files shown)


dgryski:
I just ran into this problem with carbonapi.

obfuscurity:
I have to agree with @mleinart. I won't argue that @Dieterbe is wrong, only that the existing behavior is not unreasonable and is predictable for the majority of our users. Perhaps it would make sense to offer a configuration setting or render parameter to override this behavior, but I don't think we should be changing a foundational behavior such as this one.

PR #113, opened 2012-11-30
bar charts

piotr1212:
Hi, I've added graphType=bar to Graphite on request of my collegues. I thought it would be nice to share. Please let me know what you think,

Regards, Piotr


piotr1212:
I won't be finishing this, Grafana does this way better than I could ever built in Graphite.

PR #120, opened 2012-12-15
enable special character support

mingbowan:
allow user to send URLEncoded metric name to Graphite and display correctly, it won't break any existing data, but if user want some thing not allowed before, ex "server/name/test", if they URLEncode it, this change will allow Graphite to display/show metric name correctly.


grzegorz-dubicki:
+1

obfuscurity:
Closing this PR in favor of #1663. @mingbowan thanks for your contribution!

PR #150, opened 2013-02-10
When using functions like removeBelowValue, data series might be empty. ...

bryantsai:
When using functions like removeBelowValue, data series might be empty. Legends of such empty series are still shown, which is not so perfect especially when the number of legends are not small. Added check for such condition so such legends are ignored automatically.


obfuscurity:
Closing in favor of #1618.

PR #271, opened 2013-04-03
added a link to the header back to itself

linkslice:
yeah, it might be dumb. But I've gotten used to clicking on the 'logo' in the top left of pages to return to the home.  It's a quick n easy way to clear a composer window that you don't want anymore.


obfuscurity:
@linkslice It's not a bad idea at all, but unfortunately we already moved away from the text header to a background image with the new logo. If you can come up with a good way to link the logo (e.g. http://stackoverflow.com/a/25422362/2275058) I'd be happy to merge it in. Thanks!

PR #319, opened 2013-05-03
RFC: Add a history button to the dashboard kind of like Cacti

SEJeff:
This pull request should supercede #123.

@spellik: Please do look over [this commit](https://github.com/SEJeff/graphite-web/commit/1c0b97f3d16c7276b69d547545f6034d670cac75). I manually squashed your 3 commits in the previous pull request and removed all of the bonzai bits while keeping your name as the author so you get credit.

I then fixed a bug in that commit where graphs without title names would show up as `undefined 1 hour`, `undefined 1 day`, etc using a fancy javascript ternary operator. Since zoom support was removed, it didn't make much sense to keep the "Cacti" button, so I removed that as well and put the History button under "Graph Operations".

@nleskiw, @mleinart, and/or @obfuscurity please do give this a shot and merge if you like it. It certainly works well in master.

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -362,7 +362,7 @@ function initDashboard () {
     '<tpl for=".">',
       '<div class="graph-container">',
         '<div class="graph-overlay">',
-          '<img class="graph-img" src="{url}" width="{width}" height="{height}">',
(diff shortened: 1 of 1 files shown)


obfuscurity:
@SEJeff Are you still interested in seeing this go in?

cbowman0:
Changed applied to master in #2100.

PR #329, opened 2013-05-14
enable get all metrics in CLUSTER_SERVERS

sykp241095:
Enable:
"/metrics/index.json?cluster=True"
or
"/metrics/index.json?cluster=1"
styles to get all metrics in a cluster not only a single backend.

If "cluster=True" is provided, it will iterate all servers in this cluster to "GET /metrics/index.json", and merge all json response.

### DIFF
--- a/webapp/graphite/metrics/views.py
+++ b/webapp/graphite/metrics/views.py
@@ -31,28 +31,39 @@ except ImportError:
 
 def index_json(request):
   jsonp = request.REQUEST.get('jsonp', False)
+  cluster = request.REQUEST.get('cluster', False)
+
+  def find_matches():
+    matches = []
+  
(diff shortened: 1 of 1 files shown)


webapp/graphite/metrics/views.py
```diff
@@ -31,28 +31,39 @@
...
-
-  matches = [
-    m
-    .replace('.wsp', '')
-    .replace('.rrd', '')
-    .replace('/', '.')
-    .lstrip('.')
-    for m in sorted(matches)
-  ]
+  if cluster and len(settings.CLUSTER_SERVERS) > 1:
```
obfuscurity: Shouldn't this be `>= `? Technically there's no reason why you couldn't have a single `CLUSTER_SERVERS` on the backend.

PR #341, opened 2013-05-30
Fall back to cairocffi if pycairo is not installed

anotherthomas:
I think the dependency on pycairo makes it hard to pip-install graphite-web in a virtualenv. This patch adds a fallback option to use cairocffi (https://github.com/SimonSapin/cairocffi) if pycairo ist not installed.

### DIFF
--- a/check-dependencies.py
+++ b/check-dependencies.py
@@ -28,13 +28,16 @@ except ImportError:
     required += 1
 
 
-# Test for pycairo
+# Test for pycairo or cairocffi
 try:
   import cairo
 except ImportError:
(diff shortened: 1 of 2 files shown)


obfuscurity:
@SEJeff Any idea why we haven't switched to preferring cairocffi over pycairo? https://github.com/gescheit/graphite-web/blob/3eaba824df79eaf04cdd9f033bb063e01861efb8/webapp/graphite/render/glyph.py#L16-L19

SEJeff:
No preference from me. Which is easier to install from pip would be my first question. Generally I'd preference that onr

PR #372, opened 2013-07-20
support multiple WHISPER_DIRS

joemiller:
This adds support to graphite-web for searching multiple WHISPER_DIR's by renaming the WHISPER_DIR setting to WHISPER_DIRS and to a list from a string.

Includes backwards compatibility for WHISPER_DIR setting.

example:

```
WHISPER_DIRS = ['/opt/graphite/storage/whisper/01', '/opt/graphite/storage/whisper/02']
```

I'm experimenting with a graphite install where each carbon-cache instance is configured to store its whisper DB's in its own directory to facilitate easier cluster scaling. Supporting multiple WHISPER_DIRS makes it easier to move instances to new servers, for example: `rsync -a /opt/graphite/storage/whisper/01/ newserver:/opt/graphite/storage/whisper/01/`.

/tnx to @phobos182 for the suggestion here.

### DIFF
--- a/docs/config-local-settings.rst
+++ b/docs/config-local-settings.rst
@@ -84,7 +84,7 @@ CONF_DIR
 
 STORAGE_DIR
   `Default: GRAPHITE_ROOT/storage`
-  The base directory from which WHISPER_DIR, RRD_DIR, LOG_DIR, and INDEX_FILE default paths
+  The base directory from which WHISPER_DIRS, RRD_DIR, LOG_DIR, and INDEX_FILE default paths
   are referenced
 
 CONTENT_DIR
--- a/webapp/graphite/local_settings.py.example
+++ b/webapp/graphite/local_settings.py.example
@@ -65,7 +65,7 @@
 ## Data directories
 # NOTE: If any directory is unreadable in STANDARD_DIRS it will break metric browsing
 #CERES_DIR = '/opt/graphite/storage/ceres'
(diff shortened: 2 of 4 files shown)


obfuscurity:
@joemiller I like this idea but it should maintain backwards compatibility with `WHISPER_DIR`. Would you be willing to rebase and extend this PR to support that?

joemiller:
@obfuscurity Yeah I'll give it a shot. I figure my approach would be to retain the existing `WHISPER_DIR` variable but extend the parsing to accept a string or list of strings. Does that sound OK?

obfuscurity:
Almost. `local_settings.py.example` should introduce `WHISPER_DIRS` to replace/deprecate `WHISPER_DIR` but both should be supported in the parsing code. Thanks!

diceone:
Is there any update on this feature?

deniszh:
It should be rebased and altered to support backward compatibility with WHISPER_DIR.
@joemiller - do you still have an interest in this?

joemiller:
I don't have much time to work on this at the moment. Last I picked it up I
got bogged down in getting a working local dev environment up. If someone
is interested in picking up this work and taking it across the finish line
that would be awesome. I don't anticipate having time to do this on my own
time soon, and at $dayjob we're considering outsourcing our graphite so I'm
unlikely to have work-sponsored time either.

PR #402, opened 2013-08-22
Fix search tab in composer to match all (not any) entered patterns

ajablonski:
Currently, the search tab in the Composer treats a line in the index file as matching if it matches ANY of the patterns in the search box, but the help text suggests that the only lines that match are those that match ALL entered patterns.

This fix causes the search tool to demonstrate the behavior documented in the help text.


obfuscurity:
The search help was corrected in https://github.com/graphite-project/graphite-web/pull/1162 to match the actual (intended) behavior. Thanks for your submission though!

PR #420, opened 2013-09-03
Dashboard - Allow flot to draw the graphs

cbowman0:
Allow flot to draw the graphs on the dashboard.

This work was done a while back by Aleh Filipovich.

When applying it to the master branch it required an upgrade of flot as I was getting obscure errors when attempting to plot.  I went ahead and dropped the entire flot library in content/js/flot to make upgrades easier in the future.

We are open for suggestions on how to make this look and interoperate better.


cbowman0:
Welp, this is useless now.  Closing to clear the noise.

PR #424, opened 2013-09-06
Manipulate graph targets and implement New Graph -> New Empty Graph

cbowman0:
Buttons to re-order and remove graph targets on the Graph Target grid.

Add a new target to a graph.

Up and Down buttons are from http://tango.freedesktop.org/.  Delete button already existed in the source.

Add New Graph -> Empty Graph.


cbowman0:
For @obfuscurity: 
maniptargets

webapp/content/js/dashboard.js
```diff
@@ -1952,6 +2043,24 @@ function graphClicked(graphView, graphIndex, element, evt) {
...
+      width: buttonWidth,
+      handler: function() {
+                 // Hide the other menus
+                 operationsMenu.hide();
+                 optionsMenu.doHide(); // private method... yuck
+                 functionsMenu.hide();
+
+                 targetStore.add([ new targetStore.recordType({target: 'Edit to save'}) ]);
+                 targets.push('Edit to save');
+                 targetGrid.setHeight((rowHeight * Math.min(targets.length, maxRows)) + frameHeight);
```
obfuscurity: Does the indention look off here or is it my imagination?

cbowman0:
Looks like it's 4 space instead of 2 space.  I'll correct it.

cbowman0:
Can I get a peer review of this?

obfuscurity:
I'll try to take a look at the code later tonight. Initial UX thoughts: the navigation elements (green arrows) are different than the approach taken in the Composer (e.g. Move > Move Up). I think I like your approach a little better anyways, so NBD. But we probably shouldn't allow illegal choices (e.g. arrow up for the top element) unless that implies a wraparound action to the list.

obfuscurity:
@cbowman0 Sorry for the delay, looking into this now.

obfuscurity:
@cbowman0 Played around with it a bit. I think the feature would be much more useful if the new target field supported autocompletion. Without it, the feature feels not altogether fleshed out imho.

Is this something you'd be willing to take a stab at?

cbowman0:
Do you mean in the pop-up after clicking a graph, choosing new target and it adds a blank target line?

obfuscurity:
Correct.

cbowman0:
I'm not sure where to start with adding auto-completion to the add new target box.  Do others believe this is required to merge this in?

cbowman0:
@DanCech or @deniszh, any objections to merging this as is?

DanCech:
It's definitely an improvement, also not sure where to start with auto-complete as I haven't done any dev on the graphite-web ui.  I'd suggest opening an issue for adding auto-complete to it, then going ahead with the merge.

cbowman0:
#2097 created.  Merging now.

PR #431, opened 2013-09-17
Add a refresh button to Graphite Tree

chrishenry:
While developing new metrics, it's often really difficult for a developer to tell if their code is successful, particularly when working with Statsd. Instead of refreshing the entire composer view, a dev can click a button to refresh what's loaded in the tree, or refresh the last selected node of the tree.

Additionally, some javascript has been refactored. The code to recursively expand metrics has been pulled out of the composer.html view.


obfuscurity:
@chrishenry Thanks for your contribution. I've moved this over to #1652 to resolve conflicts.

PR #447, opened 2013-10-02
add function divideEachSeries

aaron2:
takes two series lists of equal length and divides each series in the first list by its corresponding series in the second. useful for calculating multiple or variable percentages


obfuscurity:
@SEJeff @deniszh You guys mind reviewing this one?

obfuscurity:
I'm wondering if there's still value in this change considering that https://github.com/graphite-project/graphite-web/pull/1567 added support for multiple series in `asPercent`. Thoughts? /cc @cbowman0

obfuscurity:
Pretty sure this change is no longer needed. Closing it out but please let me know if I've overlooked something and we can reopen.

PR #465, opened 2013-10-11
Implement a config file for the render engine

cbowman0:
First two settings are:

1. Remove most recent datapoint in aggregate functions if there are any NULLs.

2. Sort the result set from fetchData so colors per series are consistent between graphs.

### DIFF
--- a/conf/render.conf.example
+++ b/conf/render.conf.example
@@ -1,0 +1,9 @@
+# This configuration file controls the behavior of the render function.  These
+# are settings that will be applied to all render calls.
+#
+# This file must contain a [render] section that defines values for all of the
+# following settings.
+
+[render]
+trim_recent = 1
+sort_series = 1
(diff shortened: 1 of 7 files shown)


cbowman0:
Rebased against master.

obfuscurity:
Why do we need a new config file for graphite-web instead of just adding new settings to `local-settings.py` and `settings.py`?

cbowman0:
It made sense to me 1.5 years ago when I wrote it up.  I'm attempting to remember that logic.  From what I recall, there aren't any similar items in settings.py or local-settings.py that manipulate content, so I mimicked graphTemplates.conf and dashboard.conf.  Whether that made sense then or makes the most sense now, I'll let you guys tell me.  If the consensus is we put them in settings.py, then that's a-ok with me.

obfuscurity:
Unless there's a compelling reason for the new file, I think we should probably stick with the existing configs. Closing for now but if you can think of the original motivation please let us know.

PR #493, opened 2013-11-12
add verticalLine() url function

nharkins:
hi. i've written a verticalLine() function for graphite's url api
which draws a vertical line on the graph at a specific
timestamp specified by the first argument function, 
in the same time formats as &from= and &until=, e.g.

&target=verticalLine("12:3420131108","event","blue")

this will be handy for quickly marking ad-hoc events 
on graphs without having to think of where it belongs
in the namespace and cluttering the database.

it seems like at least one other person wanted this functionality:
https://answers.launchpad.net/graphite/+question/187677

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2271,7 +2271,7 @@ def verticalLine(requestContext, ts, label=None, color=None):
 
   .. code-block:: none
 
-    &target=verticalLine(12:3420131108,'event','blue')
+    &target=verticalLine("12:3420131108","event","blue")
 
   """
   ts = timestamp( parseATTime(ts, requestContext['tzinfo']) )
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -22,7 +22,7 @@ import random
 import time
 
 from graphite.logger import log
-from graphite.render.attime import parseTimeOffset
+from graphite.render.attime import parseTimeOffset, parseATTime
 
 from graphite.events import models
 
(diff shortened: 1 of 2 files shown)


softwareklinic:
&target=verticalLine("12:3420131108","event","blue")

How did you create this number - and what does it represent   "12:3420131108"

nharkins:
it is a timestamp, e.g. any value valid to the &from= arg
as per http://graphite.readthedocs.org/en/latest/render_api.html
"ABSOLUTE_TIME is in the format HH:MM_YYMMDD, YYYYMMDD, MM/DD/YY, or any other at(1)-compatible time format."
it seems that now an underscore is often used: HH:MI_YYYYMODD
but i seem to recall that was without it in the docs a while back, and
the old style still seems to work in my instance (backwards compat?)

i found that on the docs somewhere. looks like 
HH::MI_YYYYMODD is preferred now.

softwareklinic:
Thank you.



Regards,
Keyur

On Oct 20, 2015, at 5:39 PM, Neil Harkins <notifications@github.com<mailto:notifications@github.com>> wrote:


it is a timestamp, e.g. any value valid to the &from= arg
as per http://graphite.readthedocs.org/en/latest/render_api.html
"ABSOLUTE_TIME is in the format HH:MM_YYMMDD, YYYYMMDD, MM/DD/YY, or any other at(1)-compatible time format."
it seems that now an underscore is often used: HH:MI_YYYYMODD
but i seem to recall that was without it in the docs a while back, and
the old style still seems to work in my instance (backwards compat?)

i found that on the docs somewhere. looks like
HH::MI_YYYYMODD is preferred now.

obfuscurity:
@nharkins Can you rebase this so we can merge it in cleanly?

obfuscurity:
Closing in favor of #1631.

Issue #217, opened 2013-03-13
timeShift() nonexistent target raises IndexError


mleinart:
#### Imported from https://bugs.launchpad.net/bugs/1037178

|  |  |
| --- | --- |
| Reported by | [jpellerin](https://launchpad.net/~jpellerin) |
| Date Created | Aug 15, 2012 |

The target ending in 'fail' here doesn't exist. The graph renders correctly when requested without timeshift() -- no line at all for the target that doesn't exist. I'd expect the same when timeShift()'ed.

Environment:

Request Method: GET
Request URL: http://gw-qalog:8080/render/?width=586&height=308&_salt=1345045206.006&target=timeShift(stats_counts.celery.magnet.sealab.tasks.sync_campaigns.ok%2C%221h%22)&target=timeShift(stats_counts.celery.magnet.sealab.tasks.sync_campaigns.fail%2C%221h%22)

Django Version: 1.3.1
Python Version: 2.6.6
Installed Applications:
['graphite.metrics',
 'graphite.render',
 'graphite.cli',
 'graphite.browser',
 'graphite.composer',
 'graphite.account',
 'graphite.dashboard',
 'graphite.whitelist',
 'graphite.events',
 'django.contrib.auth',
 'django.contrib.sessions',
 'django.contrib.admin',
 'django.contrib.contenttypes',
 'tagging']
Installed Middleware:
('django.middleware.common.CommonMiddleware',
 'django.middleware.gzip.GZipMiddleware',
 'django.contrib.sessions.middleware.SessionMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware')

Traceback:
File "/usr/lib/python2.6/site-packages/django/core/handlers/base.py" in get_response
1.                         response = callback(request, _callback_args, *_callback_kwargs)
   File "/opt/graphite/webapp/graphite/render/views.py" in renderView
2.         seriesList = evaluateTarget(requestContext, target)
   File "/opt/graphite/webapp/graphite/render/evaluator.py" in evaluateTarget
3.   result = evaluateTokens(requestContext, tokens)
   File "/opt/graphite/webapp/graphite/render/evaluator.py" in evaluateTokens
4.     return evaluateTokens(requestContext, tokens.expression)
   File "/opt/graphite/webapp/graphite/render/evaluator.py" in evaluateTokens
5.     return func(requestContext, *args)
   File "/opt/graphite/webapp/graphite/render/functions.py" in timeShift
[...]

PR #505, opened 2013-11-22
Make sure timeShift(...) doesn't crash for non-existing series

zaptm:
Fixes https://github.com/graphite-project/graphite-web/issues/217

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2225,17 +2225,18 @@ def timeShift(requestContext, seriesList, timeShift, resetEnd=True):
   myContext = requestContext.copy()
   myContext['startTime'] = requestContext['startTime'] + delta
   myContext['endTime'] = requestContext['endTime'] + delta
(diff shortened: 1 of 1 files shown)


duclet:
So I see this is merged into master all the way from 2013 but it was not merged in the 0.9.x branch which seems like the releases are tagged from. As such, this is still not working on the current version.

deniszh:
I can backported it to 0.9.x, but please note that we have no plans to release 0.9.x anymore.

duclet:
If you are not going to release it anymore, then it is fine. I've been trying to setup a graph that needs this feature and was surprised that a fix was merged into master in 2013 and with the current latest version didn't have the fix (don't see any other version besides the 0.9.x). If there will be a new version based on master soon, then I can wait. Thanks.

jkandula:
Has this been part of any release yet? Thanks

deniszh:
@jkandula: No, not yet. Made it a part of 0.9.16 milestone.
It's in the master branch, so, will be part of 1.0 release too - when it is done.

PR #551, opened 2013-12-23
Make asPercent/divideSeries accept multiple series as total/divisor

### LINKED ISSUE #207: [feature request] asPercent taking two series lists
#### Imported from https://bugs.launchpad.net/bugs/1005366
|||
|----|----|
|Reported by|[harisuthan](https://launchpad.net/~harisuthan)|
|Date Created|May 28, 2012|
Originial Desc:
I have a use case where I need to find the percentage between two series lists like:

asPercent(groupByNode(serieslist1, x, "sumSeries"), groupByNode(serieslist2, y, "sumSeries"))

Is there way to achieve this? If not and if you find this useful may I request a feature addition where, when two series lists with "equal length" are passed as parameters to asPercent, it will return a serieslist with the same length where each series is a percentage of the individual input series?

Michael's Response:
I think asPercent and divideSeries could probably support this - rather
than the 2nd series being required to have a length of 1, it could be
required to have either: a length of 1 (current behavior) or a length
equal to the 1st series (your proposed behavior).

Unfortunately I cant spot a way to do this without having to enumerate
every series out.

If you'd like the new feature to be explored, convert this question to a
bug report and mark it as a feature request so that it can be tracked.

leandro-lucarella-sociomantic:
If multiple series as used as total for asPercent, the length of the
total series must match the length of the seriesList, and they will be
matched together to calculate the percentage of the individual values.

Also remove the current limitation of divisorSeriesList to have only one
series and accept multiple series. In this case the number of divisor
series must match the number of dividend series, and they are matched so
each dividend series get divided by the corresponding divisor.

The old behaviour when only one series is used is retained, so this is
only an extension.

Fixes #207.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -17,7 +17,7 @@ from numpy import median
 
 from datetime import date, datetime, timedelta
 from functools import partial
-from itertools import izip, imap
+from itertools import izip, imap, repeat
 import math
 import re
 import random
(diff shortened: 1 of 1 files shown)


obfuscurity:
Pretty sure this change is no longer needed. Closing it out but please let me know if I've overlooked something and we can reopen.

PR #565, opened 2014-01-07
Add now=None to branch

bmhatfield:


bmhatfield:
I don't remember if this is still valid. I think this still fixes a bug though. Looking through #523, this appears to be still useful. Thoughts?

obfuscurity:
Fixed in https://github.com/graphite-project/graphite-web/commit/a88ba855139b8c90dfffdbbefc7f3aa83bcda427.

PR #622, opened 2014-02-12
Implementation called django-admin.py

dongweiming:
In `bin/run-graphite-devel-server.py`, cost many code to find `django-admin.py`, I think can direct use django mthod : `management.execute_from_command_line(command)` , Once you've installed django, django-admin.py script does not necessarily need to also be executed

### DIFF
--- a/bin/run-graphite-devel-server.py
+++ b/bin/run-graphite-devel-server.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 
 import sys, os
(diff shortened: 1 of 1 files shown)


iksaif:
Looks good to me, could you rebase it ?

dongweiming:
@iksaif time long past. this repository had deleted. i submit a new pull request https://github.com/graphite-project/graphite-web/pull/1763

PR #630, opened 2014-02-20
GraphiteLogger methods, LeafNode() optional param for DB reader use; exception logging.

justanyone:
Three changes:
- Adding LeafNode() optional param avoidIntervals, defaults off:  enables reader hitting a database.
- Adding GraphiteLogger() methods corresponding to standard Python logging methods.  
- Added a try/except block around IndexSearcher.reload() to log exception.  

Explanations:
- When using a custom finder/reader of a database (in our case, MongoDB), querying to get_intervals() was just as expensive as reading the data itself.  Note that Nodes are instantiated during graphite tree expansion.  This was seriously wasteful since we'd have to do the same exact query again during the reader() call.  Change was adding an optional param 'avoidIntervals' to delay the get_intervals() call until the actual fetch() and eliminate the extra useless data retrieval from the DB.  **Default behavior is unchanged**.
- When adding a custom finder/reader that incorporates classes also used in ancillary processes, logging is necessary for those classes, yet could not use the standard Python logging methods of (debug/info/warn/error/critical/exception).  Custom finder/reader should use Graphite logging, yet GraphiteLogger did not have those methods.  Added these methods.
- During an exception in a custom finder/reader, Graphite fails invisibly with no traceback or clue where/why.  Added a try/except block around IndexSearcher.reload() helps this by catching and logging the exception.  

Added Test coverage:
- Added tests for GraphiteLogger in new test class test_logger.py, exercising both old and new methods.
- Added tests for BranchNode and LeafNode that provide 100% coverage and many asserts.
- All pre-existing tests also pass.


obfuscurity:
@justanyone Sincere apologies for this PR languishing so long. While writing the Graphite book I became acutely aware of how much love Graphite's logging subsystem needs, so this patch resonates deeply with me. However, it looks like it's going to need a fair amount of rebasing work to get it in a clean state. I'll try to take a stab at it as my time permits but for now I'm going to add it to the post-1.0 milestone.

PR #636, opened 2014-02-22
Fix #635 -- event tags creation and serialization

### LINKED ISSUE #635: Event doesn't accept multiple tags
It looks like /events can accept an array of tags, even the "tags" plurality seems to imply that... however it doesn't appear to be the case
```json
[
{
data: "deploy of master branch happened at Fri Jan 3 22:34:41 UTC 2014",
what: "Event - deploy",
when: 1393015951,
id: 1,
tags: "deploy"
},
{
data: "deploy of master branch happened at Fri Jan 3 22:34:41 UTC 2014",
what: "Event - deploy",
when: 1393015985,
id: 2,
tags: "[u'deploy', u'foo']"
},
{
data: "deploy of master branch happened at Fri Jan 3 22:34:41 UTC 2014",
what: "Event - deploy",
when: 1393016580,
id: 3,
tags: "[u'OutRouting', u'Patched']"
}
]
```

i would expect  _https://graphite.xxxxx/events/get_data?tags=OutRouting_ to return something, but that does not appear to be the case.  

am I misunderstanding something or is this a bug ? 

-John

brutasse:
Tags were created properly with `"tags": "space-separated tags"` but not `"tags": ["actual", "array"]`. And the serialization was always wrong, returning tags as a string.

Now there are arrays everywhere, which makes more sense given "tags" is plural.

### DIFF
--- a/docs/releases/0_10_0.rst
+++ b/docs/releases/0_10_0.rst
@@ -35,6 +35,10 @@ Changes
   has been rewritten in Python. ``build-index.sh`` is still available but is
   just an alias to ``build-index``.
 
+* The events API now requires ``tags`` to be an array when creating tagged
+  events. Previous versions only accepted string attributes. Tags are also
+  serialized as arrays.
+
 New functions
 -------------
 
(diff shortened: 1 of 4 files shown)


tuxinaut:
:+1:

obfuscurity:
Deprecated by #1627.

Issue #625, opened 2014-02-18
Is constantLine function broken in 0.9.12?


pythiannunez:
I'm trying curl "http://localhost/render/?target=constantLine(21)&format=json" and I'm getting "TypeError: range() integer end argument expected, got float."

I'm pretty sure I used that function before, but maybe in another version.

Full error here https://gist.github.com/pythiannunez/9bb7e4e3b29d9208ae64

imkin:
Has fix been merged?

obfuscurity:
I can confirm that this is working fine in master.

```
$ curl -k "https://127.0.0.1:8443/render/?target=constantLine(21)&format=json"
[{"target": "21", "datapoints": [[21, 1469455685], [21, 1469498885], [21, 1469542085]]}
```

PR #641, opened 2014-02-28
Fix for https://github.com/graphite-project/graphite-web/issues/625.

pauled23:
Confirmed to render line correctly.


obfuscurity:
Confirmed, this was fixed by @g76r in #645.

PR #649, opened 2014-03-02
Add upperBound and lowerBound render target functions. Add time series support to 'offset'.

bluecmd:
These functions are the equivalent of min(x, C) and max(x, C).
This allows the user to clip the data to fit other operations.

A real world example would be if you want to see if a switch
interface is up or not, but don't care why it's down.
Since ifOperStatus is >= 2.0 if there is an error:
sum(offset(upperBound(switch.ifOperStatus.*,2.0),-1))

===

Add support for adding and substracting time series by using offset.
This allows for graphing of timestamp based events where you might
want to find out how long ago something happened.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -715,22 +715,33 @@ def absolute(requestContext, seriesList):
 
 def offset(requestContext, seriesList, factor):
   """
-  Takes one metric or a wildcard seriesList followed by a constant, and adds the constant to
-  each datapoint.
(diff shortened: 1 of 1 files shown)


obfuscurity:
I looked at this a little closer and each of these "features" can easily be handled by existing functions imho. The change to `offset()` just makes it act like `sumSeries()` afaict. The `upperBound` and `lowerBound` functions act like our data filters `removeAboveValue` and `removeBelowValue`.

Closing this for now but if I've overlooked something please let me know and we can reconsider.

Issue #671, opened 2014-03-27
How can I use CactiStyle and/or get human readable Legend values ?


toni-moreno:
use ususaly a CactiStilye Legend but absolute values are to large to easily read values
( by example on current  memory-used i have   6012440576.00 bytes ) 

![image](https://cloud.githubusercontent.com/assets/5883405/2539217/fb249904-b5c5-11e3-846b-dd39aaf74948.png)

Any easy way to see  " 5,59 Gb or 5733 Mb  " instead of  "6012440576.00" values?

PR #682, opened 2014-04-09
added units to cactistyle function

toni-moreno:
This pull request is made in response on the issue 

https://github.com/graphite-project/graphite-web/issues/671

I've implemented an easy way to indicate units and perhaps we will improve a lot the look&Feel by adding a "unit" parameter to the cactiStile() function to avoid a "M/G/K" or "Mi/Gi/Ki" that has no sense without units.

So this graph

```
cactiStyle( "some*memory*metric","binary") 
```
image

Can now be as follow by adding a "b" as unit parameter.

```
cactiStyle( "some*memory*metric","binary", "b" ) 
```

image

or 

```
cactiStyle( "some*memory*metric","binary", "bytes" ) 
```

image

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -1115,7 +1115,7 @@ def alias(requestContext, seriesList, newName):
       series.name = newName
   return seriesList
 
-def cactiStyle(requestContext, seriesList, system=None):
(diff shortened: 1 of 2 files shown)


obfuscurity:
Deprecated by #1629.

Issue #681, opened 2014-04-07
Dashboard authentication is not working ?


toni-moreno:
I'm trying to improve access control by uncommenting dashboard auth settings in /opt/graphite/webapp/graphite/local_settings.py

```
###############################
# Authorization for Dashboard #
###############################
DASHBOARD_REQUIRE_AUTHENTICATION = True
DASHBOARD_REQUIRE_EDIT_GROUP = 'dashboard-editors-group'
DASHBOARD_REQUIRE_PERMISSIONS = True
```

I'm working with the embedded authentication ( using mysql as de database backend), and any authentication related log is written on the /opt/graphite/storage/log/webapp/*.log files.

What is what I'm doing wrong ?

changbl:
I am seeing the same issue: Graphite-web dashboard authentication simply does NOT work. Every user which can access the dashboard via a web browser can modify/save dashboard without the need to login.

OS: Ubuntu 14.04
Graphite-web: 0.9.12

galdreiman:
Is there any solution for this issue?
I really like Graphite and want to use it, but I cannot allow access to my systems data without authentication.

deniszh:
You can use some external dashboard with authorization (e.g. Grafana http://grafana.org ) and limit Graphite access only from that Grafana host (localhost, if installing Grafana on the same host as Graphite)

PR #686, opened 2014-04-14
added GENERIC_LOGIN_REQUIRED setting for a full authenticated graphite site

toni-moreno:
I've added a new setting to enable a full authenticated site, related to issue  https://github.com/graphite-project/graphite-web/issues/681

Is only tested with internal django user handler, not tested yet with remote or ldap authentication.

Also I've added a new look&feel for the login form: 

The old login form was:

image

And the new one:

image


changbl:
No one bother to review this commit and merge it?

toni-moreno:
Yes !!! It would be great !!!!

tony-dillon:
Hey Toni-Moreno,

Do you think you could fix the merge conflicts?  I would love to get this merged in!

Thank you for your efforts.

webapp/graphite/settings.py
```diff
@@ -200,6 +204,14 @@
 if USE_LDAP_AUTH and LDAP_URI is None:
   LDAP_URI = "ldap://%s:%d/" % (LDAP_SERVER, LDAP_PORT)
 
+LOGIN_EXEMPT_URLS = (
+ r'^render',
+ r'^metrics/'
```
micolous: I think this means that data can still be accessed, just the standard web interface is unavailable.  Is there a way to protect these endpoints too?

PR #692, opened 2014-04-18
adding groupByNodes() function, same as groupByNode() w/ several nodes

g76r:
This make it possible to draw such targets:
groupByNodes(metrics.server*.database.*.*.*,'sum',1,4,5)
Whereas groupByNode can only group by one node.

Actualy, it was already possible to do that with groupByNode, but this required to use such ugly patterns:
aliasSub(groupByNode(aliasSub(aliasByNode(metrics.server*.database.*.*.*,1,4,5),'\\.','_'),0,'sum'),'_','.')

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2596,9 +2596,12 @@ def groupByNodes(requestContext, seriesList, callback, *nodes):
     if key not in metaSeries.keys():
       keys.append(key)
       metaSeries[key] = [series]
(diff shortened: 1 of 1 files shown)


g76r:
I would be pleased if this PR was rejected in favour of #1171 (as long as #1171 is merged, of course :-)).
Otherwise I would be ok to rebase this one.

g76r:
no longer needed since #1551 has been merged

PR #698, opened 2014-04-29
Move code from metrics/views.py findmatches into the respective finders

ggs-sglimm:
We use a custom finder. We noticed, that /metrics/index.json was only returning [].
Investigation showed, that the index view hard-coded only whisper, rrd and ceres backends.
This change moves this responsibility into the finder.

### DIFF
--- a/webapp/graphite/finders/ceres.py
+++ b/webapp/graphite/finders/ceres.py
@@ -9,6 +9,7 @@ from graphite.node import BranchNode, LeafNode
 from graphite.readers import CeresReader
 
 from . import get_real_metric_path
+import fnmatch
 
 
 class CeresFinder:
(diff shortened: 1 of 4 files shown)


DanCech:
This idea dovetails well with the work in #2093 so I went ahead and made a branch that follows on from those changes.

I'll open a PR once #2093 has been merged and I've had a chance to add some tests, but in the meantime the diff is at https://github.com/DanCech/graphite-web/compare/backend-failures...DanCech:metrics-index?expand=1

I changed the name of the function to `get_index` (since it doesn't return nodes) and added a base implementation so it'll work out of the box with any existing finders (but finder authors can override `get_index` with a more efficient implementation if desired).

The other change I did make was to remove the old `cluster` parameter to `/metrics/index.json` and replace it with the same `local` behavior as the other find & render endpoints (as well as adding support for `pretty`).  Supporting `cluster` is possible, but would require some messy code in `STORE.get_index`, it could be done if we think it's important to maintain strict BC though it's much cleaner and more consistent to use `local` IMHO.

DanCech:
Closing in favor of #2103

PR #717, opened 2014-05-15
bugfix in perSecond():  if the series has null data, diff the existing d...

bruce-lyft:
...ata.

        Consider a series that looks like this:
            X1,null,X2,null,X3,null,...
        In the current implmenentation, perSecond() will return all nulls
        for this series.  With this bugfix, perSecond will return instead:
            (x2-x1)/step,null,(x3-x2)/step,null,...

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -927,9 +927,12 @@ def perSecond(requestContext, seriesList, maxValue=None):
     prev = None
     for val in series:
       step = series.step
-      if None in (prev,val):
+      if prev == None:
         newValues.append(None)
         prev = val
+        continue
+      if val == None:
+        newValues.append(None)
         continue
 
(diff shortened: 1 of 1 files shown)


cbowman0:
I'm not sure about this one.

If the series is defined with a 60s step and the following data  `[0, 60, 120, 180, 240, 300, 360, 420, 480, 540]`

Then the perSecond is: `[None, 1, 1, 1, 1, 1, 1, 1, 1, 1]`

If the data is defined: `[0, 60, None, 180, None, 300, None, 420, None, 540]`

Then the perSecond is: `[None, 1, None, 2, None, 2, None, 2, None, 2]` ?

Is the intention is to end up with: `[None, 1, None, 1, None, 1, None, 1, None, 1]` ?

bruce-lyft:
Thanks!

PR #718, opened 2014-05-17
added get-node function to get all existing values in a node position

toni-moreno:
I've added a new function usefull to get a list of existing values under a selected node position given a metric pattern, i json format

__parameters:__
* query: metric pattern 
* node: portion/s of the target name. Node indices are 0 indexed.

example:

`http://graphite_url/metrics/get-node/?query=*.bbdds.sys*.system.*&node=2`

give us

`{"nodes": [{"name": "system00"}, {"name": "system01"}]}`

this will be useful for external dashboards  to build dinamically combo boxes  to select a substring for a complete metric name.

### DIFF
--- a/webapp/graphite/metrics/urls.py
+++ b/webapp/graphite/metrics/urls.py
@@ -25,5 +25,6 @@ urlpatterns = patterns(
         name='metrics_get_metadata'),
     url('^set-metadata/?$', views.set_metadata_view,
         name='metrics_set_metadata'),
+    url('^get-node/?$', views.get_nodelist_data,name='metrics_get_node'),
     url('', views.find_view, name='metrics'),
 )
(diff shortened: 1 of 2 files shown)


toni-moreno:
Hi  guys. 

Any Plans to merge this PR to current master , please???

I need it to get node info to our dashboards. And will be usefull also for any other dasboard frontend.

Thank you very much !!

brutasse:
Like @obfuscurity, I still think this change is a bit too narrow and we should rather improve other node-discovery APIs than add other very specialized API calls.

toni-moreno:
Hi @obfuscurity @brutasse .

Do you prefer if I can modify  the "expand" API to get this data?

something like:

```
http://graphite/metrics/expand?query=webservers.production.*.exim.*&groupByNode=2
```

Is there any other API where better to place this new functionality?

torkelo:
I think this feature sounds very useful, but feel that it belong to the `/metric/find` api is just another way of finding metric paths / nodes.

obfuscurity:
@torkelo see #1630

PR #733, opened 2014-05-30
Fixed  jsonp support for metrics view

hamaxx:
Jsnop support existed for index_json but as far as I could tell it didn't work.
This request fixes a string formatting issue in json_response_for and ads jsonp support to other views.

I also cleaned up import and fixed urlopen call since it wasn't imported at all.


webapp/graphite/metrics/views.py
```diff
@@ -11,7 +11,9 @@
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License."""
-from urllib2 import urlopen
+import fnmatch
+import os
+import urllib2
```
obfuscurity: Why import all of `urllib2` when we're still only using `urllib2.urlopen`?

obfuscurity: Note that `from urllib2 import urlopen` works fine for me. I've used `index.json` extensively in the past and it's never been an issue afaik.

obfuscurity:
Closing in favor of https://github.com/graphite-project/graphite-web/pull/1655.

PR #737, opened 2014-06-05
Add time-ms yUnitSystem parameter for displaying millisecond values

markolson:
Shows values in milliseconds, scaling to seconds, minutes, hours, days and years.


markolson:
#1220 took care of this

PR #748, opened 2014-06-24
Provide all digits for float numbers when using the raw format of render

heliodor:
Float numbers are currently truncated in responses from the ‘render’
url if using the ‘raw’ format. Using the json format prints out all
necessary digits for floats, so it would be nice for 'raw' to do the correct thing too.


obfuscurity:
@heliodor or @cbowman0 Would you be willing to add some tests to this before we consider merging?

cbowman0:
We don't have tests for raw output?  Bummer.  I took a stab at it in #1624.

obfuscurity:
Thanks @heliodor and @cbowman0, closing this one.

PR #771, opened 2014-07-12
improve rendering performance dramatically in certain situations

### LINKED ISSUE #770: graph rendering slows down exponentially with number of data-points -> O(n²)
I'm having a metric collected every second and being stored with following aggregation:

``` [everything_1s7d_5min3mon]
pattern = .*\.1s\..*
retentions = 1:604800,300:26784
```

I noted that rendering graphs for e.g 7days is extremely slow (~49s) and when I changed the time-span between 1day and 7days i noted that the rendering time is exponentially increasing.

When selecting e.g 2weeks, it's ways faster, because the aggregated data is used.

First, I'm wondering why graphite is using the 1s resolution data for rendering a 500px wide graph, because would be much more reasonable to use the aggregated data therefore, but that's not the issue for me now. 

My bigger problem is the "exponential slowness".

I've put a lot of timing-logging into the rendering(glyph.py) stuff and it turned out that every iteration over the timeseries itself is causing the problem. I think the reason therefore is in the `TimeSeries` Class in `datalib.py`:

``` python
def __consolidatingGenerator(self, gen):
    buf = []
    for x in gen:                             # <<--- outer loop
      buf.append(x)
      if len(buf) == self.valuesPerPoint:
        while None in buf: buf.remove(None)
        if buf:
          yield self.__consolidate(buf)
          buf = []
        else:
          yield None
    while None in buf: buf.remove(None)
    if buf: yield self.__consolidate(buf)
    else: yield None
    raise StopIteration


  def __consolidate(self, values):
    usable = [v for v in values if v is not None]  # <<--- inner loop
    if not usable: return None
    if self.consolidationFunc == 'sum':
      return sum(usable)                           # <<--- inner loop
    if self.consolidationFunc == 'average':
      return float(sum(usable)) / len(usable)      # <<--- inner loop
    raise Exception, "Invalid consolidation function!"

```

For my case I had a `valuesPerPoint` of >1000 and there is a lot of potential to optimise in this two methods (in terms of performance, eventually loosing elegant/clean code)

powo:
should resolve #770

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -42,35 +42,35 @@ class TimeSeries(list):
     self.valuesPerPoint = int(valuesPerPoint)
 
 
+  __consolidation_functions = {
+    'sum': sum,
+    'average': lambda usable: float(sum(usable) / len(usable)),
+    'max': max,
+    'min': min,
+  }
   def __consolidatingGenerator(self, gen):
-    buf = []
+    try:
+      cf = self.__consolidation_functions[self.consolidationFunc]
+    except KeyError:
+      raise Exception, "Invalid consolidation function!"
+    buf = []    # only the not-None values
+    valcnt = 0
     for x in gen:
-      buf.append(x)
-      if len(buf) == self.valuesPerPoint:
-        while None in buf: buf.remove(None)
+      valcnt += 1
+      if x is not None:
+        buf.append(x)
+      if valcnt == self.valuesPerPoint:
+        valcnt = 0
         if buf:
-          yield self.__consolidate(buf)
+          yield cf(buf)
           buf = []
         else:
           yield None
-    while None in buf: buf.remove(None)
-    if buf: yield self.__consolidate(buf)
-    else: yield None
+    if buf:
+      yield cf(buf)
+    else:
+      yield None
     raise StopIteration
-
-
-  def __consolidate(self, values):
-    usable = [v for v in values if v is not None]
-    if not usable: return None
-    if self.consolidationFunc == 'sum':
-      return sum(usable)
-    if self.consolidationFunc == 'average':
-      return float(sum(usable)) / len(usable)
-    if self.consolidationFunc == 'max':
-      return max(usable)
-    if self.consolidationFunc == 'min':
-      return min(usable)
-    raise Exception("Invalid consolidation function!")
 
 
   def __repr__(self):


mbethke:
Whow, this is such a massive improvement I'm surprised it hasn't made it in a release in over three years! We have a large amount of metrics that would be a headache to configure manually WRT resolution so we'd rather waste a bit of cheap disk space and use a fine resolution for everything if it didn't result in several minutes of CPU time for a handful of graphs.
I'd be happy to contribute any unit tests that might be missing if that would help get this into v1.0.3, alas I can't figure out in what way the existing ones aren't adequate. Any ideas, @esc and @drawks?

@steve-dave You're right, the first line of the `__consolidate` function is redundant. When I ran into this problem I thought about dropping the whole `buf` logic and using only this list comprehension but in the end @powo's optimization is probably more efficient as it does less copying.

deniszh:
@mbethke : First, it should be rebased. Did you really try that change? Is it really a big improvement?
Second - that's a good question about tests in this particular case. It's a change of TimeSeries class, main core thing of whole Graphite. Do we really need a unit test for such case? For what then - for a private function? 
Thoughts? /cc @DanCech @iksaif @brutasse ?

DanCech:
The patch itself looks reasonable, adding unit tests is definitely a good idea, the suggested patch for checking valcnt makes sense to me.  I'll check it out.

mbethke:
@deniszh It's such a big speedup that I believed it to be a fluke at first. One of our Whisper files takes just over a minute to condense down to 10 JSON data points on reasonably beefy hardware (2.4.GHz E5-2630, 128 GB), most of which is spent pointlessly manipulating lists of `None` in `__consolidatingGenerator` (as I think @powo has pointed out already, the `remove` operation is O(n) over the length of the list, and with most entries being `None` this is basically done n times for an effective O(n²)).
With the patch, this time is down to a little over two seconds!

iksaif:
I think I would be fine with that as long as it's rebased, tested, and you provide a graph showing the improvements.

DanCech:
I just made my own rebased branch and it seems to work ok, test coverage actually isn't bad but the tests will need to be updated to be able to fix the issue with the extra None values.  I'll dig into it more tomorrow.

mbethke:
As I couldn't rebase this PR myself, I worked on my own repo and just opened  #2051.
I didn't see any issues with extra values but the 'average' function didn't work due to a cast in the wrong order and the new-since-2014 functions 'first' and 'last' were missing. Feel free to cherry-pick and merge/reject as is easiest :)

DanCech:
Closing in favor of #2052

PR #783, opened 2014-07-17
stdev/stddev typo fix

teroxik:
The name of the stdev function doesn't correspond to the generated name of series.

I think that the function stddevSeries should be renamed to stdevSeries but at least it is consistent. So it doesn't cause any problems when using raw JSON data.


cornelf:
👍

PR #831, opened 2014-08-13
backport sortByName() to 0.9.X

Krylon360:
Backporting PR #522 to the 0.9.x branch. This seems to have been in at one point; because the original patch for master (a different PR) was upported from 0.9.x. While testing 10.0-alpha with Grafana; the sortByName() Function did just that..function. When reverting back to the latest 0.9.x source; the function caused all of my grafana panels to go blank. Panel was calling for a function that's not in 0.9.x.

see (https://github.com/grafana/grafana/issues/690) for more details.


jssjr:
:+1: to merging this.

jssjr:
The function tests need to be backported to 0.9.x before we can add in a test for this. I also feel like its unnecessary considering that we're just calling sort/cmp.

deniszh:
:+1: 
I backported that to my branch, but was disttracted from backporting tests, sorry...

PR #890, opened 2014-08-25
Update config-local-settings.rst

mauron85:
Without specifying correct PYTHONPATH static files generation fails with module: graphite.settings cannot be found.


obfuscurity:
Deprecated by #1613.

PR #958, opened 2014-09-14
drawing legend in single column mode

freeseacher:
I need to draw legend in single column mode for some reason. 
that patch provide such possibility


obfuscurity:
While I appreciate the contribution, this feels like a solution trying to find a problem. It would be slightly more practical if instead it allowed you to specify the number of columns (e.g. `numLegendColumns`) rather than limiting it to a (seemingly arbitrary) single column.

PR #971, opened 2014-09-23
Coverage integration

esc:
Coverage integration using ``django-coverage``.

Still fails on Django 1.4 and 1.5 for whatever reason.

Also, it doesn't seem to measure the coverage correctly yet, for example the ``functions`` don't seem to be included.


obfuscurity:
Sorry for the lengthy delay in getting around to this. Fortunately we've been using codecov for a while which seems to work pretty well, so I think this is no longer needed.

PR #973, opened 2014-09-25
Add function to remove empty Series.

kdebski:
In my environment I spawn new servers for services all the time. 
After some time I kill instances but I have leftovers in graphite. It is useful to have it for some time, but I would like to show only live metrics on graphs.

Eg. single instance stats: 

    stats.service.instance1.getData.count

I would like to have query 

    removeEmptySeries(stats.service.*.getData.count) 

that would return only live instances.


bruce-lyft:
:+1:  for this...   There's no good workaround for this right now.

obfuscurity:
Love this idea. Do you have a screenshot of the before-and-after? Just curious.

obfuscurity:
:boom*

Issue #976, opened 2014-09-30
Request for json new subformat  parameter to add / improve returned information to the side client


toni-moreno:
# request for new subformat render parameter ( only for json format)

Hi guys.

I've been playing with some dashboard front-ends  and I've notice that most of them works natively with javascript charting libraries that works with   time series with a different format that graphite  ( highcharts and flot) by example.

when you request data from client JavaScript libraries we request:

```
http://<graphite_host>//render?format=json&from=XXXXX&until=YYYY&target=ZZZZZZ
```

And response could be an array of data
[ { dataseries, target } ]

where data-series is a time-series array and target a string ( usually the label as rendered by the png renderer in graphite).

The main goal of this is request/proposal is improve efficiency (response time) in the client side  and open the  data needed   
## current data-series format

nowadays graphite returns data on this format

```
[ 
  [ y1, x1 ( unix timestamp) ]
  [ y2, x2 ( unix timestamp ) ]
  ...
  ...
  [yN,xN ( unix timestamp ]
]
```

where X are time represented in standard unix timestamp (  seconds ) .

we can sumarize as  **"YXs"** subformat

But the javascript chart libraries ( at least which I know) has works with timestamp in milliseconds and datapoints in the reverse order time first and value of that serie after. We can summarize as  "XmsY"
## proposal

I propose to add a subformat parameter only valid with format=json 

```
http://<graphite_host>//render?format=json&subformat="YXs"from=XXXXX&until=YYYY&target=ZZZZZZ
```

By default if missing graphite renderer should return as ussual ( "YXs" ) for backward compatibility and if present should format datapoins as subformat  
## various  subformats proposed.
- **"YXs"** 
  ( current and default):  
   [ Y , X( in unix timestamp in seconds) ]
- **"XmsY"** 
  ( valid for flot , highcharts ) :  
  [ X(  unix timestamp in milliseconds),Y ]
## adding useful data to every datapoint

These two formats could in the future extended to add information that could not be computed on the client side by example  ( based on how is currently working other dashboard and performance proprietary software like "Wily Introscope or APM  of CA" (http://www.ca.com/us/opscenter/ca-application-performance-management.aspx)
 when  APM Introscope summarizes data it always give us the following "very useful" data.

 [ X , Yavg , Ymax, Ymin , Number of time points from to get computed Y's ]

So I proposed also in the future add 2 new subformats  ( added "C" as Count ) 
- **"YXsC"** 
  ( current and default):  
[...]

PR #980, opened 2014-10-02
added  new parameter to json format [ jsondpfmt = "XmsY" ] 

toni-moreno:
added jsondpfmt (datapoint format) to  XmsY to json format allowing client side apps based on javascript  accelerate data import

As I proposed in https://github.com/graphite-project/graphite-web/issues/976

I've added a new datapoint format parameter with format XmsY , when you query data with this new parameter server will return series datapoints in a more suitable javascript format

X= time
Y=value
 [ X in timestamps in ms , Y] 

So if you do a query

```
http://<graphite>/render?format=json&from=-6h&until=now&target=aliasByNode(dwily.pro.c1.s1.nono.system.disk.sda.disk_octets.read,7)
```

you will get that.

image

else if you do this other query

```
http://<graphite>/render?format=json&jsondpfmt=XmsY&from=-6h&until=now&target=aliasByNode(dwily.pro.c1.s1.nono.system.disk.sda.disk_octets.read,7)
```

image

Now client side can support querys with this format and check if returned serie[i].dpfmt='XmsY' 

a) if returned,  the graphite server support new jsondpfmt parameter and can get data direct from retured json data. without need of reorder and transform X value.

b) else, the  server side is not yet supporting  jsondpfmt  parameter an server will return data in the old format  "YXs" ( default if no parameter  jsondpfmt  is queried)  and client side will have to reorder and transform data.

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -129,6 +129,7 @@ def renderView(request):
       return response
 
     if format == 'json':
+      jsondpfmt = requestOptions.get('jsondpfmt')
       series_data = []
       if 'maxDataPoints' in requestOptions and any(data):
         startTime = min([series.start for series in data])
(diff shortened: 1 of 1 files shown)


brutasse:
@toni-moreno regarding metadata, if you address https://github.com/graphite-project/graphite-web/pull/980#issuecomment-59705231 and make it a *separate* pull request that includes docs and tests, I'm happy to consider it for inclusion. JSON format = 1 pull request, metadata = another pull request.

toni-moreno:
Thank you @brutasse  to let me add this new features.

Correct me if I'm wrong

I should  make :

One PR adding parameter jsondpformat="XmsY"

````
http://<graphite>/render?format=json&jsondpformat=XmsY
```
And  another separate PR with new parameter generic metadata=xxxxx

````
http://<graphite>/render?format=json&metadata={true,cur,avg,max,min,sum}
```

Isn't it?

toni-moreno:
Hi @brutasse I've finally leave this PR for the  jsondpformat parameter. I've also rebased it to have only one valid commit.

I will be enormously grateful if you could merge it.  ( I need  urgently to fix issues importing data with IE8 and  highcharts, ... that my customer has a requirement).

In next days I will do a new PR with parameter metadata as you suggested some weeks.

Thank you very much.

brutasse:
@toni-moreno we're not going to rush a feature into an open source project because you have a customer requirement :) You can always maintain your fork while we're iterating on the feature.

`jsondpfmt`, `XmsY` and `YXs` still look too cryptic to me. No other committers have commented so I don't know how they feel about that change.

In any case this will also need tests, and since it touches a public HTTP API, docs need to be updated.

obfuscurity:
I don't understand why adding this metadata can't be as simple as calling a single parameter, e.g. `verbose=1`, and any formats (json, csv, raw, etc) can be extended independently to support that. Why the focus on arcane tertiary params?

toni-moreno:
Hi @brutasse Of course I hope you will merge it because  you think is a good improvement also for other dashboard projects and not for my customers.:wink:

Once merged,  I have plans to build a PR for this new parameter for grafana ( open source project at which I've already contributed).

@obfuscurity , I had focused , metadata only for json  based format, but we can open  a new discussion about  metadata if you wish in a new PR. I would like to contribute writing this code in the way you think is better for graphite, so will be grateful if you say me how do you like to query metadata and how do you expect to get response from graphite

obfuscurity:
I'd fine with seeing this go in considering our recent additions for `format=dygraph` and `format=rickshaw` but I'm going to simplify it a bit.

toni-moreno:
Hi @obfuscurity  , I'm happy to see your comment. feel free to simplify/refactors as your own PR.

Thank you very much

PR #982, opened 2014-10-06
Handle missing divisor series in divideSeries gracefully.

andrewbaxter:
If all the data points for a series go out of scope then the series effectively disappears, with datalib.fetchData returning an empty series list.  This caused divideSeries to throw an exception.  In practice, graphs would suddenly stop rendering if they contained at least one divideSeries expression with data that got too old.

This is a minimal fix - None values are generated on the fly when required by divideSeries.  Since the divisorSeries is empty, we can't get the path expression for labeling.  A better fix (I think) would be to have datalib.findData return the series, but with all None data points.  I'm not sure how much work that would involve or how much would break, where the current behavior is expected.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -522,7 +522,14 @@ def divideSeries(requestContext, dividendSeriesList, divisorSeries):
 
 
   """
-  if len(divisorSeries) != 1:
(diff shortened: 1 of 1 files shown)


obfuscurity:
Closed in favor of #1656 to resolve merge conflicts. Thanks for your contribution!

PR #983, opened 2014-10-06
Add match parameter to transformNull

andrewbaxter:
None is only transformed when the match (another series) has a valid value at the same position.

We have multiple stats that are reported simultaneously and we expect to compare.  Some stats are from dynamic counters, where the counter doesn't exist until the first event is counted.  The service runs periodically with a varying duration, so there are also periods where all stats are None.  In this case we want to transform Nones to 0 only when a report was generated.


obfuscurity:
The equivalent functionality was added in https://github.com/graphite-project/graphite-web/pull/1322. Thanks for the submission though!

PR #984, opened 2014-10-06
Delay function

andrewbaxter:
This delays a series by a number of time steps.  The delay method is the same as in derivative (Nones appended to the beginning).

Our immediate use case is computing a derivative using divideSeries, to find the percent difference from the previous data point.

### DIFF
--- a/webapp/content/js/composer_widgets.js
+++ b/webapp/content/js/composer_widgets.js
@@ -1088,6 +1088,7 @@ function createFunctionsMenu() {
         {text: 'Power', handler: applyFuncToEachWithInput('pow', 'Please enter a power factor')},
         {text: 'Square Root', handler: applyFuncToEach('squareRoot')},
         {text: 'Time-adjusted Derivative', handler: applyFuncToEachWithInput('perSecond', "Please enter a maximum value if this metric is a wrapping counter (or just leave this blank)", {allowBlank: true})},
+	{text: 'Delay', handler: applyFuncToEachWithInput('scale', 'Please enter the number of steps to delay')},
         {text: 'Integral', handler: applyFuncToEach('integral')},
         {text: 'Percentile Values', handler: applyFuncToEachWithInput('percentileOfSeries', "Please enter the percentile to use")},
         {text: 'Non-negative Derivative', handler: applyFuncToEachWithInput('nonNegativeDerivative', "Please enter a maximum value if this metric is a wrapping counter (or just leave this blank)", {allowBlank: true})},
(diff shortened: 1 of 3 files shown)


obfuscurity:
@andrewbaxter Just curious, is the thinking here (vs using `timeShift()`) that with `delay()` you don't have to think about the step (interval units)?

andrewbaxter:
The goal was really to be able to reimplement derivative on the fly, so that's probably correct for the same reason this method was used in derivative itself.  There may have been other uses for which that property was important as well, but it's been a while.

obfuscurity:
P.S. I know it's completely wrong but your name keeps triggering this for me...

s5e19_james_baxter

andrewbaxter:
No relation :horse:

obfuscurity:
http://adventuretime.wikia.com/wiki/James_Baxter_the_Horse

andrewbaxter:
Yeah, I'm fairly certain we're not related.

andrewbaxter:
It may be a good way to participate in future github discussions though.

obfuscurity:
Apologies in advance, the curl patching method is breaking on the first conflict so I'm going to have to re-commit the changes anew. There may be a better way to do this (command-line rebasing with `git am`) but I'm a little inexperienced with these.

webapp/content/js/composer_widgets.js
```diff
@@ -1088,6 +1088,7 @@ function createFunctionsMenu() {
         {text: 'Power', handler: applyFuncToEachWithInput('pow', 'Please enter a power factor')},
         {text: 'Square Root', handler: applyFuncToEach('squareRoot')},
         {text: 'Time-adjusted Derivative', handler: applyFuncToEachWithInput('perSecond', "Please enter a maximum value if this metric is a wrapping counter (or just leave this blank)", {allowBlank: true})},
+	{text: 'Delay', handler: applyFuncToEachWithInput('scale', 'Please enter the number of steps to delay')},
```
obfuscurity: Note this should be `delay` instead of `scale` here. I'll fix this in my branch.

webapp/tests/test_functions.py
```diff
@@ -354,6 +354,19 @@ def test_changed(self):
             result = functions.changed({}, series)
             self.assertEqual(result, expected)
 
+    def test_delay(self):
+        source = [
+            TimeSeries('collectd.test-db1.load.value',0,1,1,[range(18)] + [None, None]),
+        ]
+        delay = 2
+        expectedList = [
+            TimeSeries('collectd.test-db1.load.value',0,1,1,[None, None] + [range(18)]),
```
obfuscurity: Also missing the `delay()` in the expected TimeSeries name. Fixing.

obfuscurity:
Closing in favor of #1658.

PR #990, opened 2014-10-20
Support for different time formats

penpen:
Support for loading local_settings from conf directory


obfuscurity:
I don't see any advantage to this. Rather, I think adding unnecessary options like this will only serve to confuse users. Regardless, thank you for the submission and my apologies that it took this long to review.

PR #997, opened 2014-10-21
Added restriction for dashboards and metrics #2:)

leoleovich:
Please, find my comment to commit.
I have already test it and add some instructions for you.


toni-moreno:
Hi @leoleovich I've seen this PR  and  I've been looking for something like this, but I would like to know if you can add metric permissions as regexp ?
@brutasse there is any plans to merge this PR or any other which give us  at least metric filtering per user with regexp ?

We are discussing how to add security to grafana

https://github.com/grafana/grafana/issues/1313       

but any approach seems not to be good enough in the frontend.

I'm evaluating  cost of developing a per user metric filtering solution for grafana + graphite  but i would like support from graphite/grafana project owners/main contributors.

What do you think about ?

obfuscurity:
I prefer the approach taken in #1435. It's not ready to merge yet but I think it's a more mature implementation. Closing this one.

Issue #832, opened 2014-08-14
Last value in cached query is not always correct.


obfuscurity:
Per @LukaszStrzelecki in https://github.com/graphite-project/carbon/issues/203 (moving issue to the correct project):

Hi,
Quite recently we started using graphite for our internal monitoring and we are very happy with it so far.
There is one issue we can't understand/resolve though. Our current graphite version is 0.9.12 but we had the same problem in 0.9.10.
The problem we are having is that **when we query graphite (regardless of chosen output format) value for last sample is not always correct.**
Our retention policy is quite simple: 
```
pattern = .*
retentions = 20s:6h, 10m:14d, 1h:42d, 4h:182d, 3d:3y
```
and our aggregation is:
```
[sum]
pattern = .*\.sum[^.]*$
xFilesFactor = 0
aggregationMethod = sum
````
When we make request for any metrics that falls into 10m:14d bucket quite often last value returned is taken from cache for 20s:6h (by cache I mean value emitted but not yet flushed to disk) rather than from value that already exists on disk (or is in cache but for correct time bucket rather than from better resolution one).

**Request 1: (invalid last value)**
```
Last 20  sec (uses 20s:6h) : [{"target": "my.namespace.workers.4.sum", "datapoints": [[3227.0, 1389005520]]}]  
Last 381 min (uses 10m:14d): [{"target": "my.namespace.workers.4.sum", "datapoints": [(...), [129046.0, 1389004800], [3227.0, 1389005400]]}]
Corresponding log
06/01/2014 10:52:40 :: [127.0.0.1:36085] cache query for "my.namespace.workers.4.sum" returned 1 values
```
Last value for second timerange is 3227.0 (same as for first timerange) even though few request before it was bigger than that. (we use aggregationMethod = sum and do not emit negative values)

**Request 2: (valid last value)**
```
Last 20  sec: [{"target": "my.namespace.workers.4.sum", "datapoints": [[null, 1389005740]]}] 
Last 381 min: [{"target": "my.namespace.workers.4.sum", "datapoints": [(...), [129046.0, 1389004800], [68246.0, 1389005400]]}]
Corresponding log
06/01/2014 10:57:08 :: [127.0.0.1:36089] cache query for "my.namespace.workers.4.sum" returned 0 values
```

File info from whisper-info.py:
```
maxRetention: 94608000
xFilesFactor: 0.0
aggregationMethod: sum
fileSize: 66808
Archive 0
retention: 21600
[...]

PR #1002, opened 2014-10-23
[critical] fix bug on carbonlink cache

penpen:
Hi!
there is fix for a bug on carbon-link cache: its not downscale value for lower retention interval.
Example:
lowest precision in archive 10s, for 1d - 60s, agg method -- average. Value sent to graphite: 0.1 (10 req/s for example)
10s: 0, 0, 1, 0, 0, 1, 0.1
60s: 0.5, 0.1 // 0.1 * 60 is 60 req/s, instead of 10.
This fix disable carbonlink query for requests with step higher than lowers pointPerSecond in archive.

### DIFF
--- a/webapp/graphite/readers.py
+++ b/webapp/graphite/readers.py
@@ -164,9 +164,13 @@ class WhisperReader(object):
     time_info, values = data
     (start,end,step) = time_info
 
+    meta_info = whisper.info(self.fs_path)
+    lowest_step = min([i['secondsPerPoint'] for i in meta_info['archives']])
     # Merge in data from carbon's cache
+    cached_datapoints = []
     try:
-      cached_datapoints = CarbonLink.query(self.real_metric_path)
+        if step == lowest_step:
+            cached_datapoints = CarbonLink.query(self.real_metric_path)
     except:
       log.exception("Failed CarbonLink query '%s'" % self.real_metric_path)
       cached_datapoints = []


LukaszStrzelecki:
Hi,
I am an autor of #832 it indeed this one looks like the exact same problem we have.
We have worked around it in similar way (probably less elegant ;)).

Let me try to explain the problem one more time:
Assumptions:
- metric: "my.metric.sum" 
- aggregated with "sum" function (so wider bucket will be a sum of smaller buckets)
- an aggregation scheme of : 10s for 2min, 1min for 1h  (for simplicity)
- imagine that you emit my.metric.sum each 10 second with value of "100" (and you are doing that for long time).

Then if you query for last 2 minutes of data you would get `[100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100]` (12 times)
BUT if you query for last 10 minutes (so we ended up using the 1min:1h aggregates) you will get `[600, 600, 600, 600, 600, 100] `
The last value in response is invalid because it will be taken from cache (which will override the value already stored for most recent minute (which depending when within a minute you query, could be 100, 200, 300, 400 or 500))

I hope it sheds more light on this problem.

Thanks,
Lukasz

deniszh:
OK, problem is clear. But how we need to fix that? @penpen's patch fixes that problem but will break caching, which is not very good. I personally will prefer to have last wrong value than break caching and I suspect I'm not alone. At least we need some trigger to enable/disable fix.

LukaszStrzelecki:
Hi,
I am not sure what is the correct fix here either.
What we did say that is it better to show slightly stale data than invalid one.
I don't see a point in providing an invalid value (it can only cause confusion) 

I think that I will apply that patch locally to our instance. It still gives you the most up to date data if you query for short enough time range, and it does not lie if you look at long time trends.

deniszh:
Hi,
That's my whole point - it's not 'slightly stale data vs invalid one' - it's 'invalid-value-sometimes-if-you-are-using-sum-aggregation-only vs not-working graphite'. E.g. in my case turning off cache for lowest resolution means turning off cache almost completely which kills my graphite instantly - and I do not really care that latest value could be wrong sometimes - graphs looks OK and our scripts will ignore last value anyway.
Maybe we really need to fix that, but not with turning cache off for lowest retention.

obfuscurity:
@deniszh I think you misunderstand the patch, at least the way I'm reading it. With this change we only request metrics from CarbonLink that match the shortest step (`secondsPerPoint`) / highest precision. This makes sense imho since we only cache the raw datapoints, and I suspect this will address a number of issues currently open regarding data inconsistencies.

deniszh:
Hm. It seems you are right, @obfuscurity !
I need to apologize to @penpen and @LukaszStrzelecki then - sorry, Roman and Lukasz, I completely misread your patch! :-1: 
But I have a couple of other questions then.
1. Do we need same for Ceres fetcher too? (it's in same file)
But you can merge it and I can add Ceres later.
2. If I understand correctly 0.9.x code is different but it will be equivalent - during merge it will pick only values which match to interval -  https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/graphite/render/datalib.py#L462-L482 
Am I right?

obfuscurity:
I'm merging this in and will attempt to port it to 0.9.x.

PR #1014, opened 2014-11-02
Fix: Issue: CarbonLink ignore CARBONLINK_HOSTS = []

pha6d:
when configured with CARBONLINK_HOSTS = []
 - select_host fails with webapp/graphite/render/hashing.py, line 88, in get_nodes ZeroDivisionError: integer division or modulo by zero
 - trying to get_connection on empty host list is waste of time (and fails similarly as select_host)


deniszh:
0.9.x has similar patch already applied btw - https://github.com/graphite-project/graphite-web/pull/1270

PR #1016, opened 2014-11-03
Add Vagrantfile

justenwalker:
I've added a Vagrantfile to help with local development & testing.
- Based on Ubuntu 14.04
- Installs all required and optional dependencies during provisioning. 
- It exposes ports 2003, 2004, 7002, and 8080.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 MANIFEST
+.vagrant/
 _trial_temp
 build
 dist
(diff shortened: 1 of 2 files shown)


obfuscurity:
I missed this merge before and only stumbled across it because of https://answers.launchpad.net/graphite/+question/260408. Besides the lack of documentation for this Vagrantfile, it doesn't actually install Carbon.

I would like to revert this merge unless a solid case can be made for keeping it.

obfuscurity:
@SEJeff thoughts?

SEJeff:
@obfuscurity gah my bad. You're absolutely right. Go for it

Issue #626, opened 2014-02-18
Is diffSeries function broken using a constant in 0.9.12?


pythiannunez:
When using diffSeries with a constant (copy/pasting the example in http://graphite.readthedocs.org/en/0.9.10/functions.html) I get an error:

$ curl "http://localhost/render/?target=diffSeries(statsd.numStats,5)&format=json"
(...)
TypeError: can only concatenate list (not &quot;int&quot;) to list

When I try the same with two metrics, it works:

$ curl "http://localhost/render/?target=diffSeries(statsd.numStats,statsd.numStats)&format=json"
[{"target": "diffSeries(statsd.numStats)", "datapoints": [[0.0, 1392659580], [0.0, 1392659640], 
(...)

Full error here: https://gist.github.com/pythiannunez/9075937

PR #1021, opened 2014-11-05
Update diffSeries documentation

caipre:
diffSeries doesn't support constant values. The offset function is recommended
instead.  See https://github.com/graphite-project/graphite-web/issues/626.


obfuscurity:
This was already fixed in https://github.com/graphite-project/graphite-web/pull/1107.

PR #1030, opened 2014-11-14
Apply hashing to remote storage cache key

bitprophet:
Prevents hitting the memcache 250-char key limit when rendering very long metric paths in cluster setups.

This looks like it regressed in f18bb3c7 and since that looks like a squashed merge I can't tell why it was changed during the megacarbon work; presumably it was to update the data used to construct the key. Either way, this change doesn't modify the nature of the key, only ensures it is hashed like all other memcached-using code (eg in `render/views.py`).


drax68:
+1 Hit this issue too. Any chance that this will be merged soon?

bitprophet:
Given I barely contribute, I'd like to merge this but would :heart: a simple :+1: from any other @graphite-project/committers first to make sure I'm not committing a faux pas. An emoji will do! A day or two of silence may be taken as consent :)

JeanFred:
@bitprophet The rationale sounds sensible to me, and the code looks fine − but I’m not familiar enough with the codebase in general for even an emoji on this :-)

JeanFred:
Wait, looks like #1480 is the same as this? (Good news is that @deniszh +1 it ;-)

obfuscurity:
I prefer the bit savings in #1480 but the logic seems ok either way.

deniszh:
Yep, or this or #1480 - both looks fine. Pick one, ditch another, merge. 👍

obfuscurity:
Merged #1480.

PR #1034, opened 2014-11-16
Fix #1028 -- incorrect timezone handling in events

### LINKED ISSUE #1028: Double timezone conversion in events
I'm running current master branch of graphite-web with default sqlite database and django 1.5. Timezone is set to 'Europe/Moscow' in local_settings.py. That's UTC+3.

When I send an event without `when` parameter, it gets stored in sqlite in UTC time zone (3 hours earlier than my local time).

In standard web interface (http://graphite/events/) time is shown correctly in local timezone. But in /events/get_data all timestamps are in UTC-3. If I set timezone to UTC+5, all timestamps become UTC-5, but are still shown correctly on web page. Seems like timezone correction is applied twice only for json output.

I don't know what behaviour is intended (should time be stored in local timezone or in UTC), but it is obviously a bug.

deniszh:
Backport of PR #1029 to 0.9.X

### DIFF
--- a/webapp/graphite/events/views.py
+++ b/webapp/graphite/events/views.py
@@ -1,27 +1,22 @@
 import datetime
-import time
 
-from django.utils.timezone import get_current_timezone
+import pytz
+
+from django.utils.timezone import now, make_aware
 from django.core.urlresolvers import get_script_prefix
 from django.http import HttpResponse
 from django.shortcuts import render_to_response, get_object_or_404
-from pytz import timezone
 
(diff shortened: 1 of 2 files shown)


deniszh:
It's nice to have this fix before 0.9.13, will add it to milestone and try to fix tests.

deniszh:
BTW - I tried to find why tests generate some warnings but failed - same code on master works fine, but generate workings on 0.9.x.
But looks like commit works and fixes issue anyway.

PR #1041, opened 2014-11-28
fix multiplySeriesWithWildcards docstring

beevee:
While `sumSeriesWithWildcards` and `averageSeriesWithWildcards` both have correct documentation, `multiplySeriesWithWildcards` has incorrect and misleading one.


obfuscurity:
Looks right to me, thanks!

PR #1043, opened 2014-12-02
Use encode URL characters in target string

5c077yP:
Hey,

first thanks for this great project!

If been using e.g. '=' signs in my graphite metric name. This causes the __/dashboard__ view to not correctly create a graph view when clicking in the found metric. The generated graph view only contains the characters until the first '=' sign. This happens when the correct target string is given to the `Ext.urlDecode ` function which then internally only returns a part of that string.

E.g. i can see when debugging that the `graphTargetString ` string is `target=host.ip=127-0-0-1.version=v0-0-1.Service.Cpu`, but after calling `var graphTargetList = Ext.urlDecode(graphTargetString)['target'];`  `graphTargetList ` is `host.ip`.

With this PR the target string is first URL encoded using `encodeURIComponent ` before give to `Ext.urlDecode`.

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -913,9 +913,9 @@ function graphAreaToggle(target, options) {
   */
   var graphTargetString;
   if (target.substr(0,7) == "target=") {
-    graphTargetString = target;
+    graphTargetString = "target=" + encodeURIComponent(target.substr(7, target.length-7));
   } else {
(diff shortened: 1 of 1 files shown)


obfuscurity:
Closing this as a WONTFIX for 0.9.x. If the aforementioned concern can be addressed and it can be ported to master we'd love to consider it for inclusion. Thanks!

PR #1046, opened 2014-12-02
Preserve the sign of legendValue values when using a unit system

aroben:
This seems to have been broken since the `system` argument to `legendValue()` was added in 1a826949732c394b03c7e2e94e65af82762b9103. `format_units()` already calls `abs()` as needed so we don't need to do so before passing the value into it.

/cc @jssjr


aroben:
Anything I can do to help get this merged?

drawks:
Looks good to me, not sure why it can't be automerged. Can you do a quick rebase against current master?

aroben:
Rebased! (It was just the test file that conflicted.)

PR #1053, opened 2014-12-08
fix a bug: when color is like 808080 and passed as an int

liyichao:
fix https://answers.launchpad.net/graphite/+question/220017


obfuscurity:
Sorry for the delay in reviewing this one. I'd like to get this merged in but first I'm going to make sure we add a test for this and also get it ported over to the master branch.

PR #1069, opened 2014-12-29
Bypass loopback prevention for deliberate loopbacks (0.9.x)

steve-dave:
Sometimes you may want to query a separate local graphite-web instance and if you have configured one or more of your clusters hosts with a hostname of localhost or an IP of 127.x.x.x then this would be the case so with this patch they is treated as a remote host.

0.9.x port of https://github.com/graphite-project/graphite-web/pull/1070

### DIFF
--- a/webapp/graphite/storage.py
+++ b/webapp/graphite/storage.py
@@ -98,6 +98,10 @@ def is_local_interface(host):
   if ':' in host:
     host = host.split(':',1)[0]
 
+  # If "localhost" or a loopback IP has been specified it is a deliberate reference
+  if host == 'localhost' or host.startswith('127.'):
+    return False
+
   try:
     sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     sock.connect( (host, 4242) )


steve-dave:
(Y) for switch worst-case, but theoretically it's also possible that you may want the benefit of the auto-detection as well as deliberate loopbacks.

Is the change I have proposed horribly inefficient or is there another reason it cannot be merged as-is?

steve-dave:
@obfuscurity basically it just adds some flexibility without any real complication.  It allows for remotehost targeting to be exported to a reverse proxy, load balancer or port forward, etc.

steve-dave:
@drawks @obfuscurity if you're still against this idea I'm happy to close the PR... don't want to clog up the queue.

obfuscurity:
I defer to @deniszh and @drawks here.

drawks:
Yeah I'm :-1: on this, but would be in favor of just making an explicit "enable/disable" switch around the entire "loopback prevention" codepath. I think it [the original code] is poorly conceived and should either be removed whole cloth or minimally have a full switch around it.

SEJeff:
I'm +meh on it as well, but equally agree with @drawks. The existing code could use some sprucing up

steve-dave:
Closed in favour of https://github.com/graphite-project/graphite-web/pull/1134

PR #1070, opened 2014-12-29
Bypass loopback prevention for deliberate loopbacks (master)

steve-dave:
Sometimes you may want to query a separate local graphite-web instance and if you have configured one or more of your clusters hosts with a hostname of localhost or an IP of 127.x.x.x then this would be the case so with this patch they is treated as a remote host.

Master port of https://github.com/graphite-project/graphite-web/pull/1069

### DIFF
--- a/webapp/graphite/util.py
+++ b/webapp/graphite/util.py
@@ -77,6 +77,10 @@ def is_local_interface(host):
   if ':' in host:
     host = host.split(':',1)[0]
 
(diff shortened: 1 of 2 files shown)


steve-dave:
Tests updated and passing.  Now just depends on https://github.com/graphite-project/graphite-web/pull/1069 being accepted.

steve-dave:
Closed in favour of https://github.com/graphite-project/graphite-web/pull/1135

PR #1071, opened 2014-12-29
Use POST instead of GET for find and render requests (0.9.x)

steve-dave:
Sometimes due to the limitations imposed by various webservers the query string generated is too long which causes cluster server requests to fail.  This patch changes requests to POST instead of GET which are not subject to those limitations.

If this patch is accepted an issue should be opened to implement the equivalent for master.  Making this change in master is slightly more complicated so I haven't done this proactively yet.


steve-dave:
Now configurable :smiley:

steve-dave:
@obfuscurity do you think it's OK to merge now?

obfuscurity:
Instead of supplying a string value (`GET` or `POST`), I would change the setting to a boolean with a default state to `False`. This makes the conditional logic simpler as well.

```python
#REMOTE_STORE_HTTP_METHOD_POST = False # Use POST instead of GET for remote requests
```

steve-dave:
@obfuscurity done.

obfuscurity:
Nice. :v:

obfuscurity:
Has this been ported to master?

steve-dave:
Not yet.  Master is too different to allow a quick cherry-pick but I agree it needs to be done.

PR #1086, opened 2015-01-06
simplify README and add diagram

deniszh:
Backport from master.
N.B. - maybe it's need to be pointed to https://graphite.readthedocs.org/en/0.9.13/ instead of https://graphite.readthedocs.org/en/latest/ but 0.9.13 is not exist yet and pointing to 0.9.13-pre1 is quite pointless IMO...


deniszh:
Oops, wrong project :-1:

Issue #1105, opened 2015-01-14
rare bug in computing yMin and yMax setting yMin to 0 for negative series


g76r:
When drawing a series with negative values AND null some values AND no explicit yMin param in url AND drawNullAsZero=true in url, then yMin is computed to 0, hiding all negative values (possibly all values if series contains only negative and null values).

Yes, I know this is rare, however when diffing series, negative values can be usual, and this leads to an empty graph when the values are all below 0.

Example:
/render?target=diffSeries(small.numbers.with.some.nulls,big.numbers.with.some.nulls)&drawNullAsZero=true

I know this is not easy to reproduce since it's data-dependant, but I've got a fix (and soon a pull request...).

PR #1106, opened 2015-01-14
Fixed automatic computing of yMin and yMax when drawNullAsZero=true

g76r:
yMin was set to 0.0 for series containing at less one null value, even
if it contains negative values, hence hiding those values

yMax was not set to 0.0, even if series contains nulls and its maximum
non-null value is < 0.0

Fixes issue #1105

### DIFF
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -1019,10 +1019,10 @@ class LineGraph(Graph):
   def setupYAxis(self):
     seriesWithMissingValues = [ series for series in self.data if None in series ]
 
-    if self.params.get('drawNullAsZero') and seriesWithMissingValues:
+    yMinValue = safeMin( [safeMin(series) for series in self.data if not series.options.get('drawAsInfinite')] )
+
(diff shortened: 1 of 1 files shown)


steve-dave:
:+1: looks sane to me.  Do tests need to be updated?

g76r:
@steve-dave 

Well... I'm not sure what you mean by "need".

This patch did not broke the tests, therefore no test case MUST be updated.

However, AFAIK the modified method is not tested at all, therefore it SHOULD be a good idea in the future to add some tests on graph output (or at less on json/csv output, since svg/png are probably not so easy to test because of changes in cairo and other libraries), regardless of this pull request.

steve-dave:
@g76r sorry, what I should have said is that if there are existing tests they need to be updated to complete this patch.  IF there are existing tests they are not sufficient as they are passing without this fix.

g76r:
Thx for the merge.
Do you want me to backport the same commit to 0.9.x ?

obfuscurity:
Sure, although we'll need to really be careful about regressions here since we're close to release.

PR #1107, opened 2015-01-14
Fixed misleading doc for function diffSeries

g76r:
Constants are not supported.
See issues #626 and #415 that assume not supporting constants is a bug.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -297,6 +297,17 @@ def diffSeries(requestContext, *seriesLists):
 
     &target=diffSeries(service.connections.total,service.connections.failed)
 
+  To diff a series and a constant, one should use offset instead of (or in
+  addition to) diffSeries
+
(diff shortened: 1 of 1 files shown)


SEJeff:
Would you mind adding a blurb about using offset() to do what people obviously assume they should be able to do with diffSeries please?

g76r:
@SEJeff done.

SEJeff:
Any chance you'd mind doing this in 0.9.x as well so it makes it into 0.9.13, our last 0.9.x releases?

g76r:
@SEJeff  Done. #1108

g76r:
BTW I wonder how one can still use 0.9.x, I switched to master-in-production for a while ;-)

SEJeff:
I wish more of our users were like you :)

As @mleinart slowly merges the megacarbon bits back into master, it is going to be shaping up into a very nice thing indeed

PR #1109, opened 2015-01-16
adding timeslice method

plockaby:
This change will allow selecting only a slice of time from a line to display on the graph. This is useful in the following situation:

A customer just wants to know how much bandwidth their site used last year. They don't care about the link to their site. At the beginning of the year the customer is using port ge-0/0/0. Halfway through the year the customer is upgraded to ae0, an aggregated port that contains ge-0/0/0. All traffic that goes through ae0 necessarily goes through ge-0/0/0. Were a graph to be made over the entire year for all traffic to the site then it would have to include both ge-0/0/0 and ae0 to cover the whole year except that from 'Jul 1' to 'Dec 31' would be double the actual value because it would include traffic from both ge-0/0/0 and ae0. This new function, combined with sumSeries, allows me to say 'show ge-0/0/0 from Jan 1 to Jun 30 and ae0 from Jul 1 to Dec 31' and get one line that is an accurate representation of traffic for the customer for the entire year.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2475,11 +2475,22 @@ def timeShift(requestContext, seriesList, timeShift, resetEnd=True):
   return results
 
 
-def timeSlice(requestContext, seriesList, startSliceAt, endSliceAt):
+def timeSlice(requestContext, seriesList, startSliceAt, endSliceAt="now"):
   """
   Takes one metric or a wildcard seriesList, followed by a quoted string with the
   time to start the line and another quoted string with the time to end the line.
   See ``from / until`` in the render\_api_ for examples of time formats.
+
+  Useful when, for example, you've collected all of your metrics by router port
+  but you want to show metrics by customer location and the customer's router port
+  has changed over time.
+
+  Example:
+
+  .. code-block:: none
+
+    &target=timeSlice(network.core.port1,"00:00 20140101","11:59 20140630")
+    &target=timeSlice(network.core.port1,"12:00 20140630","now")
 
   """
 
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -21,7 +21,7 @@ import random
 import time
 
 from graphite.logger import log
-from graphite.render.attime import parseTimeOffset
+from graphite.render.attime import parseTimeOffset, parseATTime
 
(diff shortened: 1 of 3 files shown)


obfuscurity:
This is a really unusual edge case. It smells of antipattern because someone wasn't collecting their metrics properly.

plockaby:
Well, we collect our metrics by router interface and our engineers care about metrics by router interface. Our customers care about metrics by site and their sites might be served by several different interfaces over the course of time depending on infrastructure upgrades or network topology changes. This pull request also imitates a feature that we've used heavily with RRD files.

obfuscurity:
Ok, I think I grok the use case a little better now. There are still a few issues we'll need to address before we can merge it.

- [ ] documentation needs an example
- [ ] needs to be wired into the UI
- [ ] needs test(s) please :smile:

plockaby:
I'll get the tests in real soon now.

plockaby:
Sorry about all the commits. I'm not generally a python programmer so it took me a short bit to figure out how to test it. This should be ready.

obfuscurity:
Do you know how to squash commits via rebase? It would be nice to clean some of that up.

plockaby:
Squashed it down to five commits.

obfuscurity:
I'm not sure the "Useful when..." explanation will make sense to readers that aren't following this PR. Perhaps something like this instead?

```
Useful for filtering part of a series among a wider range of data.
```

webapp/graphite/render/functions.py
```diff
@@ -2474,6 +2474,45 @@ def timeShift(requestContext, seriesList, timeShift, resetEnd=True):
...
+
+  .. code-block:: none
+
+    &target=timeSlice(network.core.port1,"00:00 20140101","11:59 20140630")
+    &target=timeSlice(network.core.port1,"12:00 20140630","now")
+
+  """
+
+  results = []
+  series = seriesList[0] # if len(seriesList) > 1, they will all have the same pathExpression, which is all we care about.
```
obfuscurity: Can you move this comment to a line above by itself?

plockaby: Easy fix. That var wasn't even being used anyway.

obfuscurity:
Cool, thanks for hammering this into shape. :smile_cat:

PR #1110, opened 2015-01-17
unicode support in drawing text

gescheit:
Since version 1.1.6 pycairo text_path() has support of  text in unicode.


obfuscurity:
We switched to preferring cairocffi over pycairo a while back. Fortunately it _appears_ that the latter has always supported unicode for `text_path()` so I think it's safe to merge this in without any dependency changes (e.g. requirements.txt).

PR #1113, opened 2015-01-19
Add documentation about rewrite-rules.conf

brandonwamboldt:
This is a very useful feature and should be better documented. I couldn't find any info about it on Google and ended up having to look into the source code to see what was valid/invalid.


obfuscurity:
This is awesome. I noticed this as well when writing the current chapter of my book. Would you mind backporting this to the `0.9.x` branch as well?

PR #1115, opened 2015-01-20
Include pysqlite by default

obfuscurity:
Graphite-web uses SQLite by default (including the `DATABASES` setting in `local_settings.py`) so it makes no sense imho not to install this driver by default.


brutasse:
https://github.com/graphite-project/graphite-web/pull/1116#issuecomment-70630825

PR #1116, opened 2015-01-20
Include pysqlite by default

obfuscurity:
Graphite-web uses SQLite by default (including the DATABASES setting in local_settings.py) so it makes no sense imho not to install this driver by default.

Cherry-picked into `master` from `0.9.x`, refs #1115.


brutasse:
It's not needed :) Python has native sqlite support since version 2.5.

https://docs.python.org/2/library/sqlite3.html

obfuscurity:
Ha, nice.

PR #1118, opened 2015-01-22
fix XSS issue in user saved graphs

piotr1212:
Fixes XSS vulnerability when saving a graph.

When logged, click in a graph to save to my graphs, then enter html/js code. When exanding the User Graph tree this code will get executed.


SEJeff:
From my phone I can't tell if this is against master or 0.9.x branches. Can you also forward or back port this to the other branch so this fix makes it into both 0.9.13 and 0.10 please?

PR #1119, opened 2015-01-22
Adding two more render functions

chicagobuss:
Native formats for rickshaw (d3) and dygraph (canvas) javascript graphing libraries.

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -173,6 +173,34 @@ def renderView(request):
         add_never_cache_headers(response)
       return response
 
+    if format == 'json_dygraph':
+      labels = ['Time']
+      datapoints = [[ts] for ts in range(data[0].start, data[0].end, data[0].step)]
+      for series in data:
+        labels.append(series.name)
+        for i, point in enumerate(series):
+          datapoints[i].append(point if point is not None else 'null')
+      line_template = '[%%s000%s]' % ''.join([', %s'] * len(data))
+      lines = [line_template % tuple(points) for points in datapoints]
+      result = '{"labels" : %s, "data" : [%s]}' % (json.dumps(labels), ', '.join(lines))
+      response = HttpResponse(content=result, mimetype='application/json')
+      response['Pragma'] = 'no-cache'
+      response['Cache-Control'] = 'no-cache'
+      return response
+
+    if format == 'json_rickshaw':
+      series_data = []
+      for series in data:
+        timestamps = range(series.start, series.end, series.step)
+        datapoints = [{'x' : x, 'y' : y} for x, y in zip(timestamps, series)]
+        series_data.append( dict(target=series.name, datapoints=datapoints) )
+      if 'jsonp' in requestOptions:
+        response = HttpResponse(
+          content="%s(%s)" % (requestOptions['jsonp'], json.dumps(series_data)),
+          mimetype='text/javascript')
+      else:
+        response = HttpResponse(content=json.dumps(series_data), mimetype='application/json')
+
     if format == 'raw':
       response = HttpResponse(content_type='text/plain')
       for series in data:


obfuscurity:
I resisted this sort of opinionated formatting a while back but I'm softening to the idea if it helps make everyone's lives easier.

chicagobuss:
Yeah, I just really want to get these into the next official release so we can stop patching them into our deploy every time we make another graphite instance :)

SEJeff:
Doing the transforms in JavaScript is quite trivial but I'm not strongly against this either.

webapp/graphite/render/views.py
```diff
@@ -173,6 +173,34 @@ def renderView(request):
...
+    if format == 'json_dygraph':
+      labels = ['Time']
+      datapoints = [[ts] for ts in range(data[0].start, data[0].end, data[0].step)]
+      for series in data:
+        labels.append(series.name)
+        for i, point in enumerate(series):
+          datapoints[i].append(point if point is not None else 'null')
+      line_template = '[%%s000%s]' % ''.join([', %s'] * len(data))
+      lines = [line_template % tuple(points) for points in datapoints]
+      result = '{"labels" : %s, "data" : [%s]}' % (json.dumps(labels), ', '.join(lines))
```
brutasse: That's a rather odd way to generate JSON. Could you refactor that to read `result = json.dums(some_structure)` with native types (lists / dicts) rather than string interpolation?

brutasse:
I'm :+1: for the same reason as @obfuscurity but it'd be nice to have tests for new features :)

toni-moreno:
Hi.

I would like to remember , that I did a PR some months ago (https://github.com/graphite-project/graphite-web/pull/980) and also opened a discussion (https://github.com/graphite-project/graphite-web/issues/976)  on how to add more datapoint formats to the render side to help  external developers to get data easier to different javascript libraries . And also extended metadata not available at client side.

While we are working with highcharts and flot these new parameters jsondpfmt,jsonseriemd could also be used by you @chicagobuss .

I invite you to resume the discussion and I'll be happy to help you.

brutasse:
@toni-moreno I like `?format=json_dygraph` more than `?format=json&jsondpfmt=something`. It's more explicit. I also just added a comment to #980.

obfuscurity:
@chicagobuss We would also need documentation patches.

SEJeff:
Sort of analogous to this, what does everyone think conceptually of making functions.py pluggable in a similar way to @brutasse's work on the storage finders?

 - default functions.py
 - magical list setting to include import paths for more functions

We could then have a graphite-functions-json that supported dyngraphs, highcharts, rickshaw, flot, etc.

cc: @gingerlime

obfuscurity:
Yes, I've made it no secret that I'd like to see pluggable functions as a first-tier citizen within graphite-web. This would also theoretically make it easier to expose Graphite's capabilities through the API for consumers like Grafana.

toni-moreno:
@brutasse I agree that  ```?format=json_dygraph``` is more explicit than ``` ?format=json&jsondpfmt=dygraphs```  or ```?format=json&jsondpfmt=XmsY``` but with this second way client side frontend like grafana can make the same query across different graphite versions thus not supporting the new format and check  if supported by only check the existence of a datapoint  as I explaide (https://github.com/graphite-project/graphite-web/pull/980#issue-44665207)

```` 
if(serie[i].dpfmt && serie[i].dpfmt='dygraph' ) ´{
 // dygraph format supported
// do something with returned data
} else {
//old version of graphite doesn't support dygraph
convert_format()
// do something with returned data
}
```

In other way as format=json is not known you wil get de default format this is a png image.


As a workarround the dashboard fronted developers should maintain a registry with supported formats on each graphite version and check graphite version explicitely.

IMHO the proposed way by adding a new jsondpfmt=xxxxx is backward-compatible and also more scalable and more mantenible for both frontend and backends developers

gingerlime:
Thanks for cc'ing me @SEJeff 

To contribute my rather not-that-well-thought-out 2 cents (and I'm not sure if my opinion should really count for much, so feel free to ignore everything I say):

I also think that those transformations are pretty easy to do client-side. This definitely wasn't something I felt was a pain-point. 

Pluggable functions seems like a good idea. I can even imagine building a completely different query language a-la influxDB / fnordmetric with an SQL-like query... But I have no clue about the implementation challenges of such parsing to be honest.

Adding metadata (as proposed on #980 / #976) has a definite appeal to me. One of the pain-points I did feel was when combining different targets on a single HTTP request - it was hard to know which results matches which target on the response. (I noticed a similar problem with influxDB btw). Forcing to use an alias was the only workaround and you can still end up with collisions if I remember correctly. Some kind of an identifier would be nice that matches the request / metadata.

Also agree with @toni-moreno about backwards compatibility despite it looking less elegant / explicit.

Completely OT, and I don't mean to hijack the issue, but I've seen a couple posts on HN declaring graphite "practically dead" (backend in particular). Does this have any merit?

obfuscurity:
@toni-moreno Please don't hijack this PR with an existing conversation taking place in #980.

@gingerlime You read a couple comments on HN that the project is dead? Wow, I guess we better close up shop. :wink:

Frankly, I'm happy to continue helping people be productive with storage and rendering features that actually work. It may not be the glitziest project, and the release cycles may run longer than we'd like, but people actually use it for production loads at significant scale, every day. It has a clustering design that works quite well and isn't subject to the sort of data loss you read about routinely in "that other project".

SEJeff:
@gingerlime: someone is wrong on the Internet? Nooooooo: https://news.ycombinator.com/item?id=8907875

gingerlime:
Thanks for the update / link. And you did a really good job with your response @SEJeff. Very reassuring.

chicagobuss:
FWIW we tried doing this client side for a long time, but doing it in
graphite was up to 10x faster for complex multi-target graphs.  Our
dashboard app can draw hundreds of dygraphs per page and without this
graphite addition we crush it.

I just wanted it added "for real" so we wouldn't have to keep patching it
in ourselves.  :)

Josh

obfuscurity:
@chicagobuss We're in favor of this change as long as there's documentation and tests.

chicagobuss:
Cool. I'll try to get that added soon.

Dieterbe:
> I resisted this sort of opinionated formatting a while back but I'm softening to the idea if it helps make everyone's lives easier.

FWIW I'm also in favor of the aility to emit data directly in the format that a renderer expects. would certainly help with flot as well.

pcn:
:+1:

obfuscurity:
Moving this change over to #1605.

PR #1122, opened 2015-01-24
Convert cached_datapoints from dict if needed.

foozlevazquez:


foozlevazquez:
I was getting an error here that the cached_datapoints were a dict, so this code addressed the issue.  Perhaps this was due to some misconfiguration (of mine) upstream?

landlord11:
You aren't alone; I was seeing this too.  This resulted in graphite-web failing any time carbon returned data from its cache.

afilipchik:
Here is another way of doing it:
if hasattr(cached_datapoints, 'iteritems'):
   cached_datapoints = cached_datapoints.iteritems()

Could someone merge it, cause graphite-web is kind of broken now?

brutasse:
This isn't misconfiguration, just a change in carbon's internal. Cache queries now returns datapoints as a dictionnary.

PR #1125, opened 2015-01-28
Add long names of reduceSeries and mapSeries that are referenced in docs

foxdan:


jaingaurav:
Would it be possible to merge this? grafana has already been updated to use mapSeries/reduceSeries, but the UI is broken without this change.

PR #1133, opened 2015-02-03
Sort nodes with numbers: [n1, n12, n2] changed to [n1, n2, n12]

lomik:
Reordering nodes with any count of numbers in name.

For example:

Nodes list before fix:
[p1, p12s1, p12s12, p12s2, p2] 

Nodes list after fix:
[p1, p2, p12s1, p12s2, p12s12]

### DIFF
--- a/webapp/graphite/metrics/views.py
+++ b/webapp/graphite/metrics/views.py
@@ -85,10 +85,10 @@ def search_view(request):
   return json_response_for(request, dict(metrics=results))
 
 
-node_sort_RE = re.compile(r"([^0-9]+|[0-9]+)")
+node_sort_RE = re.compile(r"(\d+)")
 
 def node_sort_key(node):
-  matches = node_sort_RE.findall(node.name)
+  matches = node_sort_RE.split(node.name)
   for index, value in enumerate(matches):
     if value.isdigit():
       matches[index] = int(matches[index])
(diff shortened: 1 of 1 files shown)


drawks:
The intention of this is good, but I've got some reservations about baking our own "natural sort" and then making it the default behavior. Among other concerns the strategy of using a fixed regex for pulling the numeric portions seems overly simplistic. The [natsort](https://github.com/SethMMorton/natsort) is intended for this exact use case and takes into account many other factors such as locale and speed when sorting large sets.

That said, my primary objection here is a lack of test coverage for the change ;)

lomik:
In fact natsort uses regex: https://github.com/SethMMorton/natsort/blob/master/natsort/utils.py#L128 

"split" do same, but more simple that my solution (I don't know "split" method):
```
# natsort
>>> re.compile(r"(\d+)").split("aaa3cc234xx")
['aaa', '3', 'cc', '234', 'xx']

# my
>>> re.compile(r"([^0-9]+|[0-9]+)").findall("aaa3cc234xx")
['aaa', '3', 'cc', '234', 'xx']
```

SEJeff:
I concur with @drawks, please add test coverage for this change. Also, please add a setting and make this disabled by default. Changing defaults should normally not be done lightly as it could mess up expected "features" users currently rely on.

PR #1136, opened 2015-02-04
Add missing mapSeries and reduceSeries function mappings

jaingaurav:
It seems only the short form mappings were added as opposed to the long form.


jaingaurav:
Seems pull request #1125 already addressed this

PR #1138, opened 2015-02-05
put in a short delay between retries

landlord11:
If a data request fails there was no delay between retries, I thought there probably should be something sort of pause so added a 0.1s delay.

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -12,6 +12,7 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License."""
 
+import time
 from graphite.logger import log
 from graphite.storage import STORE
 from graphite.readers import FetchInProgress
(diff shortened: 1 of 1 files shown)


deniszh:
Sorry, but why? IMO graphite is quite slow by itself :)
Looks like you trying to solve some other problem, which one?

landlord11:
On Mon, 9 Feb 2015, Denis Zhdanov wrote:

> Sorry, but why? IMO graphite is quite slow by itself :)
> Looks like you trying to solve some other problem, which one?

I'm not sure what sense it makes to catch an exception and retry when 
the retries happen so fast that the underlying condition probably 
couldn't have resolved itself?  What sort of problems is the existing 
exception handling attempting to catch that would resolve themselves so 
quickly?  Maybe it would be better to remove the exception handling 
completely?

You are correct, though, that there was another issue.  Carbonlink is 
returning a dict and not a list so the loop over the data returned by 
carbonlink fails.  Someone else has submitted a pull request (#1122) to 
fix this.  For this problem as long as carbon-cache has some data in 
cache the query fails.  Since graphite fired off the retry requests so 
quickly it was highly unlikely that a call that failed the first time 
would ever succeed.  By slowing down the retries it made it more likely 
that it could catch the carbon daemon at a time when it had flushed its 
cache to disk.

And probably a better retry logic than what I submitted could make 
sense; some sort of backoff logic so the initial attempts are faster 
than my suggestion, for example.

I do agree, it of course is better to fix the underlying condition that 
causes the exception in the first place!

Best regards,
Brian

obfuscurity:
I have to agree with @deniszh, there's no reason to insert an arbitrary delay here.

PR #1146, opened 2015-02-11
Add documentation about Dashboard CLI

blacked:
I wrote tool to export\import\copy\sync graphite dashboard (https://github.com/blacked/graphite-dashboard).
It allow to track dashboards in git, and because it keep dashboards as YAML files it also easy to modify them.
I suppose it will be useful to refer to it in Graphite dashboards documentation.

### DIFF
--- a/docs/dashboard.rst
+++ b/docs/dashboard.rst
@@ -1,4 +1,3 @@
-
 ============================
 The Dashboard User Interface
 ============================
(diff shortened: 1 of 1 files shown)


brutasse:
I think that's something that belongs to `tools.rst` (tools that work with graphite) since it's not shipped directly with graphite-web or one of the graphite projects.

You could link to your documentation directly instead of copying it -- that would avoid duplicating information and having to update it in two different places :)

PR #1152, opened 2015-02-12
Add graphite-dashboard to tools.rst

blacked:
I wrote tool to export\import\copy\sync graphite dashboard (https://github.com/blacked/graphite-dashboard).
It allow to track dashboards in git, and because it keep dashboards as YAML files it also easy to modify them.
I suppose it will be useful to refer to it in Graphite dashboards documentation.

### DIFF
--- a/docs/tools.rst
+++ b/docs/tools.rst
@@ -147,6 +147,13 @@ Visualization
   using Flot with a multitude of display options (Multiple Y-axis, Bars, Lines, Points, smart Y-axis formats and much more).
   Click and drag selection rectangle to zoom in on any graph.
 
+`graphite-dashboardcli`_
+  a tool for manage graphite dashboards from command line:
+
+  - import\Export dashboards from\to Graphite servers
+  - synchronize dashboards between multiple Graphite servers
+  - keep dashboards in YAML format
+
 `graphitus`_
   a client side dashboard for graphite built using bootstrap and underscore.js.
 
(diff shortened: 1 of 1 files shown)


obfuscurity:
nope

We're not going to mix instructions for an external project with official instructions for the native project. Please remove the bits from `dashboard.rst`. Also, the entry in `tools.rst` belongs under **Visualization**, not **Other**.

blacked:
@obfuscurity sorry, I forgot checkout it. Fixed, and moved to Visualization

docs/tools.rst
```diff
@@ -147,6 +147,12 @@ Visualization
   using Flot with a multitude of display options (Multiple Y-axis, Bars, Lines, Points, smart Y-axis formats and much more).
   Click and drag selection rectangle to zoom in on any graph.
 
+`graphite-dashboard`_
+  a tool for manage graphite dashboards from command line:
+  - import\Export dashboards from\to Graphite servers
```
obfuscurity: I think you need an empty line between the bullets and the line before them.

obfuscurity:
Also, don't take this the wrong way :heart:, but `graphite-dashboard` is a really confusing name for a cli tool used to manage the native Graphite dashboard, don't you think?

blacked:
@obfuscurity well, I was thing about name, maybe it now the best, but I try to follow pip convention, other   tools\modules called graphite-something. graphite-dashboard-cli will be too long :)

obfuscurity:
Sure, but graphite-dashboard is almost certainly going to confuse users that don't realize that Graphite's built-in dashboard is already distributed with graphite-web.

obfuscurity:
Couldn't you call it something like graphite-dash-cli instead?

blacked:
Renamed to graphite-dashboardcli

obfuscurity:
@blacked Thanks! :sparkles: :beer:

PR #1154, opened 2015-02-13
Relative time timezone fix

MFAnderson:
Changing `attime.py`'s treatment of relative times with specified timezones to have consistent behaviour with absolute times/restore behaviour from 0.9.12, as discussed in #1151

PR #1157, opened 2015-02-13
porting timezone fix to 0.9.x branch

MFAnderson:
Port of #1154 to 0.9.x branch

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -39,5 +39,6 @@ gunicorn
 pytz
 sphinx
 sphinx_rtd_theme
+mock
 cairocffi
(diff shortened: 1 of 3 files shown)


obfuscurity:
Thank you for your help with this and especially for the tests. :zap:

requirements.txt
```diff
@@ -39,5 +39,6 @@ gunicorn
 pytz
 sphinx
 sphinx_rtd_theme
+mock
```
JeanFred: @MFAnderson This should not be necessary, is it? mock is installed by tox for the tests, and is not necessary to run the application.

PR #1158, opened 2015-02-14
Add --interface flag, to allow devel server to only listen on 127.0.0.1

malkowski:
Defaults to 0.0.0.0, so existing behavior remains unchanged. But it's easier when running a dev instance on a laptop and not wanting to mess around with OS X firewall setting, to just specify --interface=127.0.0.1 instead.

### DIFF
--- a/bin/run-graphite-devel-server.py
+++ b/bin/run-graphite-devel-server.py
@@ -8,7 +8,7 @@ option_parser = OptionParser(usage='''
 %prog [options] GRAPHITE_ROOT
 ''')
 option_parser.add_option('--port', default=8080, action='store', type=int, help='Port to listen on')
-option_parser.add_option('--interface', default="0.0.0.0", action='store', help='Interface to listen on')
+option_parser.add_option('--interface', default='0.0.0.0', action='store', help='Interface to listen on')
 option_parser.add_option('--libs', default=None, help='Path to the directory containing the graphite python package')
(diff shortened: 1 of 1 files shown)


bin/run-graphite-devel-server.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/local/opt/python/bin/python2.7
```
obfuscurity: Need to remove this.

bin/run-graphite-devel-server.py
```diff
@@ -8,6 +8,7 @@
 %prog [options] GRAPHITE_ROOT
 ''')
 option_parser.add_option('--port', default=8080, action='store', type=int, help='Port to listen on')
+option_parser.add_option('--interface', default="0.0.0.0", action='store', help='Interface to listen on')
```
obfuscurity: It's a small detail, but would you mind changing those to single-quotes?

obfuscurity:
Thanks! :smile_cat:

bin/run-graphite-devel-server.py
malkowski: Man, that sticks out like a sore thumb now that you point it out. I normally go the other direction with everything in double quotes, so I know the importance of small details like this. Consistency in code is under-valued IMO.

obfuscurity:
Restarting errored build (thanks Obama).

PR #1159, opened 2015-02-14
Set env variable earlier

brutasse:


orestes:
Just came by to say I encountered this bug as well. Looking forward to the merge.

PR #1161, opened 2015-02-17
Hot data cache

melan:
This is a 2nd part of the Hot data cache feature. Idea of the feature is add an alternative location for metrics to serve dashboards with short time window -2 - -3 hours. Ideally this location should be located on an in-memory file system.

Here is how it works: There are 2 dirs are in the STANDARD_DIRS variable. Later system process these directories in order. If it can find a metric with enough data in the 1st directory - it will only read a file from there but will skip reading from the 2nd location. If the hot-data file does not provide enough information - the regular file will be used instead.


melan:
My change passed all test at least once w/o any further changes in the testable area - Travis CI fails some random tests usually because of timeout while downloading packages for the testing env.

PR #1164, opened 2015-02-22
New single-hue lineColors schemes

obfuscurity:
Some new templates to offer single-hue lineColors.

ocean1
ocean1
ocean2
ocean2
forest1
forest1
forest2
forest2
sunset1
sunset1
sunset2
sunset2
moonlight1
moonlight1
moonlight2
moonlight2
lava1
lava1
lava2
lava2


steve-dave:
:+1: can't see any reason that this should not be merged!

PR #1135, opened 2015-02-04
Make local cluster host detection configurable and disabled by default (master)

steve-dave:
As discussed in https://github.com/graphite-project/graphite-web/pull/1069 it seems to be preferable to make this behaviour switchable and (I think) disabled by default.

@drawks @obfuscurity @SEJeff 

Master port of https://github.com/graphite-project/graphite-web/pull/1134

PR #1166, opened 2015-02-24
Make local cluster host detection configurable and disabled by default (master)

steve-dave:
As discussed in #1069 it seems to be preferable to make this behaviour switchable and (I think) disabled by default.

@drawks @obfuscurity @SEJeff

Master port of https://github.com/graphite-project/graphite-web/pull/1167

Originally https://github.com/graphite-project/graphite-web/pull/1135


webapp/graphite/storage.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, finders=None, hosts=None):
 
     if hosts is None:
       hosts = settings.CLUSTER_SERVERS
-    remote_hosts = [host for host in hosts if not is_local_interface(host)]
+    remote_hosts = [host for host in hosts if not settings.REMOTE_EXCLUDE_LOCAL or not is_local_interface(host)]
```
bmhatfield: This feels like a weird `if not ... or not` to me. Can you talk out the intent here?

steve-dave: @bmhatfield would you prefer a `if not (... or ...)` or do you just want an explanation of the logic?

Basically it's just including hosts if either is_local_interface() is not enabled by config (REMOTE_EXCLUDE_LOCAL) or is_local_interface() returns false.

bmhatfield: Doing some testing locally, it looks like defaulting REMOTE_EXCLUDE_LOCAL to 'False' is the inverse of what we have today, no?
```python
hosts = ['onehost', 'twohost', 'threehost', 'localhost', 'greenhost']

def check_item(item):
  print("Checking if localhost: ", item)
  if item == 'localhost':
    return True
  else:
    return False

print([host for host in hosts if not check_item(host)])

EXCLUDE_LOCAL = True

print([host for host in hosts if not EXCLUDE_LOCAL or not check_item(host)])

EXCLUDE_LOCAL = False

print([host for host in hosts if not EXCLUDE_LOCAL or not check_item(host)])
```

ouputs:

```
('Checking if localhost: ', 'onehost')
('Checking if localhost: ', 'twohost')
('Checking if localhost: ', 'threehost')
('Checking if localhost: ', 'localhost')
('Checking if localhost: ', 'greenhost')
['onehost', 'twohost', 'threehost', 'greenhost']
('Checking if localhost: ', 'onehost')
('Checking if localhost: ', 'twohost')
('Checking if localhost: ', 'threehost')
('Checking if localhost: ', 'localhost')
('Checking if localhost: ', 'greenhost')
['onehost', 'twohost', 'threehost', 'greenhost']
['onehost', 'twohost', 'threehost', 'localhost', 'greenhost']
```

Would we not want this to be backwards compatible by default?

drawks: It is different than the current default behavior, which is intended. Not sure how best to handle communicating that change. Perhaps make it default on now with a comment that the we intend to reverse the default behavior in a future release? Honestly I think the change will affect very few people, and the current default behavior seems questionably implemented at best.

steve-dave:
@bmhatfield @drawks do you think this is mergeable?  The change of default definitely needs to be documented and how to do that is definitely the question.  That said, master is technically not released yet and when it does get released there are possibly quite a few backwards incompatibilities with 0.9.x.  How about for #1167 (0.9.x port) the default should be unchanged (I'll modify the patch if we are in agreement) and a comment added to the config highlighting the fact that the behaviour will change from 1.x (master) onwards?  It also needs to be in 1.x release notes as well I would expect, but as far as I can tell there isn't even a placeholder for 1.x release notes yet.

PR #1134, opened 2015-02-04
Make local cluster host detection configurable and disabled by default (0.9.x)

steve-dave:
As discussed in https://github.com/graphite-project/graphite-web/pull/1069 it seems to be preferable to make this behaviour switchable and (I think) disabled by default.

@drawks @obfuscurity @SEJeff 

0.9.x port of https://github.com/graphite-project/graphite-web/pull/1135

PR #1167, opened 2015-02-24
Make local cluster host detection configurable and disabled by default (0.9.x)

steve-dave:
As discussed in #1069 it seems to be preferable to make this behaviour switchable and (I think) disabled by default.

@drawks @obfuscurity @SEJeff

0.9.x port of https://github.com/graphite-project/graphite-web/pull/1166

Originally https://github.com/graphite-project/graphite-web/pull/1134


steve-dave:
@drawks test added.

steve-dave:
@drawks @bmhatfield so do we leave this as is or flip the default so that it doesn't change the current behavior?

steve-dave:
ping @drawks @bmhatfield

deniszh:
Hello @steve-dave,
Sorry for delay! We usually agree that default for 0.9.x should not change existing behaviour.
Could you please rebase this change? Much appreciated!

PR #1169, opened 2015-02-25
removing useless double call

g76r:
see discussion here 9da1d8a896ae08deaa36c2c97ea5a5710ae24c0d


obfuscurity:
Yoinks, good catch.

g76r:
Yes, I think with such one-line patches I may be awarded some Nobel prize or at less the Turing Award. ;-)

rbuckland:
You're welcome.

;-) it had me puzzled for a day.

PR #1172, opened 2015-02-26
Added a 'noNullPoints' flag to the render url

kokje:
Whenever sparse or bursty data is written to graphite and needs to be read back, the response can get considerably delayed due to the number of points returned. This is really significant when the call is made at the UI level. Adding this flag improves performance by reducing the payload size as well as not requiring the developer to iterate through datapoints because it is implied that missing points from the list are in fact 'Null'


obfuscurity:
@kokje Seems like a nice addition. At the very least would you mind adding it to the [Graph Parameters](https://github.com/graphite-project/graphite-web/blob/master/docs/render_api.rst#graph-parameters) section of the documentation. Bonus points for a new test.

obfuscurity:
Replaced by https://github.com/graphite-project/graphite-web/pull/1612.

PR #1175, opened 2015-02-27
Update docs/tools.rst

blacked:
Add graphite-to-zabbix proxy. It allow make zabbix triggers based on graphite data.


obfuscurity:
Looks good, thanks!

Issue #168, opened 2013-03-11
Metrics filter - ignore hierarchy (dots)


tomerpeled:
Is there any way to filter metrics while ignoring the hierarchy?

For example: Say I have the following metrics:

stats_count.A.B.TestMetric 
stats_count.A.TestMetric
stats.A.B.TestMetric
stats.A.B.TestMetric

How can I sum TestMetric under stats_count only?

I tried the followings with no success:

stats_counts.*.*.TestMetric - obviously this won't work...
sumSeriesWithWildcards(stats_counts.[A-Z\.]*[A-Z]*.TestMetric,1)
sumSeriesWithWildcards(stats_counts.[A-Z]*[.]*[A-Z]*.TestMetric,1)

Is it possible at all? If not I think that will be nice feature...

cbowman0:
Merged into master via #1639.

PR #1178, opened 2015-03-05
add support for globstars in standard finder

mark-5:
This should address the request originally made in #168.

The code change to support this was simple enough, that I figured I could send a pull request to start some discussion on it.

There may be a more efficient way to implement globstar, but at least this doesn't seem pathological, and has very little impacting on existing code and performance.


cbowman0:
I rebased this in #1639.  Very minor changes needed and those were primarily in test_finders, since that file had some structural changes.

cbowman0:
#1639 was merged.

PR #1180, opened 2015-03-05
Update attime.py to use pytz

kokje:
Change instantiation of tzinfo object to match that in views.py because get_current_timezone() returns the LocalTimezone implementation of tzinfo which does not support 'localize' and results in an AttributeError

### DIFF
--- a/webapp/graphite/render/attime.py
+++ b/webapp/graphite/render/attime.py
@@ -12,16 +12,18 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License."""
 
+import pytz
 from datetime import datetime,timedelta
 from time import daylight
 from django.utils import timezone
+from django.conf import settings
 
 months = ['jan','feb','mar','apr','may','jun','jul','aug','sep','oct','nov','dec']
 weekdays = ['sun','mon','tue','wed','thu','fri','sat']
 
 def parseATTime(s, tzinfo=None):
   if tzinfo is None:
-    tzinfo = timezone.get_current_timezone()
+    tzinfo = pytz.timezone(settings.TIME_ZONE)
   s = s.strip().lower().replace('_','').replace(',','').replace(' ','')
   if s.isdigit():
     if len(s) == 8 and int(s[:4]) > 1900 and int(s[4:6]) < 13 and int(s[6:]) < 32:


kokje:
I think something similar also needs to be done on the 0.9.x branch, because we're currently on that and are facing this issue

brutasse:
`get_current_timezone()` returns a proper timezone if `pytz` is installed. Are you actually seeing a `LocalTimezone` being returned with pytz installed?

kokje:
Yup, I noticed it when I got this error while inspecting code from the UI
AttributeError: 'LocalTimezone' object has no attribute 'localize'. 
But since we're on 0.9.x which is pretty different I tried out these lines 
tzinfo = timezone.get_current_timezone()
tzinfo.localize(datetime.strptime(s,'%H:%M%Y%m%d'), daylight)
I was able to fix it by explicitly using pytz. I saw that's what views.py does and passes in the object so I thought I would change it here as well

SEJeff:
@JeanFred: What do you think of this? Seems relevant to #1276

obfuscurity:
Thanks!

obfuscurity:
@brutasse @SEJeff any idea why this is failing the [lint test](https://travis-ci.org/graphite-project/graphite-web/jobs/82476434)? I grok the error:

```
/home/travis/build/graphite-project/graphite-web/webapp/graphite/render/attime.py:18:1: F401 'timezone' imported but unused
```

... but that seems incorrect. I see where we're calling `timezone` in [two](https://github.com/kokje/graphite-web/blob/ef5650c9c4deb2a8e1ff0f5a9ee1e312992edbaf/webapp/graphite/render/attime.py#L47) [different](https://github.com/kokje/graphite-web/blob/ef5650c9c4deb2a8e1ff0f5a9ee1e312992edbaf/webapp/graphite/render/attime.py#L70) places. We can add `# noqa` to the `import` but then what's the point of testing flake8?

obfuscurity:
Gonna go ahead and noqa that for now. If it can be fixed "correctly" someone let me know.

SEJeff:
@obfuscurity You sure about that?

[check this out](http://hastebin.com/piyomukehe.py)

No use of `timezone`

SEJeff:
Is it possible there was a bad merge or something weird? When I clicked view full file in the changes section of this PR, it does show `timezone` being used. However, the current code in master clearly does not.

obfuscurity:
Oh :poop:, you're right. I was looking at @kokje's master which was almost certainly behind. I'll fix that right now. thanks.

obfuscurity:
Yup, it was in https://github.com/graphite-project/graphite-web/commit/b4ee01533b6d6f68bdfd9423574d25f7ed300126.

SEJeff:
:)

Issue #1173, opened 2015-02-27
No graphite composer


sglwlb:
I install the graphite-web and config it, after that, i access it from browser, there is no "graphite composer". I use develop tool to catch the problem, the errors are:
Uncaught ReferenceError: Ext is not defined
(index):91 Uncaught ReferenceError: Ext is not defined
(index):101 Uncaught ReferenceError: GraphiteBrowser is not defined
but, the Ext and GraphiteBroser are all there.

And, i found the response body length is zero, like follow:

Remote Address:[::1]:9999
Request URL:http://localhost:9999/content/js/ext/resources/css/ext-all.css
Request Method:GET
Status Code:200 OK (from cache)
Request Headers
Provisional headers are shown
User-Agent:Mozilla/5.0 (Macintosh; Intel Mac OS X 10_9_5) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/40.0.2214.115 Safari/537.36
Response Headers
Access-Control-Allow-Headers:origin, authorization, accept
Access-Control-Allow-Methods:GET, OPTIONS
Access-Control-Allow-Origin:*
Connection:close
Content-Length:0
Content-Type:text/css
Date:Fri, 27 Feb 2015 02:03:20 GMT
Last-Modified:Wed, 21 Aug 2013 17:11:04 GMT
Server:Apache/2.2.15 (CentOS)
Vary:Accept-Encoding

Follow are the error message when i config the database, is it relate to this problem?
/etc/graphite-web/local_settings.py.back
/usr/lib/python2.6/site-packages/graphite/settings.py:231: UserWarning: SECRET_KEY is set to an unsafe default. This should be set in local_settings.py for better security
  warn('SECRET_KEY is set to an unsafe default. This should be set in local_settings.py for better security')
/usr/lib/python2.6/site-packages/django/conf/__init__.py:75: DeprecationWarning: The ADMIN_MEDIA_PREFIX setting has been removed; use STATIC_URL instead.
  "use STATIC_URL instead.", DeprecationWarning)
/usr/lib/python2.6/site-packages/django/core/cache/__init__.py:82: DeprecationWarning: settings.CACHE_* is deprecated; use settings.CACHES instead.
  DeprecationWarning
Creating tables ...
Creating table account_profile
Creating table account_variable
Creating table account_view
Creating table account_window
Creating table account_mygraph
Creating table dashboard_dashboard_owners
Creating table dashboard_dashboard
Creating table events_event
[...]

PR #1188, opened 2015-03-18
Serve staticfiles in the webapp if whitenoise is installed

brutasse:

Whitenoise is compatible with Django 1.4 and above.


obfuscurity:
Should fix #1173.

obfuscurity:
Thanks @brutasse, LGTM. :+1:

PR #1192, opened 2015-03-20
Use OrderedDict in groupByNode

jstultz:
Checking `key` for membership in `metaSeries.keys()` is inefficient and
unnecessary since we can use the `setdefault` method to replace that
logic.

Additionally, the `keys` list appears to only be necessary for the
purposes of maintaining ordering; this can be managed instead by using
an OrderedDict instead of a dict and a list.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -20,7 +20,6 @@ import time
 from datetime import datetime, timedelta
 from itertools import izip, imap
 from os import environ
-from collections import OrderedDict
 
 from graphite.logger import log
 from graphite.render.attime import parseTimeOffset, parseATTime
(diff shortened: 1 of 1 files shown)


drawks:
OrderedDict was introduced in python 2.7, requirements for this project are 2.6+

coaxmetal:
this can be changed to still maintain that ordered list of keys and then output using the comprehension at the end, and the setdefault optimization will still be good (instead of iterating over the keys every time)

jstultz:
Updated to not use OrderedDict; this is now a pathetically small pull request that just removes the unnecessary and inefficient call to `.keys()` for membership checks. Removed in two other places as well.

drawks:
thanks

PR #1197, opened 2015-03-23
Remove the metrics/search endpoint

slackhappy:
it isn't use by graphite-web itself,
and the cost of initializing it when a tree
contains millions of metrics, even when explicitly
initialized in wsgi, seems to be prohibitive.

Its functionality seems to be superseded by browser/search
(grep over the index file)
and metrics find
(uses the storage engine for searching)

I could make another change, that introduces a setting to disable loading searcher, but I thought I might start with the more controversial one to flush out any existing uses.


SEJeff:
I'm all for burning legacy crud, but am somewhat certain we have external graphite dashboards and tooling that uses this endpoint. If that is the case, I'd be pretty strongly against removing it and more in favor of fixing it to use a cache or something

slackhappy:
I've tried searching for code outside the project (there is none within) that calls `metrics/search`, or uses the `keep_query_pattern` parameter, but didn't find much.  As a concrete example, grafana uses `metrics/find`, but not `metrics/search`.  The last update to the behavior of `search_view` was 19ed37ae10 in 2011, which seemed to add some commented out code, and remove a parameter.

I'm not sure how to figure out if there are projects out there that use it, but if there are, it would be possible to bring it back.

It has a response format that isn't very helpful for someone using it for search.

Imagine this sample index file:

```
a.b.leaf
a.b.c.d.leaf1
a.b.c.d.leaf2
a.x.leaf
leaf
```

and this session (the view just jsonifies the results, so the shell will do):

```
python2.7 manage.py shell
> from graphite.metrics.search import searcher
> dict(metrics=list(searcher.search("a")))
{'metrics': [{'is_leaf': False, 'path': None}]}
```

hmm, not very useful, it didn't even return the child nodes of `a`
using `keep_query_pattern` is even worse:

```
> dict(metrics=list(searcher.search("a", keep_query_pattern=True)))
AttributeError: 'NoneType' object has no attribute 'split'
```

the only interesting results occur when you search for a node that has leaves:
```
> dict(metrics=list(searcher.search("a.b")))
{'metrics': [{'is_leaf': False, 'path': None},
  {'is_leaf': True, 'path': 'a.b.leaf'}]}

> dict(metrics=list(searcher.search("a.b.c.d")))
{'metrics': [{'is_leaf': False, 'path': None},
  {'is_leaf': True, 'path': 'a.b.c.d.leaf1'},
  {'is_leaf': True, 'path': 'a.b.c.d.leaf2'}]}
```

It seems like this hints that you could use it for querying target globs expansion if you already knew the exact targets you were looking for, but then:
```
> dict(metrics=list(searcher.search("*.*.c.d.*")))
{'metrics': [{'is_leaf': False, 'path': None},
  {'is_leaf': True, 'path': 'a.b.c.d.leaf1'},
  {'is_leaf': True, 'path': 'a.b.c.d.leaf2'},
  {'is_leaf': True, 'path': 'leaf'}]}
```
What is the `leaf` path doing at the end of that list?   It doesn't match the given pattern!  (Perhaps a longstanding bug, failing to check match occurs before the end of the pattern?)

I was thinking of suggesting implementing `metrics/search` in terms of `metrics/find`, but I don't see a specification, and my attempts to derive a useful one have failed.

SEJeff:
@graphite-project/committers thoughts?

bitprophet:
Tentative :+1:, I have the same vague concerns as @SEJeff had in his original reply, but I just checked quickly on my internal fork and don't see us using this anywhere, nor are any of our dashboardy bits (gdash, tessera etc) apparently using it.

So that's one data point in favor of removal, at least.

obfuscurity:
@jssjr @tmm1 @f3ew Are you guys using this endpoint with any of your stuff?

obfuscurity:
I'm a big :+1: in favor of removing cruft. Thanks for this!

slackhappy:
One thing i noticed, which probably depends on your configuration, is that you'll need something that kicks off `graphite.util.write_index`, which `wsgi.py` was effectively doing.

obfuscurity:
@slackhappy Folks should be using https://github.com/graphite-project/graphite-web/blob/master/bin/build-index.sh via cron for that.

slackhappy:
Agreed. Just wanted to point out.  Thanks for merging.

PR #1204, opened 2015-04-16
Fixing remote fetch timeout handling

deniszh:
Small bugfix - thread.is_alive() need to be called for timeout check, exception is not rising for that - https://docs.python.org/2/library/threading.html#threading.Thread.is_alive
With this fix you will get an error in log when you starts timeouting.


deniszh:
Strange, it seems my tree was on some strange state... 67 commits and only 2 are mine.

deniszh:
@bmhatfield - Looks what I found!

deniszh:
Blergh... will create clear one.

PR #1205, opened 2015-04-16
Add template() function to parameterize seriesList arguments

tmm1:
Adds a `template()` function which acts kind of like a search/replace + eval. Any references to `$var` inside the template are replaced with a corresponding value.

The `template()` function takes both positional and kwargs. Positional args are numbered (i.e. `$1`, `$2`) and values can be passed in either as args or via the requests' query parameters. You can also provide both, to use the positional argument as a default but allow it to be overridden by the request:

```
/render?target=template(sumSeries(hosts.$1.cpu-*.idle.value),"fe1")
/render?target=template(sumSeries(hosts.$1.cpu-*.idle.value))&template[1]=fe1
/render?target=template(sumSeries(hosts.$1.cpu-*.idle.value),"fe1")&template[1]=fe*
```

Kwargs are named, and can also be passed in via the query parameters:

```
/render?target=template(sumSeries(hosts.$hostname.cpu-*.idle.value),hostname="fe1")
/render?target=template(sumSeries(hosts.$hostname.cpu-*.idle.value))&template[hostname]=fe*
/render?target=template(sumSeries(hosts.$hostname.cpu-*.idle.value),hostname="fe1")&template[hostname]=fe*
```

All the examples above are functional on this pull request. I need to flesh out the docs (and maybe add some tests), but am looking for some feedback on this direction (does it make sense, would it be useful to others) before I continue.

cc @jssjr

### DIFF
--- a/webapp/tests/test_render.py
+++ b/webapp/tests/test_render.py
@@ -3,6 +3,7 @@ import json
 import os
 import time
 import logging
+import shutil
 
 from graphite.render.hashing import hashRequest, hashData
 import whisper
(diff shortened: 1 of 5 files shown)


SEJeff:
I'd strongly prefer there be unit tests for this function before it is merged. Would you mind taking care of that please?

tmm1:
> I'd strongly prefer there be unit tests for this function before it is merged. Would you mind taking care of that please?

Definitely, wasn't planning to merge this without tests and docs.

I've got some basic tests going in here, but would like to add some that actually operate on series. However I'm not seeing an easy way to create dummy metrics to exercise the seriesLists paths.. do you have any ideas?

tmm1:
Alright, added some docs and figured out how to create fake metrics during the test to exercise the pathExpression substitution cases.

webapp/tests/test_render.py
```diff
@@ -132,3 +158,77 @@ def test_correct_timezone(self):
...
+        response = self.client.get(url, {
+                 'target': 'template(constantLine($num),num=12)',
+                 'format': 'json',
+                 'from': '07:01_20140226',
+                 'until': '08:01_20140226',
+        })
+        data = json.loads(response.content)[0]['datapoints']
+        # all the from/until/tz combinations lead to the same window
+        expected = [[12, 1393398060], [12, 1393401660]]
+        self.assertEqual(data, expected)
```
tmm1: Need another test here that passes numeric args via the query params

tmm1: Added the new tests and made sure numbers passed via query parameters are parsed correctly.

jssjr:
:+1: 

Like @tmm1 mentioned, this PR is useful to our organization. We store graph target strings externally and render them with graphite-web. This allows us to store common graph targets for things like disk utilization and substitute host names or device names back into the parameter string before rendering.

I suspect there are other use cases as well. Would other organizations find this change valuable too?

SEJeff:
You guys really kick ass and help make graphite better. I can see it being useful in some cases and see the worth in merging it.

Thanks @tmm1

PR #1209, opened 2015-04-19
Backport of #1188 "Serve staticfiles in the webapp if whitenoise is installed" to 0.9.x

deniszh:
Whitenoise is compatible with Django 1.4 and above.

Tested with synthesize and manual whitenoise installation, but please check more.


deniszh:
@brutasse , @obfuscurity - please check, thanks!

gwaldo:
From a quick read it looks ok, but I don't have the bandwidth to test it.

esc:
deniszh: how do I go about checking this?

deniszh:
@esc: Very good question, indeed. I just found that I do not know. Synthesize working properly on Django 1.4-1.7 and there's no need to setup /static alias at all.... :confused:

obfuscurity:
@deniszh you're welcome :wink:

deniszh:
@obfuscurity: yep, that's good in general, but not in that specific case.
BTW latest Django 1.8 doesn't work with synthesize, gives strange error, will try to find out what's going on.

obfuscurity:
@deniszh If you're trying to upgrade to Django 1.8, it's no longer Synthesize. Synthesize uses requirements.txt, which is frozen to 1.4.

deniszh:
According to https://github.com/graphite-project/graphite-web/issues/1252#issuecomment-113413285 and whitenoise docs it doesn't work with python 2.6 and 0.9.13 must working with it.
Maybe we need to import whitenoise only if python version is 2.7 ?

deniszh:
Add that change and squashed commits.

SEJeff:
This looks reasonable to me, although I'm not the biggest fan of skipping using a web server like Apache or nginx.

Personally, it shouldn't be a warning in the check deps script, perhaps just an INFO message?

deniszh:
@SEJeff - will change it, thanks!

deniszh:
Is anyone has idea why Travis is b0rking on that change for all Django's except 1.8 ?

alecxvs:
Probably a similar issue to https://github.com/hw-cookbooks/graphite/issues/227
A newer django-tagging is only compatible with django 1.7+, it needs to be locked down for older djangos.

deniszh:
Ah, screw it. Will make another commit, this one is too distant from 0.9.x

Issue #1102, opened 2015-01-13
Inconsistent graphs using memcached with clustered graphite


arma26:
Hello, I'm getting inconsistent results when memcached is enabled across my 4 node cluster. Each node has 4 caches behind 1 relay. Graphite performs as expected with memcached disabled (commenting out MEMCACHE_HOSTS), albeit with higher cpu usage. The issue occurs both when picking 1 node and having the remaining nodes use it's memcached instance and when each node has it's own memcached instance.

Please note the time at the top right of the images, this is occurring between refreshes. This isn't limited to collectd metrics, but collectd's metrics don't contain my domain name. The cluster is running at REPLICATION_FACTOR = 2.

1: http://i.imgur.com/HWLm1bw.png
2: http://i.imgur.com/OwFEPB1.png
3: http://i.imgur.com/jSOavnl.png
4: http://i.imgur.com/xi2f8zU.png

These pictures were taken from only 1 node, however all nodes show identical results, even when each has it's own memcached instance. All the nodes are sharing the same mysql db so that I can use the same dashboards across all the nodes as I load balance between them.

On branch 0.9.x 5a4b3be420fad31fddcd883a71e5b2c8fc5f2296

obfuscurity:
Only your frontend render host(s) should use memcached. You should not use it with any of your backend graphite-web instances. If you're running multiple frontend webs, use a shared memcached across all of them.

deniszh:
I think we need to include that warning in release notes for 0.9.13, I'l
make a PR
I'm also wondering why local memcached is not working. @bmhatfield, what do
you think?

6:59, вт, 13.01.2015, Jason Dixon <notifications@github.com>:

bmhatfield:
Was this behavior experienced before #1010 and #1026? Or is it new since upgrading?

When tracing this out for the issue filed with 1010, I'm not sure what would lead to this. There's an interaction somewhere that I haven't picked up on.

obfuscurity:
@deniszh @bmhatfield You *should* expect inconsistencies when using memcached on backend nodes (in a multi-node cluster). This is not surprising. We should absolutely have documentation explaining how to use memcached *properly* but this report does not signify any sort of regression in my mind.

deniszh:
[...]

Issue #1037, opened 2014-11-26
upgrade from 0.9.12 to 0.9.x  breaks existing cluster returning intermittent subsets of metrics.


fessyfoo:
We tried to upgrade to  0.9.x (ce239313)  from 0.9.12 (from pip) 

After upgrading requests to graphite-web that involved multiple metrics gave back intermittent results.

I was using a query like this to inspect the results.  (no cache) 

`http://$host/render?noCache=true&target=carbon.agents.*.pointsPerUpdate&from=-1h&until=now&format=json`

each request would only return metrics for a subset of the total metrics.    

using `http://graphite-qa.otenv.com/metrics/find?query=carbon.agents.*'`  however all of the expected metrics were listed. 

we have a cluster where all nodes run  graphite-web, carbon-relay and carbon-cache, and memcached we have a load balancer in front of all the graphite-web servers.   the graphite-web servers are all configured to talk to each other. 

#1022  mentions having front end and backend graphite-webs and not allowing the backend graphite-webs to use memcached.      I think we _also_ had this problem because we are using one layer of graphite-web.    when doing the above queries without noCache I would find that all nodes returned the same subset of total metrics. 

Are we using an unsupported configuration?  if so it's very confusing.  And it is functioning differently than 0.9.12. 

Any advice is appreciated.   We had to roll back, and we may need to cherry pick the commits we need.   (maxDataPoint, and safe pickle bug with collections)  But we'd rather  be working at the head with everyone else.

fessyfoo:
I've since learned how to probe graphite better and see what's coming from where.  In particular `local=1` is required for me as `localOnly=true` didn't seem to work, though it's in the version of the code we're using. 

there is some possibility the issue was with our carbon-relays not replicating all metrics properly. (we've found that if carbon-cache is restarted carbon-relays start throwing exceptions connecting to them. `File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 451, in _startRunCallbacks
    raise AlreadyCalledError`  and need to be restarted. ) 

We'll try to replicate this issue and upgrade again.  (especially now that 0.9.13 was released) 

closing this issue for now.

Issue #1022, opened 2014-11-05
Bulk-fetch metrics from remote hosts breaks some queries


lamont:
I was excited about the commit around pull request 1010 and rolled it out today with bad results.  Specifically, queries using divideSeries(sum(),sum()) were failing with the error "divideSeries second argument must reference exactly 1 series"

My clustered setup involves consistent-hash with a replication factor of 2, and I wonder if the bulk-fetch was tested with replicated data.

Additionally, simple queries like the following were returning 200 but an empty body:

target=alias(scale(sumSeries(collectd.app112.ShipmentDownloadTimer.counter-TotalHits%2Ccollectd.app111.ShipmentDownloadTimer.counter-TotalHits)%2C60),%20'ShipmentsDownloaded')&from=-3600s&rawData=true

Rolling back and forward across the 13 node cluster is pretty easy, so if someone has some specific debugging they'd like me to look for, I'm happy to try things.

Thank you.

PR #1212, opened 2015-04-21
Namespace the request-cache with the local node address

jssjr:
This is a quick hack to enable shared memcache support in a carbon cluster using replication factor >= 2. Appending the local node information to the hash key stored in memcache will prevent false positives from occurring while rendering (and caching) pickles for the upstream graphite-web responsbile for rendering the user specified format.

I don't love the implementation, but it seems to be working well in my limited tests. (Currently on an :airplane:, and I don't have a proper local cluster setup to test with.)

/cc https://github.com/graphite-project/graphite-web/issues/1022
/cc https://github.com/graphite-project/graphite-web/issues/1102
/cc https://github.com/graphite-project/graphite-web/issues/1037

### DIFF
--- a/webapp/graphite/render/hashing.py
+++ b/webapp/graphite/render/hashing.py
@@ -33,11 +33,11 @@ def hashRequest(request):
   return compactHash(normalizedParams)
 
 
-def hashData(targets, startTime, endTime):
+def hashData(targets, startTime, endTime, node):
   targetsString = ','.join(sorted(targets))
   startTimeString = startTime.strftime("%Y%m%d_%H%M")
   endTimeString = endTime.strftime("%Y%m%d_%H%M")
-  myHash = targetsString + '@' + startTimeString + ':' + endTimeString
+  myHash = targetsString + '@' + startTimeString + ':' + endTimeString + '/' + node
   return compactHash(myHash)
 
 
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -40,6 +40,7 @@ from graphite.render.attime import parseATTime
 from graphite.render.functions import PieFunctions
 from graphite.render.hashing import hashRequest, hashData
 from graphite.render.glyph import GraphTypes
+from graphite.storage import STORE
 
 from django.http import HttpResponse, HttpResponseServerError, HttpResponseRedirect
 from django.template import Context, loader
@@ -99,7 +100,7 @@ def renderView(request):
       targets = requestOptions['targets']
       startTime = requestOptions['startTime']
       endTime = requestOptions['endTime']
-      dataKey = hashData(targets, startTime, endTime)
+      dataKey = hashData(targets, startTime, endTime, STORE.local_host)
       cachedData = cache.get(dataKey)
       if cachedData:
         log.cache("Data-Cache hit [%s]" % dataKey)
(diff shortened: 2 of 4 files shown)


webapp/graphite/storage.py
```diff
@@ -34,6 +34,11 @@ def __init__(self, directories=[], remote_hosts=[]):
     self.directories = directories
     self.remote_hosts = remote_hosts
     self.remote_stores = [ RemoteStore(host) for host in remote_hosts if not is_local_interface(host) ]
+    self.local_host = [host for host in remote_hosts if is_local_interface(host)]
```
jssjr: There must be a more *python-y* way to do this...

webapp/graphite/render/views.py
```diff
@@ -40,6 +40,7 @@
 from graphite.render.functions import PieFunctions
 from graphite.render.hashing import hashRequest, hashData
 from graphite.render.glyph import GraphTypes
+from graphite.storage import STORE
```
jssjr: Since we're determining the local node's host info in the `Store()` init, we need to import the `STORE` object here. Is this OK, or should I be determining the host info earlier. (When parsing settings?)

webapp/graphite/storage.py
deniszh: You can use 'next' to replace
```python
self.local_host = [host for host in remote_hosts if is_local_interface(host)]
if len(self.local_host) == 1:
  self.local_host = self.local_host[0]
else:
  self.local_host = 'local'
```
with
```python
self.local_host = next((host for host in remote_hosts if is_local_interface(host)),'local')
```

webapp/graphite/render/hashing.py
```diff
@@ -33,11 +33,11 @@ def hashRequest(request):
   return compactHash(normalizedParams)
 
 
-def hashData(targets, startTime, endTime):
+def hashData(targets, startTime, endTime, node):
```
deniszh: You also need to fix tests in https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/tests/test_render.py#L97-L102 btw - just humble reminder. :)

jssjr: Yep! This needs tests to verify behavior. Thanks for the reminder.

webapp/graphite/storage.py
jssjr: Yep. That is much cleaner.

SEJeff:
CAN HAS MOAR TESTS?

deniszh:
@jssjr 
@apg-pk propose another fix in PR #1278
Not sure which is better though...

obfuscurity:
I'd like to get this merged in, any lingering concerns or missing tests?

deniszh:
Really wondering if it's still needed after #1279 ...

obfuscurity:
@deniszh We seem to be dealing with two different problems. In #1279 we avoid negatively caching empty responses. Here in #1212 we're trying to avoid conflicting responses among memcached peers for the same metric.

deniszh:
So, then we need to merge it. 
It looks good and have tests, and it's last issue in next release's milestone for graphite-web, yay!
@obfuscurity ? @jssjr ?

obfuscurity:
:boom:

deniszh:
:+1: 
![:+1:] 
(http://i.giphy.com/U45ze76aj7DvG.gif)

PR #1213, opened 2015-04-22
Cache policy

klynch:
Currently, requests made for JSON data for a given metric are cached using a single expiration value. Repeatedly requesting many requests with a large query window can degrade the performance of the entire system with no real information game. In many situations, it is beneficial to have near realtime values for smaller query windows, but cache larger queries for longer amounts of time. This PR introduces a policy for caching requests of certain durations for specified amounts of time.


klynch:
Are there any issues with merging this feature in?

obfuscurity:
Why should this only apply to JSON? Also, the change to `parseOptions()` only affects line graphs, not JSON afaict. Don't you want to change https://github.com/Squarespace/graphite-web/blob/cache-policy/webapp/graphite/render/views.py#L239 as well?

klynch:
This policy is applied only in the case when an explicit `cacheTimeout` is
not requested in the query (and a policy is defined). At the time #L239 is
executed, we have not yet calculated the requested time range. Once we do
this, we can then update the `cacheTimeout` if applicable. Since we update
the `cacheTimeout` variable in `parseOptions()`, we don't need to modify
much else.

`renderView()` is a little hard to follow, but I believe `line` format
queries is cached by
https://github.com/Squarespace/graphite-web/blob/cache-policy/webapp/graphite/render/views.py#L116
and `pie`, `raw`, and `svg` format queries are cached by
https://github.com/Squarespace/graphite-web/blob/cache-policy/webapp/graphite/render/views.py#L215

JSON queries were not being cached before but should have been.

`raw`, `csv`, and `pickle` formats were not being cached before so I did
not modify that since I am unfamiliar with the uses of those formats.
Though in retrospect, I suppose it makes sense to cache at least `csv`. Do
you have any guidance regarding which of these should actually be cached?
Am I missing any code paths?

klynch:
Am I wrong about this, or are there still any outstanding issues? I believe this is a great way to cut down on load to the rendering services.

obfuscurity:
@klynch Sorry for letting this slide so long. After further review, I see that I originally misinterpreted the way this worked. I like that you're doing this on the request parsing. Thanks for this. :icecream:

PR #1220, opened 2015-05-12
Simple patch to add time units to the Y Axis unit systems

Simage:


webapp/graphite/render/glyph.py
```diff
@@ -126,6 +126,19 @@
...
     ('K', 1000.0   )),
+  'sec': (
+    ('Y', 60*60*24*365),
+    ('M', 60*60*24*30),
+    ('D', 60*60*24),
+    ('H', 60*60),
+    ('m', 60)),
+  'msec': (
+    ('Y', 60*60*24*365*1000),
+    ('M', 60*60*24*30*1000),
```
SEJeff: Isn't that kind of cheating? Not all months are 30 days.

Simage: Probably a bit, But since we're talking relative times on the y axis it 
should be good enough. I guess it could use weeks rather than months, 
unless you have an alternate suggestion..

SEJeff:
What do you think @brutasse or @graphite-project/committers? Is putting a 30 day "month" alright? Other than that small niggle, I'm happy with this patch. Just want someone else's take before merging it.

obfuscurity:
:+1:

brutasse:
:+1:

SEJeff:
It shall be so.

jazzzz:
Could we have this merged on 0.9.x?

deniszh:
Unfortunately, this patch is for master and I really doubt that it could be cherry-picked to 0.9.x - need to be ported.

obfuscurity:
Not to mention that we're not planning to cut any non-maintenance releases from 0.9.x moving forward.

jazzzz:
A cherry pick worked fine.

PR #1221, opened 2015-05-13
fixed HTTP/500 when generating a SVG graph without any data

g76r:
without this patch, generating a SVG graph leads to HTTP/500 instead of an red "no data" image like with PNG


g76r:
Hi, is someone interested in reviewing and/or merging these changes ?

obfuscurity:
@g76r I don't know the SVG code very well; it was written by @tmm1 but he's not really active anymore. Regardless, it seems like you seem to know what you're doing so... :boom::wink:

g76r:
thanks.

Issue #1223, opened 2015-05-14
Exception when fetching metrics from carbon's memory cache


zdenekpizl:
Hallo,

 using master branch of graphite-web, carbon and whisper I am getting exceptions like:

`Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/graphite/render/datalib.py", line 136, in fetchData
    seriesList = _fetchData(pathExpr,startTime, endTime, requestContext, seriesList)
  File "/usr/lib/python2.6/site-packages/graphite/render/datalib.py", line 99, in _fetchData
    fetches = [(node, node.fetch(startTime, endTime)) for node in matching_nodes if node.is_leaf]
  File "/usr/lib/python2.6/site-packages/graphite/node.py", line 30, in fetch
    return self.reader.fetch(startTime, endTime)
  File "/usr/lib/python2.6/site-packages/graphite/readers.py", line 178, in fetch
    for (timestamp, value) in cached_datapoints:
TypeError: 'float' object is not iterable
Thu May 14 12:45:50 2015 :: Failed after 2 retry! See: 'float' object is not iterable
Traceback (most recent call last):
  File "/usr/lib/python2.6/site-packages/graphite/render/datalib.py", line 136, in fetchData
    seriesList = _fetchData(pathExpr,startTime, endTime, requestContext, seriesList)
  File "/usr/lib/python2.6/site-packages/graphite/render/datalib.py", line 99, in _fetchData
    fetches = [(node, node.fetch(startTime, endTime)) for node in matching_nodes if node.is_leaf]
  File "/usr/lib/python2.6/site-packages/graphite/node.py", line 30, in fetch
    return self.reader.fetch(startTime, endTime)
  File "/usr/lib/python2.6/site-packages/graphite/readers.py", line 178, in fetch
    for (timestamp, value) in cached_datapoints:
TypeError: 'float' object is not iterable`

From my point of view these error occur when graphite-web is asking carbon's memory cache to complete data written to disk with data not yet written to disk. It does not matter what metric is it apparently.

Anybody have met this faulty behaviour? Thank you, with regards .zp.

tomasdubec:
This happens only for metrics starting with CARBON_METRIC_PREFIX, because they are handled by send_request_to_all instead of send_request and send_request_to_all is not updated for new interface between graphite-web and carbon (dict instead of list). I will create PR momentarily..

tomasdubec:
[...]

PR #1224, opened 2015-05-14
handle dict in send_request_to_all in carbonlink

tomasdubec:


ttarczynski:
I've just applied this patch on my graphite-web and it fixes the #1223 issue for me.

obfuscurity:
Thank you! :gift:

PR #1236, opened 2015-05-29
Add fallbackSeries function

tmm1:
Renders a fallback metric when no series are provided. This is helpful for preventing "No Data" graphs.

For example, I am using `fallbackSeries(  maximumAbove(queue-*.jobs, 0),   queue-high.jobs  )` to show the job queues with the most jobs (hiding any with 0 jobs), or if all jobs are empty showing a 0 line for just the high queue.


jssjr:
:+1:

PR #1239, opened 2015-06-03
- add support for hiding the x-axis when drawing graphs

leehuk:
Adds support for hiding the x-axis when rendering graphs, extending the graph to cover the full height


obfuscurity:
Looks good, thanks! :sparkles:

PR #1240, opened 2015-06-04
Added missing Quotation marks for timeShift target name.

mondainai:
timeShift was supposed to return a target name like:
timeShift(Sales.widgets.largeBlue,"-7d")
However, due to missing Quotation marks the actual result was:
timeShift(Sales.widgets.largeBlue,-7d)


brutasse:
LGTM

PR #1241, opened 2015-06-05
Added " to time offset in timeShift() series name

goir:
If you request a series with timeshift:
timeShift(my.data.series, "-2d")

Currently timeShift returns the series name without quotes:
timeShift(my.data.series, -2d)

This change adds " around the second parameter so the returned target name is valid
timeShift(my.data.series, \"-2d\")

To reproduce: Request any timeShift and look at the returned json target name.


brutasse:
Closing as a duplicate of #1240

Issue #1259, opened 2015-06-19
Graphite Dashboard: Expanding Node In Tree View Doesn't Update Data


brobotics:
Data is only loaded from an initial selection of the metric type. Instead, the data for a specific node in a tree should be loaded when the specific node is expanded.

To Reproduce:
1. Select Tree Navigation Mode for Dashboard UI
2. Select Metric Type
3. Select category
4. Open root node in tree
5. Add new data under root node
6. Collapse Node
7. Expand Node
8. Observe new data is not present.

To Fix:
1. Reset loaded for node to false before expanding node.

Patch Below:

```
diff --git a/webapp/content/js/dashboard.js b/webapp/content/js/dashboard.js
index 7cdab82..08d480c 100644
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -936,7 +936,12 @@ function metricTextSelectorShow(pattern) {

 function metricTreeSelectorNodeClicked (node, e) {
   if (!node.leaf) {
-    node.toggle();
+    if (node.expanded) {
+      node.collapse();
+    } else {
+      node.loaded = false;
+      node.expand();
+    }
     return;
   }
```

SEJeff:
Please send us a pull request changing the code instead of putting it online in an issue. That way you get credit for if and it is much easier for us maintainers.

PR #1261, opened 2015-06-22
Reset loaded for node to false before expanding node.

brobotics:
Load data for a specific node in a tree when the respective node is expanded.

#1259


SEJeff:
Thanks for taking the time to make graphite better

Issue #1268, opened 2015-06-29
Constrain the metrics available to a specific namespace prefix


jstangroome:
I was running a single server with a single Carbon-Cache and a number of independent Graphite-Web instances, one per customer. With CarbonLink disabled and `WHISPER_DIR` set to each customer's WSP subdirectory, it is trivial to constrain each Graphite-Web instance to only have access to a subset of the metrics.

I have since scaled out to a farm of Carbon-Cache servers, each server with a single corresponding "backend" Graphite-Web instance. Separately there is a number of independent "frontend" Graphite-Web instances, one per customer, but now without any local WSP file access and instead using `CLUSTER_SERVERS` to specify the farm of "backend" Graphite-Web instances to query. However this new architecture requires a new method to constrain the set of metrics that a customer's "frontend" Graphite-Web instance can query.

To achieve this I have implemented a Graphite-Web setting named `REMOTE_STORE_METRIC_PREFIX`. When this setting is specified in a "frontend" instance, its value is prepended to the metric names queried from "backend" instances. Eg, with `REMOTE_STORE_METRIC_PREFIX=customer7.` and a frontend query of `/render?target=requests.count` then the backend query becomes `/render?target=customer7.requests.count`.

The metric names presented to the user, eg in a PNG-rendered chart with a legend, will be the same metric name requested by the user, the prefix will not be revealed.

The `REMOTE_STORE_METRIC_PREFIX` setting does not apply to local WSP file access nor CarbonLink queries. If the setting is not specified, metric names are not modified.

I have implemented this as a patch on the 0.9.x branch. A PR for your consideration will follow.

PR #1269, opened 2015-06-30
Prepend REMOTE_STORE_METRIC_PREFIX to all metrics queried…

jstangroome:
… from upstream Graphite-Web instances. Refs #1268.


jstangroome:
CI needs to be re-run following merge of PR #1273

obfuscurity:
Seems like an idea worth of inclusion to 0.9.x, but would you mind adding it to local_settings.py.example with some comments or documentation?

deniszh:
Hello @jstangroome,
Could you please include some documentation for this feature, e.g. from https://github.com/graphite-project/graphite-web/issues/1268 ?

DanCech:
@deniszh at this point wouldn't we be adding this to master rather than 0.9.x?

deniszh:
@DanCech - if @jstangroome still interested in this feature we can port it to master too.
But I just found out that this thing separates only whisper dirs, with disabled Carbonlink. Not sure if viable solution...

jstangroome:
@DanCech @deniszh We are using this feature in Production very effectively and would like to see it become part of Graphite proper.

I am happy to contribute documentation. Do you feel there is a particular documentation page where this should be added, or should a new page be introduced?

The solution is working with Carbonlink in our deployment, which is as follows:

1. User-Agent requests `graphite/render/?target=some.metric` from Front End Graphite (aka `FEG`)
1. `FEG` is configured with `REMOTE_STORE_METRIC_PREFIX="customer1."` and `CLUSTER_SERVERS=["BEG1:8080", "BEG2:8080"]` where `BEG1` and `BEG2` are Back End Graphite instances.
1. `FEG` prepends `customer1.` to the beginning of `target`, forming `customer1.some.metric` then requests `customer1.some.metric` from `BEG1` and `BEG2`.
1. `BEG1` and `BEG2` do *not* have `REMOTE_STORE_METRIC_PREFIX` defined but *do* have `WHISPER_DIR = '/opt/graphite/storage/whisper'` and `CARBONLINK_HOSTS = ['127.0.0.1:7002']`.
1. `BEG*` both check the disk for `/opt/graphite/storage/whisper/customer1/some/metric.wsp` and query Carbon Cache for `dict(type='cache-query', metric='customer1.some.metric')` and has been confirmed with `tcpdump port 7002`.
1. `BEG*` return all matching data for `customer1.some.metric` to `FEG`.
1. `FEG` strips the `customer1.` prefix from the matching metric paths before returning the result to the User-Agent.

@DanCech is correct that `REMOTE_STORE_METRIC_PREFIX` does not work with Carbonlink when the architecture is reduced to fewer layers (ie User-Agent -> Graphite -> disk,Carbon-Cache) but maybe this is solvable with changes to `carbonlink.py`. It took me some time to grok the code-base just for my scenario so I didn't attempt to go further.

Perhaps documenting the current limitation with respect to Carbonlink in simpler architectures would be sufficient to allow this to be merged? Supporting Carbonlink fully could maybe wait until this capability is requested.

deniszh:
Thanks for explanation, @jstangroome !
IMO that's a nice feature to have. But we have 2 issues:
1) Lack of documentation. I think you need to add your explanation how it works and warning for indirect Carbonlnk support. You can add new page under https://graphite.readthedocs.io/en/latest/config-webapp.html and add link from [local_settings.py explanation](https://graphite.readthedocs.io/en/latest/config-local-settings.html)
2) We can add this feature to 0.9.x, but it will be abandoned very soon, and next version (1.0.0) will be derived from master very soon. So, if you want to make it part of Graphite you need to port it to `master` branch. Is that OK?

jstangroome:
@deniszh I couldn't find the mentioned documentation file in the 0.9.x branch so I used `docs/config-local-settings.rst` instead. I hope that's ok.

I will investigate what is required to port this feature to the master branch but I would appreciate if it can be merged into 0.9.x until then.

deniszh:
I'm 👍 for merging this if no objections.

@DanCech @iksaif @obfuscurity  - how do you feel about final 0.9.16 release, together with 1.0.0?
It will address some nasty bugs in 0.9.15, like https://github.com/graphite-project/graphite-web/issues/1868 and couple of others.
Of course, we should mention that upgrade to 1.0.0 is encouraged, and you shouldn't use 0.9.16 until you have some serious reason to do so, like legacy OS, python etc.

@jstangroome: Thank you for your contribution, but please do not forget to port that change to master, otherwise, it will be gone soon.

DanCech:
At first glance it appears that the way the prefix add/remove code is structured is too permissive, and would cause problems if a user defined a metric that happened to start with the same node(s) as the prefix.

DanCech:
In my opinion, if we're going to support a setting like this it needs to be enforced properly, with the prefix being added to every request and any results returned from the remote node that don't start with the prefix discarded.

jstangroome:
@DanCech Good points.

Continuing my example above, an end-user request for `?target=customer1.some.metric` would propagate to the back-end unaltered but upon return would be presented as just `some.metric`.

Similarly if there was a back-end metric named `customer1.customer1.foo` then the end-user would need to request `?target=customer1.customer1.foo` to access it and upon return it would be presented as just `customer1.foo`.

This behaviour is certainly suboptimal but should still mean that all metrics returned from the back-end are constrained to the subset defined by the prefix configuration even if the returned names don't match the requested names.

To be honest, the original commit for this PR was written so long ago that I don't recall if it was necessary for the prefix to be conditionally prepended and truncated or if it was just because I was uncertain about the how the original code would call my code.

I'll take a stricter approach when I implement this for master. Do you need a stricter approach before merging this PR for 0.9.x?

DanCech:
Closing this, if we want to pursue it we'll need to rebase onto master

PR #1276, opened 2015-07-06
Add unit tests for atttime class

JeanFred:
As I was trying to figure out what was happening in the attime class, I ended up writing a test suite. Although it probably has holes, I believe it is already useful.

All tests are passing, except the one for parsing 'noon+yesterday' as per the Render API documentation, which is skipped.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -22,6 +22,7 @@ deps =
 	pytz
 	py26: mock<1.1.0
 	py27: mock
+	py26: unittest2
 	git+git://github.com/graphite-project/whisper.git#egg=whisper
 	git+git://github.com/graphite-project/ceres.git#egg=ceres
 	pyparsing1: pyparsing==1.5.7
--- a/webapp/tests/test_attime.py
+++ b/webapp/tests/test_attime.py
@@ -1,0 +1,131 @@
+from mock import patch
+import datetime
+try:
+    import unittest2 as unittest
+except ImportError:
+    import unittest
+from django.test import TestCase
+from django.utils import timezone
+from graphite.render.attime import parseTimeReference, parseATTime
+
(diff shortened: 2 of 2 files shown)


SEJeff:
@JeanFred This looks fantastic and I'm *always* a fan of more automated testing, however it fails on [a few of our build targets](https://travis-ci.org/graphite-project/graphite-web/builds/69758063).

Is there any chance you could update the code to work on those build targets? I'd be happy to merge it assuming all goes well.

JeanFred:
@SEJeff Thanks!

Okay, so this is failing on the Python 2.6 builds, as I am using some unittest features only introduced in 2.7
- `assertRaisesRegexp`: I used because the exceptions raised felt too generic to me (Exception, ValueError), but this is not super useful anyway ;
- `skip` in order to skip the 'noon+yesterday' test ; but what I really wanted to do was `expectedFailure`

I have updated the pull request to not use 2.7 only constructs. Let’s see if Travis is happy with it. :)

JeanFred:
Oh, so `expectedFailure` is also a 2.7 construct apparently (I thought from the docs it was not). Shall I just comment out this test @SEJeff ? Or add `unittest2` package to the dependencies ?

SEJeff:
Feel free to update this yet again to make unittest2 a dependency, even in py2.6. It is just that much better and I'm ok with that.

JeanFred:
Sorry for the lag @SEJeff. I updated the pull-request to have unittest2 intstalled in a py26 environment. All tests are now passing :)

SEJeff:
You're truly a wonderful person. If you're ever in Chicago, or at a conference I'm at, I'll buy you some :beer:.

Thanks!

JeanFred:
Deal :)

Issue #1278, opened 2015-07-14
Clustered webapp race condition writing [] to memcache


apg-pk:
Using Graphite 0.9.13 in a 3 node cluster, on initial cache miss for a render request, the correct datapoints are returned, but subsequent requests return an empty array []:

```bash
$ curl "0/render/?salt=1436735833.313&target=servers.cmgm004.GenericJMX-QPID-CLIENT-consumerCount.gauge-Count&format=json&from=-1min"
[{"target": "servers.cmgm004.GenericJMX-QPID-CLIENT-consumerCount.gauge-Count", "datapoints": [[12.0, 1436767950], [12.0, 1436767960], [12.0, 1436767970], [12.0, 1436767980], [12.0, 1436767990], [null, 1436768000]]}]

$ curl "0/render/?salt=1436735833.313&target=servers.cmgm004.GenericJMX-QPID-CLIENT-consumerCount.gauge-Count&format=json&from=-1min"
[]

$ curl "0/render/?salt=1436735833.313&target=servers.cmgm004.GenericJMX-QPID-CLIENT-consumerCount.gauge-Count&format=json&from=-1min"
[]
```

Using noCache query param or turning off memcached appear to be workarounds.

Problem appears to be in: https://github.com/graphite-project/graphite-web/blob/0.9.13-pre1/webapp/graphite/render/views.py#L126-L127

```python
     for target in targets:
        t = time()
        seriesList = evaluateTarget(requestContext, target)
        log.rendering("Retrieval of %s took %.6f" % (target, time() - t))
        data.extend(seriesList)

      if useCache:
        cache.add(dataKey, data, cacheTimeout)
```

If **data** is never populated for any of the targets (because no metrics are on current node), then **data** will stay initialized to **[]** and then added to cache for subsequent requests to get.

deniszh:
Are you using shared memcached? It's known issue then. Please switch to local memcached instance for every node.

deniszh:
Adding empty list to cache is harmless btw, just turn off shared memcached.

apg-pk:
Yes, shared memcached, but unfortunately not harmless. The nodes without the metric are able to insert an empty [] into the cache. The empty list is considered a cache hit  and sent in the response even though the cache log reports a miss. In the code below, the empty list is considered **false** for the conditional but also **not None** for the actual request context:

```python
    if useCache:
      targets = requestOptions['targets']
      startTime = requestOptions['startTime']
[...]

PR #1279, opened 2015-07-14
Fix clustered webapp race condition writing to cache

apg-pk:
Check data not empty before adding to cache. Fix for #1278?


deniszh:
Tried your patch on test environment, works fine, not sure if helps with shared memcached - still didn't enable that in my env.

obfuscurity:
This looks like a good fix to me, merging. Thanks! :zap:

PR #1282, opened 2015-07-17
Travis container infrastructure

brutasse:
This should speed up travis builds. An update of the `mock` library also broke all python2.6 builds.


brutasse:
:shipit:

PR #1287, opened 2015-07-23
Fix bug to create event with the specific when datetime

redice:
failed to send HTTP request with the specific 'when' datetime to create event

  File &quot;/opt/graphite/webapp/graphite/events/views.py&quot;, line 51, in post_event
    values[&quot;when&quot;] = datetime.datetime.fromtimestamp(event.get(&quot;when&quot;, time.time()))
TypeError: a float is required

### DIFF
--- a/webapp/graphite/events/views.py
+++ b/webapp/graphite/events/views.py
@@ -49,7 +49,7 @@ def post_event(request):
         values["what"] = event["what"]
         values["tags"] = event.get("tags", None)
         values["when"] = datetime.datetime.fromtimestamp(
-            event.get("when", time.time()))
+            float(event.get("when", time.time())))
         if "data" in event:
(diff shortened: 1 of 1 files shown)


obfuscurity:
@redice I believe this has been fixed in 0.9.x, at least it [looks that way](https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/graphite/events/views.py#L44-L46) to me.

```
>>> datetime.datetime.utcfromtimestamp(time.time())
datetime.datetime(2015, 9, 27, 2, 21, 28, 594272)

>>> datetime.datetime.utcfromtimestamp(1431607046)
datetime.datetime(2015, 5, 14, 12, 37, 26)
```

If you have reason to believe otherwise please let me know and we can reopen this one.

PR #1291, opened 2015-08-02
Add target to create a new graph from a metric.

cbowman0:
Allow the creation of a graph from a metric.  For example, you can paste in 'sumSeries(carbon.carbon-daemons.\*.cache[0-9]\*.metrics_received)' into the input box and a new graph with be created with that metric.

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -595,6 +595,9 @@ function initDashboard () {
               },
               { text: "From Saved Graph",
                 handler: newFromSavedGraph
+              },
+              { text: "From Metric",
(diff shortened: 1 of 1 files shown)


obfuscurity:
Would you mind posting a gif of the new feature? We've started doing this at work for any UI changes and it works really well. :smile_cat:

cbowman0:
OK.  Sure.

http://www.gfycat.com/GivingKlutzyAsiandamselfly

obfuscurity:
Thanks. Useful feature and I'll go ahead and merge it in. Any thoughts to naming it "From Target" instead of "From Metric"?

Issue #1063, opened 2014-12-22
Port the functionality found in #1026 to master


bmhatfield:
#1026 parallelizes the updated 0.9.x behavior found in #1010. Unfortunately, it is different enough from master that there's some work involve to get this behavior moved up to it. This issue is to track that work :-)

toni-moreno:
Hi @bmhatfield I'll be happy to test it once merged to master.

jjneely:
#1293 has a small improvement for bulk-fetch and also merges duplicate time series together which I'd like to port to master.  But being this work was based on the bulk-fetch code, I'm blocked until that is also ported to master.

obfuscurity:
@bmhatfield I would buy you **so many** :beers: if you would be inclined to push this along. :smile:

deniszh:
:+1:

obfuscurity:
@bmhatfield any chance you'd be willing to pick this work up again? I personally consider this a "must-have" for a 1.0 release.

obfuscurity:
I would be willing to sponsor this work.

cbowman0:
Does this improve upon the fetch call immediately returning with a callback method that is implemented in master?

deniszh:
Maybe, need to test that.

bmhatfield:
Oh dang, I forgot about this bad boy. I am generally interested but don't see myself being able to carve out time to do this right (ie; stand up something clustered running master and trying it out) over the next few weeks. If it's more urgent than that due to me not doing it before, I am happy to review PRs in lieu of actually writing the code.

cbowman0:
I capture network traffic to confirm what I believed to be happening in master right now.

This was run with one graphite-web instance with no local data and 9 servers holding data.  All running the master branch of graphite-web.  The `/metrics/find` requests happen in quick succession without waiting for a response from each server.  The responses are then processed before issuing the `/render` requests to get data, which happen in the same method.  Here's timestamps and action (manually pulled together for readability):

```
15:01:38.771883 IP ui_instance.48944 > datastore_1.http: GET /metrics/find/?local=1&format=pickle&query=carbon.carbon-daemons.%2A.cache%2A.writer.datapoints_written&from=1471531598&until=1471532498 HTTP/1.1
[...]

PR #1293, opened 2015-08-12
datalib.py: Merge TimeSeries when more than one returned per key

jjneely:
During a rebalance of a consistent hash cluster, after a partition event
on a replication > 1 cluster, or in other cases we might receive
multiple TimeSeries data for a metric key.  Merge them together rather
that choosing the "most complete" one.

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -413,14 +413,13 @@ def fetchData(requestContext, pathExpr):
         series_handled = False
         for known in seriesList:
           if series['name'] == known.name:
-            # This counts the Nones in each series, and is unfortunately O(n) for each
-            # series, which may be worth further optimization. The value of doing this
-            # at all is to avoid the "flipping" effect of loading a graph multiple times
-            # and having inconsistent data returned if one of the backing stores has
-            # inconsistent data. This is imperfect as a validity test, but in practice
-            # nicely keeps us using the "most complete" dataset available. Think of it
-            # as a very weak CRDT resolver.
-            candidate_nones = len([val for val in series['values'] if val is None])
+            # This merges multiple series for the same key and is O(n) for
+            # each series, which may be worth further optimization.  The value
+            # of doing this is to avoid partial graphs when replication > 1
+            # and a partitioning event has happened, during a rebalance of
+            # a consistent hashing cluster, or any state where metric data
+            # is in an inconsistent state.  Think of it as a weak CRDT
+            # resolver.
 
             # To avoid repeatedly recounting the 'Nones' in series we've already seen,
             # cache the best known count so far in a dict.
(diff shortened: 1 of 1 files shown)


jjneely:
This is probably crack...but, why not?

jjneely:
Updated to replace the `seriesList` list in the bulk-fetch code with a dict.  Using the dict for large queries provides a significant improvement in performance.

This makes PR #1304 obsolete.

obfuscurity:
Has this / can this be submitted against master too?

obfuscurity:
@jjneely I hate to be a pest, but would you be willing to add a test for this? I'd like to get this into 0.9.x and adding a new test would be enough justification for me to pull the trigger. :wink:

deniszh:
IMO it's a very nice feature for Graphite clustering and it's quite good to have test. Will try to create one tomorrow, to include that in release.

obfuscurity:
@deniszh Would you be willing and/or able to port this to master?

deniszh:
@obfuscurity - I'll try to. My knowlege of master is still scarce. :imp:

jjneely:
Porting to master:  This patch builds on top of the performance improvements and code from the bulk-fetch patches, but those don't seem to be ported to 0.10.x.  

Is there an Issue/Comment to track getting that ported to 0.10.x?

I can probably still port the code to master, but that feels like the cart in front of the horse in this case.

deniszh:
IIRC current bulk-fetch implementation done mostly by @bmhatfield with some improvements by @jraby.
We can open issue for porting that to master first.

jjneely:
That makes the most sense.  

As far as unit tests:  Merging assumes there are duplicate metrics which can only be the case if there are remote metrics.  I don't see anything that serves as a guide for how I might setup unit tests for cluster functionality and I don't want to mock half of graphite.

If we can figure out the framework here, I'll be glad to write the tests.

deniszh:
I found #1010 master-port issue BTW - https://github.com/graphite-project/graphite-web/issues/1063

deniszh:
I'll try to test it on my cluster for weekend. If it works - maybe we can create separate test for that, just for inclusion in upcoming 0.9.14 ? 
We only need to test data merge itself, right - and not any other functionality here - other things are already merged and works.

deniszh:
Tested that on simple 2 node cluster. My lowest retention is 60 seconds.
1. Creating local.random.diceroll2 metric on server1 - with values 1..5:
```shell
$ for x in 1 2 3 4 5; do echo "local.random.diceroll2 ${x} `date +%s`" | nc -q0 127.0.0.1 2103; sleep 60; done
```
After finishing that script run similar on node 2
2. Creating local.random.diceroll2 metric on server2 with values 6..10:
```shell
$ for x in 1 2 3 4 5; do echo "local.random.diceroll2 ${x} `date +%s`" | nc -q0 127.0.0.1 2103; sleep 60; done
```
3. Checking Graph in Grafana:
screen shot 2015-10-11 at 18 47 48
As you can see graph is perfectly merged.
4. Dumping individual values:
Server1:
```shell
$ whisper-dump.py /opt/graphite/storage/whisper/local/random/diceroll2.wsp | grep -A 10 'Archive 0 data'
Archive 0 data:
0: 1444580340,          1
1: 1444580400,          2
2: 1444580460,          3
3: 1444580520,          4
4: 1444580580,          5
5: 0,          0
6: 0,          0
7: 0,          0
8: 0,          0
9: 0,          0
```
Server2:
```shell
$ whisper-dump.py /opt/graphite/storage/whisper/local/random/diceroll2.wsp | grep -A 10 'Archive 0 data'
Archive 0 data:
0: 1444580640,          6
1: 1444580700,          7
2: 1444580760,          8
3: 1444580820,          9
4: 1444580880,         10
5: 0,          0
6: 0,          0
7: 0,          0
8: 0,          0
9: 0,          0
```

I'm advocating for merging it before 0.9.14 release. It's very nice addition to have for graphite clustering, but not very easy to make test for that.
@obfuscurity @bmhatfield @jssjr @mleinart @SEJeff @drawks ?
(sorry in advance for poking you, guys :) )

bmhatfield:
This looks great to me. It took me a few moments to line up all the behaviors here, but I believe we ought to see a performance improvement and have merged series, which seems wonderful to me.

obfuscurity:
You all realize the operational implications here, right? And I mean that in a good way. :smile: 

Before merging this, can anyone think of any situations where this breaks a "relied on" behavior? I can't think of any but I just woke up from a nap so my neurons aren't firing at :100: yet.

deniszh:
Hm. I can do that thing optional, as I did for cabon's https://github.com/graphite-project/carbon/pull/135, for example. Still thinking that's need to be enabled by default, but it's nice to have a trigger to turn that off.

obfuscurity:
I think this is a significant enough change to the existing (admittedly sucky but still "as intended") design that it warrants a knob, at least in 0.9.x. And yes, we should enable it by default. Thanks @deniszh!

deniszh:
Done, will test tomorrow.
Although it's lot looks more clear now... :imp:

deniszh:
Closing in the favor of https://github.com/graphite-project/graphite-web/pull/1352

obfuscurity:
refs #1063

PR #1295, opened 2015-08-16
Backport of #1188 "Serve staticfiles in the webapp if whitenoise is installed" to 0.9.x

deniszh:
Whitenoise is compatible with Django 1.4 and above, but before version 2.0.2 it was compatible only with python 2.7, so in this version safecheck is included.

Tested with synthesize and manual whitenoise installation, but please check more.

Rebase of https://github.com/graphite-project/graphite-web/pull/1209

### DIFF
--- a/check-dependencies.py
+++ b/check-dependencies.py
@@ -157,6 +157,14 @@ except:
   print "Note that txamqp requires python 2.5 or greater."
   warning += 1
 
+# Test for whitenoise
+try:
+  import whitenoise
+except ImportError:
+  print "[INFO]"
+  print "Unable to import the 'whitenoise' module."
+  print "This is useful for serving static files."
+
 
 if fatal:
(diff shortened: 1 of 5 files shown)


deniszh:
@SEJeff - rebased https://github.com/graphite-project/graphite-web/pull/1209, could you please check?

obfuscurity:
This looks ok to me, @brutasse any opinions?

brutasse:
@obfuscurity LGTM, but I haven't touched 0.9.x in ages :)

obfuscurity:
@deniszh Are the benefits worth the added complexity to 0.9.x? If we can't enforce a minimum version of whitenoise (because of python/django dependencies and conflicts) in setup.py, I wonder if many users will know to take advantage of it.

obfuscurity:
Bah, humbug.

deniszh:
Latest whitenoise supports 2.6, it's just more, like, precaution. IMO will help in some cases.
:+1:
Thanks!

Issue #1297, opened 2015-08-18
Weighted average fails when wrapped in another function


pvanassen:
When executing the following command: 
`movingAverage(weightedAverage(test.data.*.all, test.data.*.weight, 5), 5)`
the following error is thrown: 
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/django/core/handlers/base.py", line 112, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python2.7/dist-packages/graphite/render/views.py", line 110, in renderView
    seriesList = evaluateTarget(requestContext, target)
  File "/usr/lib/python2.7/dist-packages/graphite/render/evaluator.py", line 10, in evaluateTarget
    result = evaluateTokens(requestContext, tokens)
  File "/usr/lib/python2.7/dist-packages/graphite/render/evaluator.py", line 21, in evaluateTokens
    return evaluateTokens(requestContext, tokens.expression)
  File "/usr/lib/python2.7/dist-packages/graphite/render/evaluator.py", line 29, in evaluateTokens
    return func(requestContext, *args)
  File "/usr/lib/python2.7/dist-packages/graphite/render/functions.py", line 739, in movingAverage
    bootstrapSeconds = max([s.step for s in seriesList]) * int(windowSize)
AttributeError: 'float' object has no attribute 'step'
```

pvanassen:
PR merged

PR #1298, opened 2015-08-18
Fixed weighted average to allow its result to be used in other methods

pvanassen:
Fix for issue https://github.com/graphite-project/graphite-web/issues/1297


obfuscurity:
Thanks!

PR #1304, opened 2015-08-25
0.9.x use a dict to search for known TimeSeries in webapp

jjneely:
This can't be merged without #1293 unlike what Github says.

However, using a dict here instead of a list give a vast performance increase with large queries.


jjneely:
or I guess this can me merged in lieu of #1293

deniszh:
:+1:
Didn't have such big queries, but will try this patch tomorrow. Thanks!

deniszh:
Yep,  it's better to add this commit to #1293 and close this one.

jjneely:
#1293 is now updated.  Thanks!

PR #1306, opened 2015-08-27
Fix #1305 -- prevent an xAxis config option from being ignored

### LINKED ISSUE #1305: Potential typo in xAxisConfig
https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/graphite/render/glyph.py#L110
https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/graphite/render/glyph.py#L109

they have same "seconds", but other parameters are different.

According to https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/graphite/render/glyph.py#L1349 this part (where code is choosing what config to use) it looks like L109 won't be used at all. Maybe it's a typo?

brutasse:
Value adapted from the next 2 configs.


obfuscurity:
Thanks! :zap:

PR #1208, opened 2015-04-19
Small bugfix for remote fetch threads.

deniszh:
thread.is_alive() need to be called for timeout check, exception is not rising for that - https://docs.python.org/2/library/threading.html#threading.Thread.is_alive
With this fix you will get an error in log when thread starts time-outing.

PR #1307, opened 2015-08-27
0.9.x timeouts buffers

jjneely:
Timeouts were not functioning on my Graphite setup with Python 2.7 and the joining of the fetch threads was hitting the worst case scenario where I was waiting FETCH_TIMEOUT * number of servers.

If we can assume that the thread's HTTP timeout feature actually works then the timeouts on the joins become unneeded.  I also found that I was silently losing data points for queries when the thread didn't join.

This requires Python 2.6.  I'm not sure if that's a planned requirement for 0.9.13 or not.

### DIFF
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -49,8 +49,8 @@ class FindRequest:
     if self.cachedResults:
       return
 
-    self.connection = HTTPConnectionWithTimeout(self.store.host)
-    self.connection.timeout = settings.REMOTE_STORE_FIND_TIMEOUT
+    self.connection = httplib.HTTPConnection(self.store.host,
+        timeout=settings.REMOTE_STORE_FIND_TIMEOUT)
 
     query_params = [
       ('local', '1'),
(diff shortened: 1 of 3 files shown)


bmhatfield:
This looks good to me, but I am not sure if the py26 requirement is OK or not. @obfuscurity ?

deniszh:
:+1:
IIRC Python 2.6 is official requirement for 0.9.13 for long time ago.
But I'm wondering is this patch obsoletes my fix in https://github.com/graphite-project/graphite-web/pull/1208 or not...

deniszh:
If I understand correctly my fix from https://github.com/graphite-project/graphite-web/pull/1208 also helps with that issue.
And still not understand why you remove timeout from https://github.com/jjneely/graphite-web/blob/0.9.x-timeouts-buffers/webapp/graphite/render/datalib.py#L351 ... Timeout in thread will not raise exception, you need to check ```fetch_thread.is_alive()``` explicitly - thats whole point of https://github.com/graphite-project/graphite-web/pull/1208. Maybe we need to merge your fix with mine.

jjneely:
With the timeout's on the joins I was silently loosing data.  The HTTPConnectionWithTimeout wasn't timing out at all for me so long queries ended up hitting the worst case performance there where the timeout was effectively multiplied by the number of servers.

Using the timeout with Python 2.6's code the timeout would work if the server didn't return any data before the timeout was reached.  However, if the client was actively downloading pickle data from the remote server the timeout was not triggered.

For the last case, I decided that the thread was working appropriately and the timeout on the join was causing harm.  We want to timeout the HTTP connection...not the unpicking or other operations that we know are safe to complete.

jjneely:
But yeah, I can tell we are on the same thought track with #1208

obfuscurity:
Python 2.4 is [still the dependency](https://github.com/graphite-project/graphite-web/blob/0.9.x/check-dependencies.py#L9-L11) for 0.9.x. That being the case, and with support for remote store fetch timeouts in 0.9.x, I think this is probably not going to work for 0.9.x. It seems like we have adequate support for remote fetch timeouts in master, so I'm going to close this one for now. Thanks though, @jjneely :sparkles:.

PR #1308, opened 2015-08-27
Enable buffering in HTTPResponse objects

jjneely:
This improves performance of downloading data from the storage backends
quite a bit.

This requires Python 2.7.  I should probably test to see if this API exists before using it. Hmm...

### DIFF
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -78,7 +78,7 @@ class FindRequest:
       self.send()
 
     try:
-      response = self.connection.getresponse()
+      response = self.connection.getresponse(buffering=True)
       assert response.status == 200, "received error response %s - %s" % (response.status, response.reason)
       result_data = response.read()
       results = unpickle.loads(result_data)
@@ -139,7 +139,7 @@ class RemoteNode:
       connection.request('POST', '/render/', query_string)
     else:
       connection.request('GET', '/render/?' + query_string)
-    response = connection.getresponse()
+    response = connection.getresponse(buffering=True)
     assert response.status == 200, "Failed to retrieve remote data: %d %s" % (response.status, response.reason)
     rawData = response.read()
 


deniszh:
Maybe we can create some wrapper or check for fallback on 2.6 ? 
2.6 is still official requirement for 0.9.x branch

webapp/graphite/remote_storage.py
```diff
@@ -139,7 +139,7 @@ def fetch(self, startTime, endTime, now=None, result_queue=None):
       connection.request('POST', '/render/', query_string)
     else:
       connection.request('GET', '/render/?' + query_string)
-    response = connection.getresponse()
+    response = connection.getresponse(buffering=True)
```
deniszh: Here is quite easy fix for 2.6 support:
Maybe do something like
```
  try: # Python 2.7+, use buffering of HTTP responses
    response = connection.getresponse(buffering=True)
  except TypeError: # Python 2.6 and older
    response = connection.getresponse()
```
Assert will catch everything else.
Not sure how to patch line 81 though.

webapp/graphite/remote_storage.py
```diff
@@ -78,7 +78,7 @@ def get_results(self):
       self.send()
 
     try:
-      response = self.connection.getresponse()
+      response = self.connection.getresponse(buffering=True)
```
deniszh: For supporting python 2.6 it can be rewritten like
```
    try:
      response = self.connection.getresponse(buffering=True)
    except TypeError:

    try:
      assert response.status == 200, "received error response %s - %s" % (response.status, response.reason)
...
```

obfuscurity: :100:

obfuscurity:
@jjneely Can you add the proposed changes by @deniszh?

deniszh:
I'll try to do that today.

deniszh:
Just create https://github.com/graphite-project/graphite-web/pull/1330, please review - @jjneely @obfuscurity 
Thanks!

obfuscurity:
Closing this one in lieu of #1330. Thanks @jjneely @deniszh! :beers:

PR #1318, opened 2015-09-09
Add unit tests for parseTimeOffset in atttime class

JeanFred:
Just like for #1276, as I was trying to understand what was going on with `summarize`, I ended up writing a test suite (cc @SEJeff).

### DIFF
--- a/webapp/tests/test_attime.py
+++ b/webapp/tests/test_attime.py
@@ -6,7 +6,7 @@ except ImportError:
     import unittest
 from django.test import TestCase
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


obfuscurity:
I defer to @SEJeff here, even more than normal. I **hate** time code. :trollface:

SEJeff:
I was thinking that this code looked familiar and then just realized I wrote some of those tests when my wife was driving home for turkey day a year or two ago. These look great and if I ever meet you @JeanFred or you're near Chicago, I owe you :beers:

SEJeff:
Feel free to write more, we need them :)

PR #1322, opened 2015-09-23
add optional parameter referenceSeries to transformNull

atnak:
This commit adds an optional third parameter, ~~keySeries~~ referenceSeries, to transformNull().
This argument, if specified, governs that transformation of nulls should not
happen unconditionally on every data interval, but rather, on the condition
that the same data interval in any series of keySeries is also non-null.

The motivation for this functionality stems from the need to differentiate
between nulls that represent "zero" and those that represent "no data".
The parameter keySeries is used to identify the data intervals to which
there are data, so that transformNulls() can transforms nulls in such
intervals as unequivocal zeros.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2633,10 +2633,11 @@ def threshold(requestContext, value, label=None, color=None):
 
 def transformNull(requestContext, seriesList, default=0):
   """
-  Takes a metric or wild card seriesList and an optional value
-  to transform Nulls to. Default is 0. This method compliments
-  drawNullAsZero flag in graphical mode but also works in text only
-  mode.
+  Takes a metric or wildcard seriesList and replaces null values with the value
+  specified by `default`.  The value 0 used if not specified.  This method
+  compliments the drawNullAsZero function in graphical mode, but also works in
+  text-only mode.
+
   Example:
 
   .. code-block:: none
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2661,10 +2661,9 @@ def transformNull(requestContext, seriesList, default=0):
 
 def isNonNull(requestContext, seriesList):
   """
-  Takes a metric or wild card seriesList and counts up how many
-  non-null values are specified. This is useful for understanding
-  which metrics have data at a given point in time (ie, to count
-  which servers are alive).
+  Takes a metric or wildcard seriesList and counts up the number of non-null
+  values.  This is useful for understanding the number of metrics that have data
+  at a given point in time (i.e. to count which servers are alive).
 
   Example:
 
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2631,12 +2631,15 @@ def threshold(requestContext, value, label=None, color=None):
 
   return [series]
 
-def transformNull(requestContext, seriesList, default=0):
+def transformNull(requestContext, seriesList, default=0, referenceSeries=None):
   """
   Takes a metric or wildcard seriesList and replaces null values with the value
-  specified by `default`.  The value 0 used if not specified.  This method
-  compliments the drawNullAsZero function in graphical mode, but also works in
-  text-only mode.
+  specified by `default`.  The value 0 used if not specified.  The optional
+  referenceSeries, if specified, is a metric or wildcard series list that governs
+  which time intervals nulls should be replaced.  If specified, nulls are replaced
+  only in intervals where a non-null is found for the same interval in any of
+  referenceSeries.  This method compliments the drawNullAsZero function in
+  graphical mode, but also works in text-only mode.
 
   Example:
 
@@ -2647,14 +2650,25 @@ def transformNull(requestContext, seriesList, default=0):
   This would take any page that didn't have values and supply negative 1 as a default.
   Any other numeric value may be used as well.
   """
-  def transform(v):
-    if v is None: return default
+  def transform(v, d):
+    if v is None: return d
     else: return v
 
+  if referenceSeries:
+    defaults = [default if any(v is not None for v in x) else None for x in izip(*referenceSeries)]
+  else:
+    defaults = None
+
   for series in seriesList:
-    series.name = "transformNull(%s,%g)" % (series.name, default)
+    if referenceSeries:
+      series.name = "transformNull(%s,%g,referenceSeries)" % (series.name, default)
+    else:
+      series.name = "transformNull(%s,%g)" % (series.name, default)
     series.pathExpression = series.name
-    values = [transform(v) for v in series]
+    if defaults:
+      values = [transform(v, d) for v, d in izip(series, defaults)]
+    else:
+      values = [transform(v, default) for v in series]
     series.extend(values)
     del series[:len(values)]
   return seriesList


obfuscurity:
All nulls represent "no data". Under what circumstance would a null represent zero?

atnak:
Let's say I toss a six-sided dice.  The result of the toss will be a number between 1 to 6.

Now, the dice is actually weighted, and the chance of throwing a one is much higher than the chance of throwing any other number.  After 100 tosses, the results are 66 ones, no twos, 24 threes, 10 fours and no fives or sixes.

I'm going to put the result into graphite:

```
face.one.count 66 t=1
face.three.count 24 t=1
face.four.count 10 t=1
```

Notice how I didn't include two, five or six?  I'll get back to why I did that in a minute, but it should be implicit that two, five, and six are *zeroes* at `t=1`.  The rule follows that, **at any `t=T`, the existence of any `face.X.count N` implies there also exists `face.Y.count 0` for all possible faces that aren't explicitly represented at time `T`**.

Getting back to the dice, I actually have thousands them.  They each have their own weights, and make few hundred tosses a second (summed minutely).  Moreover, they're all 10 sided.

```
dice.1.face.one.count 66 t=1
dice.1.face.three.count 24 t=1
dice.1.face.four.count 10 t=1
dice.2.face.two.count 78 t=2
dice.2.face.four.count 90 t=2
...
```

The reason number 1 of why I don't store explicit zeroes at every time `T` is because, if I filled zeroes for every face for every time I took a summary, the order of magnitude increase in data will swamp carbon and cripple my system by the IO involved. (I tried a few times and ultimately gave up.)

I need the zeroes there however, whether it be from `transformNulls()`, because they're fed into funky things like `derivative()`.  Also, graphs with *connected lines* plot better with proper zeros.

The reason number 2 of why I don't (or rather can't) store explicit zeros is because, the total number of dies and faces increases dynamically over time.  I'll be extremely tedious to go back and explicitly add historical zeroes every time I add a new face or die, just so that I can retrospectively treat their intervals as zeroes .

obfuscurity:
@atnak dare I ask, are you using Graphite to study DNA strands?

atnak:
> @atnak dare I ask, are you using Graphite to study DNA strands?

No, but the example was probably too abstract.  Here's another attempt.

Let's turn the analogy into web servers, HTTP requests, and HTTP status codes.

Let's say I have a server that accepts thousands of requests per minute, and I'm going to feed graphite the count of responses, summarized per status code, to plot minutely.

```
server.1.status.200 1595 t=1
server.1.status.202 37 t=1
server.1.status.503 12 t=1
server.1.status.301 407  t=1
...
```

Here's a tabular representation. (t=1 is 9:20)

server 1 | 9:20 | 9:21 | 9:22 | 9:23
-----------|-------|-------|-----|---
**200** | 1595 | 1634 | 1612 |
**202** | 37 | 41 | 10 |
**503** | 12 | 16 | 17 |
**301** | 407 | 398 | 393 |
**500** |  | 2 | 1 |
**404** |  |   | 1 |

* Notice there're no values at 9:23.  That's the *no data* zone.  It's the time interval for which server logs haven't been summarized and fed into graphite yet.  They shouldn't be plot as their values are unknown.  (These zones can also be in the past, due to missing logs.)
* Notice also how there're no values for *500* and *404* at 9:20 and 9:21.  Those are the *should be treated as zero* zones.  We know they're zeros because there exists other values at those times.

Obviously, it's infeasible to feed graphite explicit zeros for every other conceivable HTTP status code at every single time.

Here's example usage:

```
transformNulls(server.1.status.*, 0, server.1.status.*)
derivative(transformNulls(server.*.status.*, 0, server.*.status.*))
```

obfuscurity:
Oh, I didn't need another example, I was just curious. At this point I'm trying to decide if this is a feature unique to a single user (you) or if it's going to be generally applicable to others.

webapp/graphite/render/functions.py
```diff
@@ -2646,14 +2648,22 @@ def transformNull(requestContext, seriesList, default=0):
...
+    if v is None: return d
     else: return v
 
+  if keySeries:
+    defaults = [default if any(v is not None for v in x) else None for x in izip(*keySeries)]
+  else:
+    defaults = None
+
   for series in seriesList:
     series.name = "transformNull(%s,%g)" % (series.name, default)
```
obfuscurity: Shouldn't we include the `keySeries` in this output, e.g.:
```series.name = "transformNull(%s,%g,%s)" % (series.name, default, keySeries)```

File: webapp/graphite/render/functions.py [hunk 2]
```diff
@@ -2631,12 +2631,14 @@ def threshold(requestContext, value, label=None, color=None):
...
   return [series]
 
-def transformNull(requestContext, seriesList, default=0):
+def transformNull(requestContext, seriesList, default=0, keySeries=None):
   """
-  Takes a metric or wild card seriesList and an optional value
-  to transform Nulls to. Default is 0. This method compliments
-  drawNullAsZero flag in graphical mode but also works in text only
-  mode.
+  Takes a metric or wild card seriesList and an optional value to transform
```
obfuscurity: Some grammatical typos here. Some are inherited from the previous version, some are new. Would you mind fixing these?

`wild card` should be `wildcard`

webapp/graphite/render/functions.py
```diff
@@ -2631,12 +2631,14 @@ def threshold(requestContext, value, label=None, color=None):
...
+  Nulls to. Default is 0. The optional keySeries, if specified, is a metric
+  or wild card series list that governs which data intervals nulls should be
```
obfuscurity: `wildcard seriesList` and `Nulls`

webapp/graphite/render/functions.py
```diff
@@ -2631,12 +2631,14 @@ def threshold(requestContext, value, label=None, color=None):
...
+  transformed. If specified, nulls are transformed only in intervals where
```
obfuscurity: `Nulls`

webapp/graphite/render/functions.py
```diff
@@ -2631,12 +2631,14 @@ def threshold(requestContext, value, label=None, color=None):
...
 
-def transformNull(requestContext, seriesList, default=0):
+def transformNull(requestContext, seriesList, default=0, keySeries=None):
   """
-  Takes a metric or wild card seriesList and an optional value
-  to transform Nulls to. Default is 0. This method compliments
-  drawNullAsZero flag in graphical mode but also works in text only
-  mode.
+  Takes a metric or wild card seriesList and an optional value to transform
+  Nulls to. Default is 0. The optional keySeries, if specified, is a metric
```
obfuscurity: Let's switch `Nulls` to `nulls` for consistency with the rest of the document.

webapp/graphite/render/functions.py
```diff
@@ -2631,12 +2631,14 @@ def threshold(requestContext, value, label=None, color=None):
...
+  or wild card series list that governs which data intervals nulls should be
+  transformed. If specified, nulls are transformed only in intervals where
+  the same interval in any of keySeries is non-null. This method compliments
+  drawNullAsZero flag in graphical mode but also works in text only mode.
```
obfuscurity: `drawNullAsZero flag` to `the drawNullAsZero function` and `text only` to `text-only`

obfuscurity:
Aside from those small issues, LGTM.

atnak:
Thanks for the feedback! I'll look them over as soon as I get back home at the end of the week.

I'm a bit on the fence about the term `keySeries`.. Maybe `referenceSeries` might be more self-documenting.

Also, more I look at the second example, more I wonder if there should be some consideration to grouping (like sumSeriesWithWildcard et al.)..  Though I suppose this could be done by someone else if they feel the itch.. (Maybe its possible w/ groupByNode())

webapp/graphite/render/functions.py
```diff
@@ -2646,24 +2650,34 @@ def transformNull(requestContext, seriesList, default=0):
...
 
+  if referenceSeries:
+    defaults = [default if any(v is not None for v in x) else None for x in izip(*referenceSeries)]
+  else:
+    defaults = None
+
   for series in seriesList:
-    series.name = "transformNull(%s,%g)" % (series.name, default)
+    if referenceSeries:
+      series.name = "transformNull(%s,%g,%s)" % (series.name, default, referenceSeries)
```
atnak: This needs testing for verboseness

atnak:
Need to test the `series.name` part when I get into work, but fixes aside from that.

atnak:
#### With regards to `series.name`...

Made it:
* when `referenceSeries` is not specified:
  `transformNulls(server.1.status.202,0)`

* when `referenceSeries` is specified:
  `transformNulls(server.1.status.202,0,referenceSeries)`

i.e. The word "referenceSeries" is printed.

This could be changed to something like:

```python
series.name = "transformNull(%s,%g,[%s])" % (series.name, default, ','.join(x.name for x in keySeries))
```

which would output:

```
transformNulls(server.1.status.202,0,[server.1.status.200,server.1.status.202,server.1.status.503,server.1.status.301])
```

But it's pretty verbose and I don't know how useful this is.. What do you think? @obfuscurity

PR #1324, opened 2015-09-23
Add LDAP user template.

HeMan:
This allows graphite to authorize to LDAP with the user supplied in the login
window instead of static user, similar to how Django LDAP does it.

### DIFF
--- a/docs/config-local-settings.rst
+++ b/docs/config-local-settings.rst
@@ -267,6 +267,13 @@ LDAP_USER_QUERY
   Sets the LDAP query to return a user object where ``%s`` substituted with the
   user id. E.g. ``(username=%s)`` or ``(sAMAccountName=%s)`` (Active Directory)
 
+LDAP_USER_DN_TEMPLATE:
+  `Default: ''`
+
+  Is this is set the LDAP query is authenticated with the usename and passowrd
+  supplied from the user.
+
+
 
 Other Authentications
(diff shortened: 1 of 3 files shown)


obfuscurity:
As someone who's blessedly ignorant of LDAP, I don't understand the advantages/disadvantages of this approach vs setting `LDAP_BASE_USER` and `LDAP_BASE_PASS`. Can we add any more comments to the `local_settings.py.example` that will help users?

Otherwise, this looks benign.

HeMan:
I've updated the documentation somewhat. Should I add a short note in local_setting.py.example as well?

obfuscurity:
Yes please, that would be helpful.

HeMan:
Like that?

obfuscurity:
Yup, could you just fix the spelling typo ("templat") and move it above the line? It's ok to add an extra line above for separation, e.g.

```
#LDAP_USER_QUERY = "(username=%s)"  #For Active Directory use "(sAMAccountName=%s)"
#
# User DN template to use for binding (and authentication) against the LDAP server. 
# %(username) is replaced with the username supplied at login.
#LDAP_USER_DN_TEMPLATE = "CN=%(username)s,OU=users,DC=mycompany,DC=com"
```

HeMan:
There!

obfuscurity:
Thanks!

PR #1326, opened 2015-09-27
Updated about GapLabs using Graphite..

kkdk5535:
At GapLabs, a division of Gap Inc. we use Graphite as our key APM stack, so I am proud to showcase that info in this page that GapLabs uses Graphite as their APM tool.


obfuscurity:
Nice, thanks!

PR #1330, opened 2015-09-27
Enable buffering in HTTPResponse objects (with python 2.6 fallback)

deniszh:
Same as https://github.com/graphite-project/graphite-web/pull/1308 but with python 2.6 fallback


obfuscurity:
@jjneely Thanks for the patch in #1308, and thanks to @deniszh for making it 2.6 friendly here.

PR #1335, opened 2015-09-30
fix: name 'fatal' is not defined

lfckop:


obfuscurity:
Good catch, looks like a botched copypasta from the 0.9.x branch. Maybe. :smile:

PR #1337, opened 2015-10-01
Never merge CarbonLink results with Whisper rollups

obfuscurity:
We shouldn't merge CarbonLink results with Whisper results if the latter is pulled from a rollup archive. Proposed fix was merged to master in #1002, this is an attempt to port the same logic to 0.9.x.

/cc @penpen @jssjr @deniszh

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -464,8 +464,6 @@ def mergeResults(dbResults, cacheResults):
 
   if not dbResults:
     return cacheResults
-  elif not cacheResults:
-    return dbResults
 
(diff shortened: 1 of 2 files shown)


obfuscurity:
Yes, this PR is indeed #1337. :wink:

deniszh:
LGTM!
Will put it to our testing cluster and check how it affects performance.

deniszh:
Just did that - works fine, no signs of performance problems and graphs still looks sane. :smiling_imp:

obfuscurity:
Awesome, thanks for testing. This should resolve a lot of issues. :smile_cat:

PR #1338, opened 2015-10-01
Remove duplicate test import of pytz

JeanFred:
It is already tested on line 72

Follow-up to #1335


JeanFred:
(I caught this by flake8-ing this file. We should enable flake8 for that file too (although it contains quite a lot of legitimate F401))

Issue #863, opened 2014-08-22
Test coverage for 0.10


obfuscurity:
Now that we have a decent setup for testing with `tox` and TravisCI, we should look at filling in the missing coverage of tests, particularly in `render/functions.py`. Consider this an umbrella issue for tracking any missing tests.

/cc @brutasse @esc @SEJeff

JeanFred:
We now have https://codecov.io/github/graphite-project/graphite-web so I’m closing this.

We may want to add more bells & whistles (like posting on PRs etc.) but that can come later.

PR #1342, opened 2015-10-04
Add test coverage support

JeanFred:
- Run tests through coverage
- Get Travis-CI to upload them to CodeCov

Coverage will be computed for all test targets thanks to tox.

[See results on CodeCov](https://codecov.io/github/JeanFred/graphite-web)

(I’m usually going for Coveralls but as I quite impressed by CodeCov I thought I would give it a try. I think it’s a good pick but happy to change either way).

Later on we can look at adding some more bells and whistles like badge in ReadMe, posting on Pull Requests and so on :)

See #863

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,7 @@ install:
 
 script:
   - tox -e $TOXENV
+
+after_success:
+  - pip install codecov
+  - codecov
--- a/tox.ini
+++ b/tox.ini
@@ -15,8 +15,11 @@ setenv =
 	GRAPHITE_NO_PREFIX=true
 changedir = webapp
 commands =
-	django-admin.py test
+	coverage run --branch manage.py test
+	coverage xml --include=graphite/*
+	coverage report --include=graphite/*
 deps =
+	coverage
 	cairocffi
 	django-tagging>=0.3.5,<0.4
 	pytz


obfuscurity:
Anything that adds test coverage sounds good to me but I'm pretty ignorant of Graphite's test innards. Looking for input from @SEJeff @brutasse.

SEJeff:
I suspect seeing th coverage will just lead me to drink... More. This is totally fine by me but we are already using coveralls on whisper, so I'm not sure if we should switch them both or just use a single service.

JeanFred:
Having it all consolidated under one service is certainly the best plan. As I said, I’m happy to change that to Coveralls.

That being said, CodeCov appears more polished & feature-complete − subjectively nicer reports ; but especially branch coverage (which is a pretty big miss from coveralls, see lemurheavy/coveralls-public#31).

obfuscurity:
@SEJeff I defer to you and @brutasse here in matters of python testing. Please advise.

JeanFred:
Gently upping this. :)
(I just noticed that Sentry also uses CodeCov − see eg https://github.com/getsentry/sentry/pull/2345#issuecomment-158218706)

brutasse:
Sorry for the wait… :+1: from me!

obfuscurity:
@JeanFred If we merge this in, can we count on you to update Carbon and Whisper to use CodeCov too?

SEJeff:
Just remove th coveralls bits in whisper. Coveralls is a bit annoying

JeanFred:
@obfuscurity Definitely! I’ll submit PRs to Carbon and Whisper in the coming weeks.

@SEJeff Just for clarification (we can take that further on the Whisper issue), do you want me to remove Coveralls in Whisper and replace it by CodeCov, or you’d rather not have any such tool on Whisper ?

SEJeff:
Seems like if we like codecov, we can just switch everything to it and only it.

PR #1343, opened 2015-10-04
Remove Twisted dependency (0.9.x)

obfuscurity:
We continue to trip over mismatched Twisted dependencies in our requirements (vs Carbon). Considering that Twisted **isn't even needed** for graphite-web, it seems stupid to keep this intact. Testing this against 0.9.x for now.

PR #1344, opened 2015-10-04
Remove Twisted dependency (master)

obfuscurity:
Port of #1343 to master. /cc @deniszh


deniszh:
Master looks much cleaner then 0.9.x...
:+1:

PR #1351, opened 2015-10-12
Remove mock from requirements.txt

JeanFred:
mock is a test dependency and as such should not live in the requirements.txt file.
Its usage is directly specified in tox.ini

This was introduced in 64cf924 (#1157)


brutasse:
:+1:

JeanFred:
That was fast − thanks @brutasse!

(By the way, I did all my recent few PRs against master − should I backport all or part to 0.9.x ?)

PR #1355, opened 2015-10-16
Remove install section on RTD

brutasse:
This removes the permission error caused by RTD's build environment changes.


gwaldo:
:+1: Let's try it

PR #1364, opened 2015-10-19
Add Natural Sorting to sortByName function (untested)

sylr:
Natural sorting allows names containing numbers to be sorted more naturally, e.g:

Normal sort: server1, server11, server12, server2
Natural sorting: server1, server2, server11, server12

To do so we left pad all number sequences within a name with a variable number of
zeros depending on the length of the sequence :

 - server1  -> server0000000001
 - server11 -> server0000000011
 - server2  -> server0000000002

Signed-off-by: Sylvain Rabot <sylvain.rabot@f-secure.com>

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -1902,16 +1902,28 @@ def limit(requestContext, seriesList, n):
   """
   return seriesList[0:n]
 
-def sortByName(requestContext, seriesList):
+def sortByName(requestContext, seriesList, natural = False):
   """
   Takes one metric or a wildcard seriesList.
-
-  Sorts the list of metrics by the metric name.
+  Sorts the list of metrics by the metric name using either alphabetical order or natural sorting.
+  Natural sorting allows names containing numbers to be sorted more naturally, e.g:
+  - Alphabetical sorting: server1, server11, server12, server2
+  - Natural sorting: server1, server2, server11, server12
   """
-  def compare(x,y):
+  def paddedName(name):
+    return re.sub("(\d+)", lambda x: "{0:010}".format(int(x.group(0))), name)
+
+  def compare(x, y):
     return cmp(x.name, y.name)
 
-  seriesList.sort(compare)
+  def natSortCompare(x, y):
+    return cmp(paddedName(x.name), paddedName(y.name))
+
+  if natural:
+    seriesList.sort(natSortCompare)
+  else:
+    seriesList.sort(compare)
+
   return seriesList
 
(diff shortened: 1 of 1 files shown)


obfuscurity:
Nice idea but I think it would make more sense to extend [sortByName()](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L1905-L1915) to support an optional arg (e.g. `natural=false`) rather than an entirely new function. What do you think?

sylr:
@obfuscurity If you think it would be more coherent I'm totally fine with it.

I however am not very familiar with graphite code so I don't really know what needs to be done to extend sortByName().

deniszh:
@sylr :
You can check example of boolean function parameters for example at [hitCount()](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/functions.py#L3087) function.
So, it could be something like
```python
def sortByName(requestContext, seriesList, natural = False):
  """
  Takes one metric or a wildcard seriesList.
  Sorts the list of metrics by the metric name.
  Natural sorting allows names containing numbers to be sorted more naturally, e.g:
     Normal sorting: server1, server11, server12, server2
     Natural sorting: server1, server2, server11, server12
  """
  if natural:
    def natName(name):
       return re.sub("(\d+)", lambda x: "{0:010}".format(int(x.group(0))), name)
    def compare(x,y):
       return cmp(natName(x.name), natName(y.name))
  else:
    def compare(x,y):
      return cmp(x.name, y.name)

  seriesList.sort(compare)
  return seriesList
```
Just an example, it can be written more efficient and elegant ofc...

sylr:
OK thanks I'll provide another patch soon.

sylr:
Any comments ?

deniszh:
Doc build still failed:
```Warning, treated as error:
/home/travis/build/graphite-project/graphite-web/webapp/graphite/render/functions.py:docstring of graphite.render.functions.sortByName:4: ERROR: Unexpected indentation.
``` in https://travis-ci.org/graphite-project/graphite-web/jobs/86390850

sylr:
Fixed.

obfuscurity:
LGTM!   :sparkles: :zap: :+1: :icecream: 

Default/normal sort (`natural=False`):


Natural sort (`natural=True`):

addshore:
I would love this to magically make its way into 9.15!!!
It would be a great new years present for me :D

obfuscurity:
@addshore see https://github.com/graphite-project/graphite-web/pull/1414

PR #1371, opened 2015-10-27
Add ipv6 support in is_local_interface function

miguelxpn:


webapp/graphite/util.py
```diff
@@ -72,13 +72,20 @@ def getProfileByUsername(username):
...
     return None
 
-
 def is_local_interface(host):
+  is_ipv6 = False
   if ':' in host:
-    host = host.split(':',1)[0]
+    try:
+      socket.inet_pton(socket.AF_INET6, host)
+      is_ipv6 = True
```
obfuscurity: Don't we need to manipulate `host` here for ipv6?

miguelxpn: Not really, if host is already a valid ipv6 address then we don't need to manipulate it further, if it's not then socket.error will be raised and we'll split the port out of it and then everything will proceed as it did before, should work unless there's something I'm missing.

miguelxpn:
Turns out I had actually missed something, I changed the code and added support for bracket notation so that the function works with ipv6 addresses with a port. Hopefully that's it now.

webapp/graphite/util.py
obfuscurity: Ok, I'm trusting you here since I have little practical ipv6 experience to compare against. :smile_cat:

obfuscurity:
Thanks!

PR #1381, opened 2015-11-04
adding 'alternate storage backends' section to the tools doc

gwaldo:
Adding influx is not _just_ a :trollface: for @obfuscurity

### DIFF
--- a/docs/tools.rst
+++ b/docs/tools.rst
@@ -268,6 +268,7 @@ Other
 .. _collectd-graphite: https://github.com/joemiller/collectd-graphite
 .. _collectd-to-graphite: https://github.com/loggly/collectd-to-graphite
 .. _Collectl: http://collectl.sourceforge.net
+.. _Cyanite: http://cyanite.io/
 .. _D3.js: http://mbostock.github.com/d3
 .. _Descartes: https://github.com/obfuscurity/descartes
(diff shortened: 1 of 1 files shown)


bmhatfield:
I'm in favor of this change. I know @obfuscurity is not a fan of all of these, but I think it's still OK and good to mention that the backends exist.

I also saw a comment where someone was using this: https://github.com/InMobi/level-tsd - is it worth including?

obfuscurity:
It's probably fine to capitalize Ceres. Even the project README capitalizes it throughout.

gwaldo:
I'm fine with including other backends.  This was just a list off of the few that I'd had in recent memory and seemed most-common alternates (for some value of "most").

gwaldo:
Good point, @obfuscurity Thanks!

deniszh:
>I also saw a comment where someone was using this: https://github.com/InMobi/level-tsd - is it worth including?

Yep, yesterday I saw question in Graphite-dev mailing list about it - it's a first time when I saw person using that. Maybe need to include to doc though...

obfuscurity:
Going to merge this in its current state. level-tsd looks like it's not under active development and even specifies that you have to use the megacarbon branch.

gwaldo:
Thanks!
I also backported this in https://github.com/graphite-project/graphite-web/pull/1382

Issue #1374, opened 2015-10-28
Moving average Winter time change


floppy84:
Hi,

I open this ticket because i got an issue or a misunderstanding on how to.

So i graph a value during time on first query.
On second query i graph the same data but with a timeshift of 7d 14d 21d 28d ::

![image](https://cloud.githubusercontent.com/assets/3499449/10790311/cf2416e6-7d82-11e5-98e2-a9b9be175291.png)

Until this week end all worked fine. but until the change of time (French Hour), i got a gap on my graphe like you see at the end of graphe ::

![image](https://cloud.githubusercontent.com/assets/3499449/10790293/ba6f9c5c-7d82-11e5-8777-e148dd894fdc.png)

My question is, is there a way to manage this change of time.

Thanks
regards

deniszh:
Similar issue - https://github.com/graphite-project/graphite-web/issues/1024
Not much to add here, at least from my side.

obfuscurity:
What happens if you use a timeshift in terms of minutely units, e.g. 10080 mins instead of 7d?

floppy84:
Hi thank you for your answer, i got the same issue if i set timeshift in minute ::

![image](https://cloud.githubusercontent.com/assets/3499449/10819581/2776905a-7e49-11e5-83f2-c9a08395305f.png)

deniszh:
Unfortunately no good solution exist for that bug currently - except running your servers in UTC (which is easy if you doing that from start, but quite hard if you need to migrate there).
Even if we (over)complicate [timeOffset()](https://github.com/graphite-project/graphite-web/blob/c542449047d53e7c05d4cc43cfd0e36f9718ca44/webapp/graphite/render/attime.py#L120) function to check DST boundaries it will work only when both ("from" and "to") boundaries is after or before switch time. 
I see no way how we can fix graph above - timeShift() value must be the same for all graph values, it can't be 10080m before DST moment, and 10020 / 10140 after, on same graph....

floppy84:
I don't need to mgrate data, i will set my server in UTC , and next time change in summer i will not have same issue.
 Thanks for your help

peterstory:
Would it be possible to add another argument to timeShift, toggling whether DST-compensation is applied? It seems like a common use case: using timeShift(-1w) to compare metric values at 7am this week to values at 7am last week... not values at 7am to values at 6am. 

[...]

PR #1384, opened 2015-11-06
Added timeShift option to compensate for DST

peterstory:
See issue #1374. Currently, because ``timeShift`` offsets time by a fixed number of seconds, comparing a time period with DST to a time period without DST, and vice-versa, results in an apparent misalignment. For example, 8am might be overlaid with 7am. To compensate for this, I added the ``alignDST`` option to the ``timeShift`` function. This option is disabled by default. 

``alignDST`` is basically a shortcut for adding +/- 1 hour to the time shift. It only takes affect if the from and until times were both in DST, and the timeShift-ed time was outside DST (and vice-versa).

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2465,7 +2465,7 @@ def timeStack(requestContext, seriesList, timeShiftUnit, timeShiftStart, timeShi
 
   return results
 
-def timeShift(requestContext, seriesList, timeShift, resetEnd=True):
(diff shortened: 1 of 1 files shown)


obfuscurity:
@deniszh look ok to you?

deniszh:
Yep. It's quite simple to merge, even w/o testing. :+1:

deniszh:
s/testing/test code/, sorry

PR #1393, opened 2015-11-12
Removing unnecessary call to log function when no handler

ShalomCohen:
The function log function is getting called with or without the handler (which is setup based on the settings). The problem is that if LOGGING is set to false and the function is getting called, it will show an error about missing handler.

This was the simplest fix for that, it should apply to 0.9.14 (had issues committing it to the tags/0.9.14 so I'm committing to this branch instead (it should apply to the whole branch anyhow).

### DIFF
--- a/webapp/graphite/logger.py
+++ b/webapp/graphite/logger.py
@@ -63,13 +63,22 @@ class GraphiteLogger:
     return self.exceptionLogger.exception(msg,**kwargs)
 
   def cache(self,msg,*args,**kwargs):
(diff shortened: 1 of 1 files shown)


obfuscurity:
To clarify, this only gets triggered once (per restart) when the code path is hit, and is logged by the webserver (e.g. Apache `error.log`), not the application logs. Seems like a reasonable fix. Only affects 0.9.x.

obfuscurity:
Thanks @ShalomCohen! :sparkles: :boom: :cake:

PR #1401, opened 2015-11-18
Improve axis range and ticmark handling

mhagger:
I've noticed that the handling of the y axis range and the choice of tic marks are often a bit strange. For example:

* For linear axes the "pretty values" are often not all that pretty. I don't think a human would usually choose tic marks like [0, 12.5, 25, 37.5, ...]; one would use [0, 10, 20, 30, ...]. And surely nobody would choose [0, 22.5, 45, 67.5, 90, ...] when [0, 20, 40, 60, 80, ...] is so close by.

* The handling of `yMax="max"` is totally broken.

* There are several places where rounding errors could cause labels to be omitted.

* Logarithmic tic marks are totally broken, too.

So I've completely rewritten the axis-handling code while retaining backwards compatibility of the parameters. The new code restricts the mantissas of tic mark steps to be [..., 0.1, 0.2, 0.5, 1, 2, 5, 10, ...], which (in my opinion) is adequate to give really convenient tic marks without wasting too much space on the edges of the graph. I've rewritten the algorithm for optimizing (step, divisor) for the tic marks from scratch (documented in the docstring for `_LinearAxisTics.chooseStep()`. I've eliminated some places where rounding errors could cause problems. And I have fixed the logarithmic tic marks.

I've tested the new `_AxisTics` classes in isolation to some degree and they seem to work. But I haven't tested them together with the `LineGraph` class because I don't have a test setup. (I think the design is sound but there are surely typos and other minor problems.) If somebody could help me figure out some idiot-proof steps and the bare minimum configuration to run tests locally without becoming a django expert or figuring out how to feed data into the system, that would be really helpful.

### DIFF
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -182,7 +182,7 @@ class _AxisTics:
     elif maxFixed:
       self.minValue = self.maxValue - self.chooseDelta(self.maxValue)
     else:
-      delta = max(abs(self.minValue), abs(self.maxValue))
+      delta = self.chooseDelta(max(abs(self.minValue), abs(self.maxValue)))
       average = (self.minValue + self.maxValue) / 2.0
       self.minValue = average - delta
       self.maxValue = average + delta
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -1665,7 +1665,7 @@ def makeLabel(yValue, yStep=None, ySpan=None, yUnitSystem=None):
     return "%.2f %s" % (float(yValue), prefix)
   if ySpan > 10 or spanPrefix != prefix:
     if type(yValue) is float:
-      return "%.1f %s" % (float(yValue), prefix)
+      return "%.1f %s " % (float(yValue), prefix)
     else:
       return "%d %s " % (int(yValue), prefix)
   elif ySpan > 3:
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -157,6 +157,7 @@ class _AxisTics:
 
   def applySettings(self, axisMin=None, axisMax=None, axisLimit=None):
     if axisMin is not None:
+      self.minRoundable = False
       self.minValue = axisMin
 
     if axisMax is not None:
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -1137,33 +1137,28 @@ class LineGraph(Graph):
       self.yLabelWidth = 0.0
 
   def setupTwoYAxes(self):
-    # I am Lazy.
-    Ldata = []
-    Rdata = []
-
-    Ldata += self.dataLeft
-    Rdata += self.dataRight
-
     # Lots of coupled lines ahead.  Will operate on Left data first then Right.
 
-    missingValuesL = any(None in series for series in Ldata)
-    missingValuesR = any(None in series for series in Rdata)
+    missingValuesL = any(None in series for series in self.dataLeft)
+    missingValuesR = any(None in series for series in self.dataRight)
 
     if self.params.get('drawNullAsZero') and missingValuesL:
       yMinValueL = 0.0
     else:
-      yMinValueL = safeMin( [safeMin(series) for series in Ldata if not series.options.get('drawAsInfinite')] )
+      yMinValueL = safeMin( [safeMin(series) for series in self.dataLeft
+                             if not series.options.get('drawAsInfinite')] )
     if self.params.get('drawNullAsZero') and missingValuesR:
       yMinValueR = 0.0
     else:
-      yMinValueR = safeMin( [safeMin(series) for series in Rdata if not series.options.get('drawAsInfinite')] )
+      yMinValueR = safeMin( [safeMin(series) for series in self.dataRight
+                             if not series.options.get('drawAsInfinite')] )
 
     if self.areaMode == 'stacked':
-      yMaxValueL = safeSum( [safeMax(series) for series in Ldata] )
-      yMaxValueR = safeSum( [safeMax(series) for series in Rdata] )
+      yMaxValueL = safeSum( [safeMax(series) for series in self.dataLeft] )
+      yMaxValueR = safeSum( [safeMax(series) for series in self.dataRight] )
     else:
-      yMaxValueL = safeMax( [safeMax(series) for series in Ldata] )
-      yMaxValueR = safeMax( [safeMax(series) for series in Rdata] )
+      yMaxValueL = safeMax( [safeMax(series) for series in self.dataLeft] )
+      yMaxValueR = safeMax( [safeMax(series) for series in self.dataRight] )
 
     if yMinValueL is None:
       yMinValueL = 0.0
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -1040,20 +1040,20 @@ class LineGraph(Graph):
     missingValues = any(None in series for series in self.data)
     finiteData = [series for series in self.data if not series.options.get('drawAsInfinite')]
 
-    yMinValue = safeMin( [safeMin(series) for series in finiteData] )
+    yMinValue = safeMin(safeMin(series) for series in finiteData)
 
     if yMinValue > 0.0 and self.params.get('drawNullAsZero') and missingValues:
       yMinValue = 0.0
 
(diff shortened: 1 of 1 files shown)


deniszh:
WOW. Didn't check patch yet, just want to say that is look MASSIVE! Much appreciated!
Not sure if I can help with testing... My testing skills are weak. :(

mhagger:
I forgot to mention one thing I'm unsure about...`_LogAxisTics` sets `self.step` to `None` because that parameter is ignored and unneeded. That value later gets copied to `LineGraph.yStep`, which eventually can get incorporated into the output of `Graph.output` [here](https://github.com/mhagger/graphite-web/blob/8af9d1b10a649eb3888199171962f2cd4c45789e/webapp/graphite/render/glyph.py#L722). Does anybody know whether `None` is permissible there, or whether we maybe have to record some sensible fake value as the step for logarithmic axes, or whether maybe `Graph.output` should get smarter about omitting that key/value pair altogether?

mhagger:
I just pushed some more commits that have to do with how the data range and user options are combined, especially in the case that they are inconsistent with each other. This fixes a bug that sometimes the "rounding up" behavior for `maxValue` sometimes made it exceed `axisLimit`.

mhagger:
I've done quite a bit of manual testing of the `AxisTics` classes in isolation, fixed some edge cases, and added more input validation. I think these classes are pretty solid now.

I still haven't done any integration testing because I'm not set up for it. So this still could be terribly broken, though hopefully only in trivial ways (plus possibly the point mentioned [above](https://github.com/graphite-project/graphite-web/pull/1401#issuecomment-158234317)).

I'm going to stop pushing now to give people a chance to look over the changes.

cbowman0:
I have noticed that the Y-Axis has an alignment issue between the 0 and other listed values:
screen shot 2016-05-10 at 2 10 28 pm

mhagger:
@cbowman0, thanks for the testing and the feedback.

I suspect it comes down to a one-character difference that I noticed between two versions of `makeLabel` in the code before my changes: [here](https://github.com/mhagger/graphite-web/blob/34e730f926ed781466b916a49246b7badc142da5/webapp/graphite/render/glyph.py#L1149) vs. [here](https://github.com/mhagger/graphite-web/blob/34e730f926ed781466b916a49246b7badc142da5/webapp/graphite/render/glyph.py#L1324). The discrepancy originated in 53bfcd04389054b439503be21ae6eea9d11fee25. I took it to be a mistake in that commit that the two functions were not adjusted in the same way. I added a space to one version of the function in e1dde2d7815d9ff82c594654a645c04c16585f33 because it looked like the space was forgotten from that version. After that the two versions of the function were identical, so I unified them.

It's easy to take that space back out [here](https://github.com/mhagger/graphite-web/blob/8b5a7358adc2bd1738e837bc3b747f1c11b57388/webapp/graphite/render/glyph.py#L235), and I expect that would fix your problem. What I'm not sure of is whether the *other* version of `makeLabel` really *needed* the space. @nleskiw, you wrote 53bfcd04389054b439503be21ae6eea9d11fee25; can you shed some light on the situation?

cbowman0:
I tried to boil down the differences in label layout before/after this change.

No patch:
master

With patch:
master_with_patch

Existing has some inconsistencies:
1. In graph a between <1 and >=1 values.
2. In graph b when the label has a prefix and when it doesn't.
3. In graph d, the prefix has no spacing after the prefix.

I think there's an opportunity to normalize the spacing between the label and the graph.

With these changes to your makeLabel:

```
--- glyph.py	2016-05-12 14:29:26.415428819 +0000
+++ glyph.py.bak	2016-05-12 14:28:22.361647434 +0000
@@ -226,19 +226,17 @@
   def makeLabel(self, value):
     value, prefix = format_units(value, self.step, system=self.unitSystem)
     span, spanPrefix = format_units(self.span, self.step, system=self.unitSystem)
-    if (prefix != ''):
-      prefix = "%s " % prefix
     if value < 0.1:
       return "%g %s" % (float(value), prefix)
     elif value < 1.0:
       return "%.2f %s" % (float(value), prefix)
     if span > 10 or spanPrefix != prefix:
       if type(value) is float:
-        return "%.1f %s" % (value, prefix)
+        return "%.1f %s " % (value, prefix)
       else:
-        return "%d %s" % (int(value), prefix)
+        return "%d %s " % (int(value), prefix)
     elif span > 3:
-      return "%.1f %s" % (float(value), prefix)
+      return "%.1f %s " % (float(value), prefix)
     elif span > 0.1:
       return "%.2f %s " % (float(value), prefix)
     else:
```

The labels now look like:
master_with_patch_adjusted_spacing

mhagger:
@cbowman0: thanks for the testing! (I don't have a graphite-web setup that I can use for testing, so I've been testing the input/output as text.)

Regarding graph (a): I take it that the inputs are all exactly zero? If so, I personally don't see a strong reason to prefer [-1..1] or [0..1]. If the latter is preferred, let me know and I will implement it (the place is `_AxisTics.reconcileLimits()` [here](https://github.com/mhagger/graphite-web/blob/8b5a7358adc2bd1738e837bc3b747f1c11b57388/webapp/graphite/render/glyph.py#L192-L196)).

Regarding graph (b): there is also an aesthetic difference in the choice of axis labels. I personally think that the new version looks better.

Regarding graph (c): here, to me, the new choice of y axis endpoints and tic marks is clearly better than the old. In fact it was examples like this that led me to start work on this PR.

Thanks for your proposed change to `makeLabel`. I'd be happy to make this change in my name or to pull the change from your fork, whichever you prefer.

However, taking a step back, I have to question why `makeLabel` has to add whitespace at the right side of label strings in the first place. Wouldn't it make more sense for the labels to be generated without padding, and for the rendering part of the code to position the labels more intelligently? And when there is no unit string, surely the extra space (which would otherwise separate the number from the label) should be omitted, too? I don't know that part of the code, though, so take this suggestion with a grain of salt.

If it were up to me, I would also omit the `.0` on label values when none of the labels' numbers have fractional parts. For example, instead of

    "0", "5.0 k", "10.0 k", "15.0 k"

, why not

    "0", "5 k", "10 k", "15 k"

? I'd be happy to implement that, too (though preferably as part of a separate PR).

Thanks again for the feedback, and let me know how you would like to proceed.

obfuscurity:
@mhagger @cbowman0 invited you both to the organization, really appreciate all the effort you've been putting into this. 🍰

cbowman0:
Thanks @obfuscurity .

Can you weigh in on @mhagger most recent questions about presentation?

obfuscurity:
I don't have any feedback on the code itself right now. The glyph stuff is so hairy it'll take me a bit of time to ramp up on it again. Or did you just mean afa aesthetic preferences?

cbowman0:
I was asking about the aesthetics preferences on the graph images.

obfuscurity:
Frankly they're all improvements as far as I'm concerned. The rendering stuff is hard to get perfect for all scenarios but I think this is generally better than where we started. Specifically wrt "should we present floats or integers" I think the current behavior is ok. I would typically argue in favor of consistency ("either all floats or none at all") but here we're using K units where I think it makes sense to be show some significant digits.

Does that address everything? 😛

webapp/graphite/render/glyph.py
```diff
@@ -226,19 +226,21 @@ def applySettings(self, axisMin=None, axisMax=None, axisLimit=None):
...
+        return "%.1f %s" % (value, prefix)
       else:
-        return "%d %s " % (int(value), prefix)
+        return "%d %s" % (int(value), prefix)
     elif span > 3:
-      return "%.1f %s " % (float(value), prefix)
+      return "%.1f %s" % (float(value), prefix)
     elif span > 0.1:
-      return "%.2f %s " % (float(value), prefix)
+      return "%.2f %s" % (float(value), prefix)
```
mhagger: @cbowman0: your suggested patch didn't include this last change. I think this was an oversight, so I removed the space here, too.

mhagger:
I think that changing where padding is implemented and whether to omit trailing zeros from label values should be addressed in separate PRs.

I just pushed a commit that should fix the spacing problem [pointed out by @cbowman0](https://github.com/graphite-project/graphite-web/pull/1401#issuecomment-218778147), using more or less his suggested change.

If there's anything else that people would like to see changed in this PR, please let me know. Otherwise, as far as I know, I've addressed all points that have been brought up.

obfuscurity:
Thanks for all your work on this @mhagger it's greatly appreciated. 🍰 ❤️ 🔥 🍦

PR #1402, opened 2015-11-22
- add missing documentation for hideXAxis to render api

leehuk:
hideXAxis was merged in #1239, unfortunately I missed the documentation for it so here it is


obfuscurity:
Thanks! :sparkles:

PR #1403, opened 2015-11-22
- correct yDivisor -> yDivisors and add an explanation of 'pretty'

leehuk:
The documentation uses 'yDivisor' where it should be using 'yDivisors'

This also confused me a little as to what it meant by 'pretty', so I added an explanation of what the codes trying to do in deciding the number of divisors, which I think is what its trying to do but not 100%!


obfuscurity:
Thanks!

PR #1405, opened 2015-11-24
Added fallback fonts where they were missing.

VibyJocke:
Mainly just added the fallback fonts, but also removed a duplicate key entry (and did some PEP8 compliant formatting while I was at it...).


obfuscurity:
Thanks!

PR #1277, opened 2015-07-13
Add APIs to control carbon-cache flushing

llicour:
This patch adds APIs that control carbon-cache mermoty flushing

* /metrics/persist-cache
This allow to ask carbon cache instances to dump their cache memory to disk. This action is executed during carbon-cache shutdown
and at regular interval. This can be force with this API
Sample : curl "http://localhost/metrics/persist-cache" -X POST

* metrics/flush-cache?metric=<regexp>
This allow to ask carbon cache instances to flush queues matching the given regexp
The current cache write strategy is hold during this operation, and restore after all matching queues are flushed
This operation can be cancel by calling the API with DELETE method (without parameters)
Sample : (start flushing) curl "http://localhost/metrics/flush-cache?metric=^carbon" -X POST
         (stop flushing)  curl "http://localhost/metrics/flush-cache" -X DELETE

Note: This patch requires a patch in carbon-cache (https://github.com/graphite-project/carbon/pull/419)

PR #1420, opened 2015-12-05
Query carbon cache when metrics not found in file system.

wooparadog:
For the circumstances when new metrics are not synced to file system, hence not found when querying.

Need help:

This is actually a hack, I don't have enough knowledge of how graphite-web is structured, so it needs a better way to get integrated into this project.

### DIFF
--- a/webapp/graphite/finders/standard.py
+++ b/webapp/graphite/finders/standard.py
@@ -4,7 +4,8 @@ from django.conf import settings
 
 from graphite.logger import log
 from graphite.node import BranchNode, LeafNode
-from graphite.readers import WhisperReader, GzippedWhisperReader, RRDReader
+from graphite.readers import WhisperReader, GzippedWhisperReader, RRDReader, \
+  CacheReader
 from graphite.util import find_escaped_pattern_fields
 
 from . import fs_to_metric, get_real_metric_path, match_entries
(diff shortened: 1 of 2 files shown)


deniszh:
An interesting feature, thanks, @wooparadog!
Not sure about realisation too... The separate reader looks like bit hackish, indeed.

Tests are broken because of docs again... we need to find out why...

bmhatfield:
I think this is a step towards a great idea, but I think this PR misunderstands the problem slightly.

Graphite-web *already* queries carbon-cache for datapoints in the case where it's aware a metric exists. The problem is that this awareness purely comes from globbing the file system. 

Carbon-Cache doesn't support any sort of "list keys" or globbing behavior, so this PR will only fix the case where the full path to the metric is specified.

A more correct fix, I think, is to add list-keys or globbing type behavior to carbon-cache as part of the "metric find" functionality. Then graphite-web will know there's some uncommitted metrics in carbon-cache and will naturally attempt to retrieve them. 

On the carbon side, that lookup will need to be performant, as it will happen frequently.

wooparadog:
Yeah, I realized that this implementation can't have lookup support. I'm working on carbon-cache to add a new `DrainStrategy` so that new metrics can get written faster.

wooparadog:
Closing this idae now...

deniszh:
@wooparadog - you can also check these PRs:
https://github.com/graphite-project/carbon/pull/419
https://github.com/graphite-project/graphite-web/pull/1277
Both need to be ported to master, though - but I tested that on 0.9.x - at least metric flushing works.

PR #1421, opened 2015-12-05
Release notes correction

obfuscurity:
... from @sylr


obfuscurity:
Merging this in even though the docs build is failing; something has broken outside of this diff.

PR #1423, opened 2015-12-05
Stop testing python 2.6 (0.9.x)

obfuscurity:
Twisted 15.5 officially breaks Python 2.6 support so let's stop testing it.

refs https://github.com/graphite-project/carbon/issues/482

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 python:
-  - "2.6"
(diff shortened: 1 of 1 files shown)


obfuscurity:
Note that I stopped short of requiring 2.7 in `check-dependencies.py` since 0.9.x is more likely to be running on older hosts.

PR #1424, opened 2015-12-05
Deprecate python 2.6

obfuscurity:
Twisted 15.5 officially breaks Python 2.6 support so let's stop testing it. Also updating `check-dependencies.py` to require 2.7.

refs https://github.com/graphite-project/carbon/issues/482


obfuscurity:
The docs build is broken, everything else passes. Merging.

PR #1431, opened 2015-12-08
Fix: wrong metrics data when it takes a lot of time writing to db.

wooparadog:
If one metrics stays in carbon-cache for too long(like over 1h
not committing to disk) and when we are querying one metrics within
that time, it's possible `i = int(interval - start) / step` is negative,
hence overwriting the correct value.


obfuscurity:
Seems fine. It'd be nice to know why we have cached data that old (i.e. we should purge it before we try to use it) but this looks like a reasonable fix for now.

PR #1434, opened 2015-12-14
Use settings.LOGOUT_URL in UI

oxyum:
Now graphite-web hardcodes logout url in template, which doesn't allow easy use of external authentication systems. This PR fixes it.


cbowman0:
This appears to have been corrected via #1487.  Closing this request because of that.

Issue #1285, opened 2015-07-20
[Feature Request] Metric Filtering by User.


toni-moreno:
Hi @cdavis  @SEJeff @brutasse .

I've been studying the best way to add some metric filtering per user in a big platform with Grafana + Graphite.

After having a initial discussion with  @torkelo grafana owner ( https://github.com/grafana/grafana/issues/1313 ) we have concluded that there is only a valid way to filter metrics and this work should be done in the graphite backend.

We (me and 2 job mates)  have been working together in our platform and we have  already built a graphite patch that adds metric filtering by user and already working  ok . 

The best of our PoC is the low impact on the code and also could be enabled/disabled with a ALLOW_METRICFILTER_BY_USER in local_settings.py

In our PoC we only added a new "MetricFilter" model with  userid and FilterRegex fields and  we have intercepted any node/metric search query  in storage.py file , and applied  the filter over the query result. 

Would you like to test it ? 
What do you think about this new Feature in graphite?

We think this feature could be a great improvement to this great tool , and we would like to contribute.

Thank you very much.

brutasse:
There is a pull request #997 but it hasn't been updated in months. The [comment I left](https://github.com/graphite-project/graphite-web/pull/997#issuecomment-60050868) stills sums up my opinion about the feature.

What does your patch look like?

toni-moreno:
Hi @brutasse I saw this pull request some days ago. But I think we have not exactly the same goals.
In our PoC, changes are much lighter and only affect on metric visibility by user. ( nothing to do with dashboards ).

We will  prepare  a new git branch with our changes over a graphite-web fork to show you how we apply filters

toni-moreno:
Hi @brutasse . 

You can see code here.

https://github.com/graphite-project/graphite-web/compare/master...cosm0s:metric_filter

As you can see  , we have only created a model in a new folder (metric_filter) but we can relocate this code if needed, and intercepted the returned nodes/metrics previosly filtered.( only applied if ALLOW_METRICFILTER_BY_USER is set to true).

### Use Case (3 users) ###
Supose we have a metric scheme like that

```
<group>.<datacenter>.<server_type>.<hostname>.<product>.<metric_name>
```

by example 2 groups with 2 hosts running two products ( linux/oracle).

* marketing.datacenter1.bbdd.mktdb01.linux.xxxxx
* marketing.datacenter1.bbdd.mktdb01.oracle.xxxxx
[...]

PR #1435, opened 2015-12-15
Added new metric filtering System to graphite

cosm0s:
Hi @brutasse . 

We have finally built the proposed filtering system in https://github.com/graphite-project/graphite-web/issues/1285.

As described before with this PR admin can include/exclude metrics by user  and group  based on the metric names.

### DIFF
--- a/webapp/graphite/metric_filter/models.py
+++ b/webapp/graphite/metric_filter/models.py
@@ -3,13 +3,12 @@ from django.contrib.auth.models import Group
 from django.contrib.auth.models import User
 from django.core.exceptions import ObjectDoesNotExist
 
-
 class MetricFilter(models.Model):
   id = models.AutoField(primary_key=True)
   filter = models.CharField(max_length=256)
   description = models.CharField(max_length=200)
   type_filter = models.IntegerField(choices=[(0, 'include'), (1, 'exclude')])
-  __str__ = lambda self: "%s: %s" % (getTypeFilter(self.type_filter) ,self.filter)
+  __str__ = lambda self: "[%s]: %s " % (getTypeFilter(self.type_filter) ,self.filter)
 
   class Meta:
     db_table = "metric_filter"
@@ -17,7 +16,7 @@ class MetricFilter(models.Model):
 class UserMetricFilter(models.Model):
   user = models.ForeignKey(User)
   metric_filter = models.ManyToManyField(MetricFilter)
-  __str__ = lambda self: "User: [%s]" % self.use
+  __str__ = lambda self: "%s" % self.use
 
   class Meta:
     db_table = "user_metric_filter"
@@ -34,7 +33,7 @@ class UserMetricFilter(models.Model):
 class GroupMetricFilter(models.Model):
   group = models.ForeignKey(Group)
   metric_filter = models.ManyToManyField(MetricFilter)
-  _str__ = lambda self: "%s: %s" % (getTypeFilter(self.metric_filter.all()[0].type_filter), self.group)
+  __str__ = lambda self: "%s" % self.group
 
   class Meta:
     db_table = "group_metric_filter"
--- a/webapp/graphite/metric_filter/models.py
+++ b/webapp/graphite/metric_filter/models.py
@@ -5,49 +5,47 @@ from django.core.exceptions import ObjectDoesNotExist
 
 
 class MetricFilter(models.Model):
-    id = models.AutoField(primary_key=True)
-    filter = models.CharField(max_length=256)
-    description = models.CharField(max_length=200)
-    type_filter = models.IntegerField(choices=[(0, 'include'), (1, 'exclude')])
-    __str__ = lambda self: "Filter: [%s]" % self.filter
+  id = models.AutoField(primary_key=True)
+  filter = models.CharField(max_length=256)
+  description = models.CharField(max_length=200)
+  type_filter = models.IntegerField(choices=[(0, 'include'), (1, 'exclude')])
+  __str__ = lambda self: "Filter: [%s]" % self.filter
 
-    class Meta:
-        db_table = "metric_filter"
-
+  class Meta:
+    db_table = "metric_filter"
 
 class UserMetricFilter(models.Model):
-    user = models.ForeignKey(User)
-    metric_filter = models.ManyToManyField(MetricFilter)
-    __str__ = lambda self: "User: [%s]" % self.user
+  user = models.ForeignKey(User)
+  metric_filter = models.ManyToManyField(MetricFilter)
+  __str__ = lambda self: "User: [%s]" % self.user
 
-    class Meta:
-        db_table = "user_metric_filter"
+  class Meta:
+    db_table = "user_metric_filter"
 
-    @staticmethod
-    def getFilters(id):
-        try:
-            userFilter = UserMetricFilter.objects.get(user = id)
-        except ObjectDoesNotExist:
-            userFilter = None
-        if userFilter is not None:
-            return userFilter.metric_filter
-
+  @staticmethod
+  def getFilters(id):
+    try:
+      userFilter = UserMetricFilter.objects.get(user = id)
+    except ObjectDoesNotExist:
+      userFilter = None
+    if userFilter is not None:
+      return userFilter.metric_filter
 
 class GroupMetricFilter(models.Model):
-    group = models.ForeignKey(Group)
-    metric_filter = models.ManyToManyField(MetricFilter)
-    __str__ = lambda self: "Group: [%s]" % self.group
+  group = models.ForeignKey(Group)
+  metric_filter = models.ManyToManyField(MetricFilter)
+  __str__ = lambda self: "Group: [%s]" % self.group
 
-    class Meta:
-        db_table = "group_metric_filter"
+  class Meta:
+    db_table = "group_metric_filter"
 
-    @staticmethod
-    def getFiltersByUser(user):
-        for userGroup in user.groups.all():
-            try:
-                groupsFilter = GroupMetricFilter.objects.filter(group = userGroup.id)
-            except ObjectDoesNotExist:
-                groupsFilter = None
-        if groupsFilter is not None:
-            for groupFilter in GroupMetricFilter.objects.filter(group = userGroup.id):
-                yield groupFilter.metric_filter
+  @staticmethod
+  def getFiltersByUser(user):
+    for userGroup in user.groups.all():
+      try:
+        groupsFilter = GroupMetricFilter.objects.filter(group = userGroup.id)
+      except ObjectDoesNotExist:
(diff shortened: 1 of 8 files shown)


toni-moreno:
Hi @brutasse .

We have been working for last weeks in this PR from (https://github.com/graphite-project/graphite-web/issues/1285)

Of course this new feature is optional and disabled by default (for backwards compatibility), to enable you should change ALLOW_METRICFILTERING_BY_USER= True in local_settings.

After this change has been done all metrics in STORE.find() will pass a filter list , this filter list is made from user filter first and group filters last.

If filtering enabled nobody will be able to see any metrics until added some filters.

You can add filter to user and groups with the django admin in the "Metric Filter Administration" section.

metric_filter

## 1.- Create  metric filters

You should first create all metric name filters 

The filter table has 3 fields.

1.- filter regex field
2.- Description
3.- Type ( include / Exclude) .

image


## 2.- Asign filters to user / groups.
Add filters to user  / group in the "User Metric filter  / Group Metric Filter " table , by choosing the previously configured filters

seleccion_024

At this point you will see only selected metrics in the tree

webapp/graphite/metrics_filter.py
```diff
@@ -0,0 +1,37 @@
+import re
+
+from django.conf import settings
+from graphite.metric_filter.models import UserMetricFilter
+from graphite.metric_filter.models import GroupMetricFilter
+
+
+def nodeFilter(user, nodePath):
+  if settings.ALLOW_METRICFILTER_BY_USER:
```
SEJeff: Better would be to left indent all of the code below and instead make this:
```python
if not setitngs.ALLOW_METRICFILTER_BY_USER:
    return True
if user.id is not None:
...
```

webapp/graphite/metrics_filter.py
```diff
@@ -0,0 +1,37 @@
...
+    if user.id is not None:
+      filters = UserMetricFilter.getFilters(user.id)
+      if filters is not None:
+        for filter in filters.all():
```
SEJeff: Minor nit, but please don't ever use variables that override builtin python functions (such as filter). It often leads to much pain and gnashing of teeth when debugging down the road.

webapp/graphite/metrics_filter.py
```diff
@@ -0,0 +1,37 @@
...
+          matching = match(nodePath, filter)
+          if matching is not None:
+            return matching
+    for filters in GroupMetricFilter.getFiltersByUser(user):
+      if filters is not None:
+        for filter in filters.all():
```
SEJeff: Same as above, please don't use variable names that override stdlib builtin functions. I'm almost certain a linter such as flake8 would pick this up as a no-no.

webapp/graphite/metrics_filter.py
```diff
@@ -0,0 +1,37 @@
...
+          if filter is not None:
+            matching = match(nodePath, filter)
+            if matching is not None:
+              return matching
+    return False
+  else:
+    return True
```
SEJeff: Since there is no way to fall through from a return ie the `return matching` or `return False`, you can change:

```python
    return False
  else:
    return True
```

To simply:

```python
  return False
return True
```

webapp/graphite/metrics_filter.py
```diff
@@ -0,0 +1,37 @@
...
+
+def match(path, filter):
+  #log.info('testing %s' % filter.filter)
+  pattern = re.compile(filter.filter)
+  if pattern.match(path):
+    if filter.type_filter == 0:
+      return True
+    elif filter.type_filter == 1:
```
SEJeff: Would you mind setting these as constants instead of [magic numbers](http://stackoverflow.com/a/47902)? I had to go up to the model reference to know this was include vs exclude. For constants like `INCLUDE_FILTER` or `EXCLUDE_FILTER` I see no reason it would hurt for them to be globals, but up to you.

webapp/graphite/metrics_filter.py
```diff
@@ -0,0 +1,37 @@
...
+      return False
+  else:
+    return None
```
SEJeff: Same as above, the else can go away and please left indent the `return None`. Redundant code is extra code to maintain, which makes things harder to maintain.

SEJeff:
I'm a pretty big fan of this feature as it is something I could have used long ago. However, I can't stress enough that a large new feature, which could result in leaking information if regressed, needs tests so we don't break in the future. That is my only serious nit, is no tests. The rest are mostly just stylistic to improve the code.

Thanks for doing this @cosm0s

cosm0s:
Thanks for the answer, the advices and the useful guide you provide us to keep on with this new feature.
This PR is a litte sample to show what we want to do. With this good feedback, I'm going to spend my time and my efforts to complete the stylistic and implement all the necessary test to complete the asked feature.

brutasse:
+1 to everything @SEJeff said:

- thank you for working on this!
- tests are important :)

I'll add a couple of comments inline as well.

webapp/graphite/metric_filter/models.py
```diff
@@ -0,0 +1,56 @@
+from django.db import models
+from django.contrib.auth.models import Group
+from django.contrib.auth.models import User
+from django.core.exceptions import ObjectDoesNotExist
+
+class MetricFilter(models.Model):
+  id = models.AutoField(primary_key=True)
```
brutasse: This is provided by default by Django, so there's no need to specify an `id`

webapp/graphite/metric_filter/models.py
```diff
@@ -0,0 +1,56 @@
...
+  filter = models.CharField(max_length=256)
```
brutasse: Can you use another name that's not a Python builtin? Perhaps `pattern`?

webapp/graphite/metric_filter/models.py
```diff
@@ -0,0 +1,56 @@
...
+  description = models.CharField(max_length=200)
+  type_filter = models.IntegerField(choices=[(0, 'include'), (1, 'exclude')])
+  __str__ = lambda self: "[%s]: %s " % (getTypeFilter(self.type_filter) ,self.filter)
```
brutasse: The best practice in Django land in Python 2 is to define a `__unicode__` method (not `__str__`) and make it return a unicode string. Also it's fine to use `def`.

webapp/graphite/metric_filter/models.py
```diff
@@ -0,0 +1,56 @@
...
+    for userGroup in user.groups.all():
+      try:
+        groupsFilter = GroupMetricFilter.objects.filter(group = userGroup.id)
+      except ObjectDoesNotExist:
+        groupsFilter = None
+      if groupsFilter is not None:
+        for groupFilter in GroupMetricFilter.objects.filter(group = userGroup.id):
+          yield groupFilter.metric_filter
+
+def getTypeFilter(filter):
```
brutasse: For model fields with `choices`, Django provides a [get_FOO_display](https://docs.djangoproject.com/en/1.9/ref/models/instances/#django.db.models.Model.get_FOO_display) method on the instance. So you can use `self.get_type_filter_display()` in your model methods.

webapp/graphite/metric_filter/models.py
```diff
@@ -0,0 +1,56 @@
...
+  filter = models.CharField(max_length=256)
+  description = models.CharField(max_length=200)
+  type_filter = models.IntegerField(choices=[(0, 'include'), (1, 'exclude')])
+  __str__ = lambda self: "[%s]: %s " % (getTypeFilter(self.type_filter) ,self.filter)
+
+  class Meta:
+    db_table = "metric_filter"
+
+class UserMetricFilter(models.Model):
+  user = models.ForeignKey(User)
```
brutasse: You can use [related_name](https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.ForeignKey.related_name) and do `user.metric_filters.all()` instead of having to define a method for getting filters for a given user.

webapp/graphite/metric_filter/models.py
```diff
@@ -0,0 +1,56 @@
...
+  metric_filter = models.ManyToManyField(MetricFilter)
+  __str__ = lambda self: "%s" % self.use
+
+  class Meta:
+    db_table = "user_metric_filter"
+
+  @staticmethod
+  def getFilters(id):
+    try:
+      userFilter = UserMetricFilter.objects.get(user = id)
```
brutasse: Are users supposed to have at most one `UserMetricFilter`? You have a `ForeignKey` to `User` but if a user have multiple filters, calling `.get()` as done here will raise an exception.

Given that a UserMetricFilter has an m2m relation to MetricFilter, it seems that it'd be enough to make the UserMetricFilter - User relation a `OneToOneField` instead of a `ForeignKey`. Same logic for `GroupMetricFilter`.

webapp/graphite/storage.py
```diff
@@ -33,8 +34,9 @@ def __init__(self, finders=None, hosts=None):
     self.remote_stores = [ RemoteStore(host) for host in remote_hosts ]
 
 
-  def find(self, pattern, startTime=None, endTime=None, local=False):
+  def find(self, pattern, startTime=None, endTime=None, local=False, user=None):
```
brutasse: When `user` isn't provided, `user=None` breaks the `nodeFilter` function. I think you probably want to make `user` a required argument and update all `STORE.find()` occurrences to provide `user`.

webapp/graphite/storage.py
```diff
@@ -74,8 +76,9 @@ def find(self, pattern, startTime=None, endTime=None, local=False):
         if node.is_leaf:
           leaf_nodes.append(node)
         elif node.path not in found_branch_nodes: #TODO need to filter branch nodes based on requested interval... how?!?!?
-          yield node
-          found_branch_nodes.add(node.path)
+          if nodeFilter(user, node.path):
```
brutasse: `nodeFilter` probably needs to be applied one level up:

```python
for node in nodes:
  if not nodeFilter(user, node.path):
      continue
  if node.is_leaf:
    …
```

This avoids changing the logic below: only allows nodes end up in `leaf_nodes` and there is no need for the `minimal_node_set_filtered` calculation L137-146.

webapp/graphite/metric_filter/models.py
SEJeff: Agreed, I wasn't a huge fan of a lambda for this as it does sacrifice maintainability and readability.

obfuscurity:
@cosm0s Seems like this could be a very welcome addition but none of the comments have been addressed yet. Are you still able to contribute to this PR?

cosm0s:
I intend to continue with this pr, but right now I don't have enough time.

anitakrueger:
Is there any update on this by any chance? Or another workaround on how to restrict metric access in graphite/grafana? Thanks!!

PR #1439, opened 2015-12-22
Handle multiple series in aggregateLine

iain-buclaw-sociomantic:
The ``aggregateLine`` function should be able to handle multiple series lists being passed to it.

Also adjusts the result name so that when calling with ``aggregateLine(foo-*.bar)`` it will return each match, instead of ``aggregateLine(foo-*.bar, 42), aggregateLine(foo-*.bar, 45), ...``

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2715,8 +2715,8 @@ def constantLine(requestContext, value):
 
 def aggregateLine(requestContext, seriesList, func='avg'):
   """
-  Draws a horizontal line based the function applied to the series.
-
+  Takes a metric or wildcard seriesList and draws a horizontal line
+  based on the function applied to each series.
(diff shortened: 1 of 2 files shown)


obfuscurity:
@iain-buclaw-sociomantic The tests are still failing, can you fix this up and we'll get it merged in?

iain-buclaw-sociomantic:
Sure, I'll look into it.

iain-buclaw-sociomantic:
Hmm, would the desired behaviour be to ignore aggregated `None` values here?  Maybe not...

obfuscurity:
Would you mind also updating the comment to reflect that it supports multiple series?

iain-buclaw-sociomantic:
OK, done.

obfuscurity:
Thanks!

PR #1444, opened 2016-01-05
Add GOV.UK to list of users

mattbostock:
GOV.UK uses graphite extensively. I work on the GOV.UK Infrastructure Team.


obfuscurity:
Thanks!

PR #1446, opened 2016-01-06
Check generated index values when merging results with cache

jjneely:
The generated index should never be negative -- which does not generate
an exception and ends up corrupting the merged result.

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -486,6 +486,7 @@ def mergeResults(dbResults, cacheResults, lowest_step):
 
   (timeInfo,values) = dbResults
   (start,end,step) = timeInfo
+  length = len(values)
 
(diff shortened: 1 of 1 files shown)


jjneely:
Any comments / approvals here?

obfuscurity:
@jjneely Sorry for the slow response. This looks similar to https://github.com/graphite-project/graphite-web/pull/1431/commits/492ef8f8b7d49c13f5168a36cb9770f6efb05150 which merged in recently to master. Which approach do you prefer?

jjneely:
These look equivalent.  I tend to use try/except sparingly is the only real difference.  Let's get 492ef8f into this branch as well as its already in master.  That's what I lean toward.

obfuscurity:
Sounds good, I'll whip up a PR.

obfuscurity:
Oops, already fixed in https://github.com/graphite-project/graphite-web/pull/1672.

Issue #1454, opened 2016-01-20
remote_storage.FindRequest harmfully swallows exceptions


drawks:
It looks like remote_storage.FindRequest swallows exceptions by design. While troubleshooting some remote rendering stuff that I thought should work I found and removed the exception eating behavior:

https://github.com/graphite-project/graphite-web/blob/0.9.15/webapp/graphite/remote_storage.py#L36

Which then opened the door to a few underlying bugs, the 0.9.15 version doesn't have "graphite.intervals" which makes the unpickler barf:

https://github.com/graphite-project/graphite-web/blob/0.9.15/webapp/graphite/util.py#L104
https://github.com/graphite-project/graphite-web/blob/0.9.15/webapp/graphite/util.py#L130

In my case I was trying to get 0.9.15 to talk to some older instances for federation so missing exceptions also masked that the returned storage objects had some of their key names changes between versions. "is_leaf" became "isLeaf" and "path" became "metric_path".

I suppose I shouldn't expect cross version compatibility here, although it is kind of bogus to break apis in minor revisions; but the main point is that we should not eat unhandled exceptions.

obfuscurity:
I assume you mean _should **not** eat_? :smile:

drawks:
#1455 Should address this issue...

mleinart:
I'm a bit confused.. Are you sure youre running 0.9.15 across the board? It sounds like you have a version running master branch trying to talk to 0.9.x branch. afaik the is_leaf=>isLeave et al changes have always only been between 0.9.x and master

drawks:
I am using cross version nodes, however the compatibility between versions was trivial to implement AND the pattern by which the exception was silently swallowed is bad practice. So, two birds with one stone. This PR corrects for the exception handling and opens the door for cross version metrics/find which allows for arbitrary federation between dissimilar versions.

PR #1455, opened 2016-01-20
Fix issue #1454

drawks:
* Remove exception supression from remote_storage.FindRequest
* Include intervals.py from Main
* Update remote_storage.FindRequest.get_results to be compatible with both
  old and new metrics/find api pickle packet formats

### DIFF
--- a/webapp/graphite/intervals.py
+++ b/webapp/graphite/intervals.py
@@ -1,0 +1,132 @@
+INFINITY = float('inf')
+NEGATIVE_INFINITY = -INFINITY
+
+
+class IntervalSet:
+  __slots__ = ('intervals', 'size')
+
+  def __init__(self, intervals, disjoint=False):
+    self.intervals = intervals
+
+    if not disjoint:
+      self.intervals = union_overlapping(self.intervals)
+
+    self.size = sum(i.size for i in self.intervals)
+
+  def __repr__(self):
+    return repr(self.intervals)
+
+  def __iter__(self):
+    return iter(self.intervals)
+
+  def __nonzero__(self):
+    return self.size != 0
+
+  def __sub__(self, other):
+    return self.intersect( other.complement() )
+
+  def complement(self):
+    complementary = []
+    cursor = NEGATIVE_INFINITY
+
+    for interval in self.intervals:
+      if cursor < interval.start:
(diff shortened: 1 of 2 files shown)


drawks:
FWIW this PR returns the exact same test failures as 0.9.x, those failures should be corrected, but I'm not sure I have a stomach for the archeology to find where they were introduced.

deniszh:
s/1554/1454/g ?

drawks:
@deniszh Yah, thanks. I fixed it

mleinart:
I'm :+1: for not swallowing the exceptions but I really dont think we should be bringing in the intervals stuff into 0.9.x or shimming in compatibility to cluster with the master branch

drawks:
The safe unpickler in 0.9.15 ALREADY referenced graphite.intervals, so it appears that it was the intention of that code for it to already be available. I've not done the archeology to find when/why intervals appeared in master or what the alternative to that type in a pickle packet should be or how that mapping should be implemented. 

I will say that if such a trivial change brings cross version api compat in a way that make federation work that it'd be a big win for facilitating changing versions in production without trying to cut everything over at once. It seems, to me, a good value from a marginally distasteful change.

drawks:
So whats the consensus here? I'm fine running this patch out of tree to scratch my own itch, as it were. I still think this looks like the right way to handle this in tree. If bringing intervals from master is really that distasteful it seems that the unpickler code in 0.9.x should also have references to it scrubbed.

mleinart:
Yeah, the unpickler code should be cleaned up - I'm sure that's just
copypasta

I'd still like to see that exception swallowing go away, but I remain
unconvinced that bringing in compatibility across major versions is a good
idea. Supporting cross-version clusters complicates debugging user problems
and could cause people silent breakage rather than a fail-fast break as
things in master change. It's also bad precedent to set - we need to be
able to break compatibility to move forward. For instance, supporting
archived timesteps in Ceres will require significant changes to
intervals.py (support for comparing intervals of different timesteps). We
already have a sort of hell keeping the changes that come in to 0.9.x added
to master, I'd rather us not start committing to going the other direction
too

obfuscurity:
+1 to @mleinart's comment

drawks:
K, fair. I'll rejigger this PR when I have a few moments free.

PR #1466, opened 2016-02-05
Fix issue #1454

drawks:
* Don't suppress exceptions in remote_storage.FindRequest


drawks:
git is hard.... this pr replaces #1455

PR #1469, opened 2016-02-17
'%l' can fail silently, updated percent_l_supported test

rvernica:
Here is a test showing that `%l` in `strftime` can fail silently:
```
# python
Python 2.7.10 (default, Aug 13 2015, 12:27:27) 
[GCC 4.9.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from datetime import datetime
>>> datetime.now().strftime("%a %l%p")
''
>>> 
```
The existing test does not catch this case. This patch fixes that.


obfuscurity:
Good catch, thanks!

PR #1470, opened 2016-02-18
Fix persistent XSS. Sanitize the path of the Graph Target

Sinkmanu:
Hello,

I found a persistent XSS in the target parameter, when we save an script code here e.g. 
```
 <svg/onload=alert()> 
```
this script is executed when we access to Graph Data.  
An malicious user could save javascript code in his graph and the javascript will be executed to any user that accessed to this graph. Is not possible steal the cookie because the httponly flag is present, but the attacker could do other malicious actions.

I fixed it with this code.

### DIFF
--- a/webapp/graphite/browser/views.py
+++ b/webapp/graphite/browser/views.py
@@ -22,6 +22,8 @@ from graphite.compat import HttpResponse
 from graphite.util import getProfile, getProfileByUsername, json
 from graphite.logger import log
 from hashlib import md5
+from urlparse import urlparse, parse_qsl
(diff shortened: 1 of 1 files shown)


gwaldo:
@graphite-project/committers Can someone with more domain expertise have a look at this?

@Sinkmanu Do you have any tests that you could provide that we could use to to confirm the fix?

obfuscurity:
This change doesn't support multiple `target` params. Fixing.

obfuscurity:
Closing this in favor of #1662.

PR #1475, opened 2016-02-23
Allow for rendering PDF files

squarebracket:
This is a very simple commit to allow for rendering PDF files through the render API. The documentation has been updated to reflect it.

I do a little hacky conversion from pixels to points so that the API is always using pixels instead of points.

Let me know if I forgot anything.

### DIFF
--- a/docs/render_api.rst
+++ b/docs/render_api.rst
@@ -207,7 +207,7 @@ Data Display Formats
 ====================
 
 Along with rendering an image, the api can also generate
-`SVG <http://www.w3.org/Graphics/SVG/>`_  with embedded metadata or return the raw data in various
+`SVG <http://www.w3.org/Graphics/SVG/>`_  with embedded metadata, PDF, or return the raw data in various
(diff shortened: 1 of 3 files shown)


deniszh:
LGTM!
but noting that fact that we will have no 0.9.x releases anymore - could you please port that change to master branch too?

squarebracket:
Oh, that sucks. Even for bug fixes? SVG rendering is broken in my environment and I was going to submit another PR for that. 

Just took a look at the master branch code. Looks like it will take me longer to set up a test rig for the master branch than porting the code. I suppose I have to open a new PR since it's merging to a different branch, right?

deniszh:
Yep, separate PR is needed for master. 
Not sure about releasing bugfixes for 0.9.x, but doubt so. Am I right, @obfuscurity ?

SVG rendering is another PITA. Maybe that pesky tag is dependent on Cairo version, or something?

obfuscurity:
Thanks!

PR #1476, opened 2016-02-24
dynamically collapse navbar on dashboard load

cbowman0:
North navbar non-collapsed by default.

Collapse the navbar when loading a dashboard from a hash change. Unify the hash change code so loading the dashboard happens and the navbar is collapsed.

Don't animate the collapse of the navbar as it makes the loading of the dashboard more fluid.


cbowman0:
Anyone have thoughts on this?  We've been running this for a few years without any complaints from my users.

obfuscurity:
Sorry, I didn't see this one earlier (or forgot about it). Seems like a good idea. 👍 

@cbowman0 feel free to merge this one if you'd like, you're on the team now. 😸

PR #1478, opened 2016-02-29
add new function named applyByNode

Roguelazer:
Add a new function, named `applyByNode`, which allows applying arbitrarily complicated expressions around prefixes (up to a given node). The example I made (reproduced from the docstring):

```
applyByNode(servers.*.disk.bytes_*, 1, "divideSeries(%.disk.bytes_used, sumSeries(%.disk.bytes_*))")
```

This would first evaluate all common prefixes up-to-and-including node 1 of the given seriesList (`servers.a`, `servers.b`, etc), and then substitute that into the template string (I used `%` as a template placeholder, but would be fine changing it) and evaluate the given result string.

Some of this is possible with `mapSeries` and `reduceSeries`, but, for example, this particular example is either impossible or super-complicated unless you go out of your way to store a separate `total_bytes` column.

This works (which is to say, we use it internally in production) and has tests, so, uh, that's a plus, right?

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -187,7 +187,6 @@ def sumSeries(requestContext, *seriesLists):
   of the other metrics is averaged for the metrics with finer retention rates.
 
   """
-
   try:
     (seriesList,start,end,step) = normalize(seriesLists)
   except:
@@ -580,7 +579,7 @@ def divideSeries(requestContext, dividendSeriesList, divisorSeries):
 
   """
   if len(divisorSeries) != 1:
-    raise ValueError("divideSeries second argument must reference exactly 1 series")
+    raise ValueError("divideSeries second argument must reference exactly 1 series (got {0})".format(len(divisorSeries)))
 
   divisorSeries = divisorSeries[0]
   results = []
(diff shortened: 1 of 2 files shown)


webapp/graphite/render/functions.py
```diff
@@ -2958,6 +2957,33 @@ def reduceSeries(requestContext, seriesLists, reduceFunction, reduceNode, *reduc
...
+def applyByNode(requestContext, seriesList, nodeNum, templateFunction, newName=None):
+  """
+  Takes a seriesList and applies some complicated function (described by a string), replacing templates with uniquified
+  values from the seriesList up to some unique node.
+
+  ...code-block:: none
+
+      &target=applyByNode(servers.*.disk.bytes_used,1,"divideSeries(%.disk.bytes_used, sumSeries(%.disk.bytes_*)))
+
+  Would find all series which match `servers.*.disk.bytes_free`, then trim them down to unique series up to the node
```
dnathe4th: nitpick: `bytes_used` is what the example uses, not `bytes_free`.

webapp/graphite/render/functions.py
```diff
@@ -2958,6 +2957,33 @@ def reduceSeries(requestContext, seriesLists, reduceFunction, reduceNode, *reduc
...
   return [ metaSeries[key] for key in keys ]
 
+def applyByNode(requestContext, seriesList, nodeNum, templateFunction, newName=None):
+  """
+  Takes a seriesList and applies some complicated function (described by a string), replacing templates with uniquified
+  values from the seriesList up to some unique node.
+
+  ...code-block:: none
+
+      &target=applyByNode(servers.*.disk.bytes_used,1,"divideSeries(%.disk.bytes_used, sumSeries(%.disk.bytes_*)))
```
JeanFred: Looks like you are opening a double-quote at `divideSeries` that does not get closed?

JeanFred:
Oh my, I spent quite some time wrestling with map/reduce trying to such things − I’d use this for sure :)

(Thanks for writing a test too!)

deniszh:
:+1:

JeanFred:
Looks good to me @Roguelazer. Would you mind squashing these?

JeanFred:
Perfect @Roguelazer, thanks :)

Unless there is opposition (thoughts @brutasse, @obfuscurity?) I’d be happy to go ahead and merge this by the end of the week (to let some time for others to chime in).

webapp/graphite/render/functions.py
```diff
@@ -2958,6 +2957,33 @@ def reduceSeries(requestContext, seriesLists, reduceFunction, reduceNode, *reduc
     metaSeries[key].name = key
   return [ metaSeries[key] for key in keys ]
 
+def applyByNode(requestContext, seriesList, nodeNum, templateFunction, newName=None):
+  """
+  Takes a seriesList and applies some complicated function (described by a string), replacing templates with uniquified
```
obfuscurity: I don't think that "uniquified" is a real word, perhaps "normalized" has the same connotation you're looking for?

webapp/graphite/render/functions.py
```diff
@@ -2958,6 +2957,33 @@ def reduceSeries(requestContext, seriesLists, reduceFunction, reduceNode, *reduc
...
+  values from the seriesList up to some unique node.
+
+  ...code-block:: none
+
+      &target=applyByNode(servers.*.disk.bytes_free,1,"divideSeries(%.disk.bytes_free,sumSeries(%.disk.bytes_*))")
+
+  Would find all series which match `servers.*.disk.bytes_free`, then trim them down to unique series up to the node
+  given by nodeNum, then fill them into the template function provided (replacing % by the prefixes)
+  """
```
obfuscurity: First, a simple nit - this sentence is missing a period at the end.

I'm pretty sure I grok the intent of this function, but perhaps you could elaborate by giving a series of intermediate examples demonstrating how each series would be interpolated into the `templateFunction`?

obfuscurity:
Other than the grammar/documentation edits I've suggested, it looks generally ok to me and I can certainly see the utility.

JeanFred:
And merged!

PR #1479, opened 2016-03-01
Graphs -> New Graph -> From Metric has the wrong function called when pressing enter on the input box

cbowman0:
Correct function call to be the correct one.


deniszh:
:+1:

cbowman0:
OK to push this?

obfuscurity:
Thanks!

PR #1482, opened 2016-03-07
Convert event time to local tz in events function

teamurko:
Events times are returned from db in UTC timezone. Convert them in local time to be aligned with query times.


JeanFred:
Thanks @teamurko! Would you be able to provide a test for this new `utc_to_local` method?

teamurko:
Thanks for feedback. Actually, it looks like it might be sufficient to use timezone aware util.epoch function because django db returns datetime objects with timezone specified. Tested on graphite-web=0.9.15.

obfuscurity:
Thanks to @cbowman0 for reminding me this one still needs to be merged.

PR #1485, opened 2016-03-12
Update storage-backends.rst

wallacejason:


gwaldo:
I agree that the wording here is goofy, but I'd rather use other wording.  Thank you for bringing it up, though!

PR #1492, opened 2016-03-28
Improve render API performance when the `target` includes many metrics.

mechairoi:
When the `target` includes many metric paths (ex. `avg(test.host0.loadavg5, test.host1.loadavg5, ..., test.host1000.loadavg5)`), render URI API calls `os.listdir` many times and its response is slower.

This PR reduces `os.listdir` calls if the metric paths contain no wildcards.

### DIFF
--- a/webapp/graphite/finders/standard.py
+++ b/webapp/graphite/finders/standard.py
@@ -68,16 +68,24 @@ class StandardFinder:
     match the corresponding pattern in patterns"""
     pattern = patterns[0]
     patterns = patterns[1:]
-    try:
-      entries = os.listdir(current_dir)
-    except OSError as e:
-      log.exception(e)
-      entries = []
 
+    has_wildcard = pattern.find('{') > -1 or pattern.find('[') > -1 or pattern.find('*') > -1
+
+    if has_wildcard:
+      try:
+        listdir = os.listdir(current_dir)
+      except OSError as e:
+        log.exception(e)
+        listdir = []
+    else:
+      listdir = []
+
+    entries = listdir if has_wildcard else [ pattern ]
     subdirs = [entry for entry in entries if isdir(join(current_dir, entry))]
     matching_subdirs = match_entries(subdirs, pattern)
 
     if len(patterns) == 1 and RRDReader.supported: #the last pattern may apply to RRD data sources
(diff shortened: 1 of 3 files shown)


webapp/graphite/finders/standard.py
```diff
@@ -68,16 +68,24 @@ def _find_paths(self, current_dir, patterns):
     match the corresponding pattern in patterns"""
     pattern = patterns[0]
     patterns = patterns[1:]
-    try:
-      entries = os.listdir(current_dir)
-    except OSError as e:
-      log.exception(e)
-      entries = []
 
+    has_wildcard = pattern.find('{') > -1 or pattern.find('[') > -1 or pattern.find('*') > -1
```
bmhatfield: A glob technically also can have a `?`, not sure if it's possible to pass it through the render API though.

webapp/graphite/finders/standard.py
```diff
@@ -68,16 +68,24 @@ def _find_paths(self, current_dir, patterns):
...
+    if has_wildcard:
+      try:
+        listdir = os.listdir(current_dir)
+      except OSError as e:
+        log.exception(e)
+        listdir = []
+    else:
+      listdir = []
+
+    entries = listdir if has_wildcard else [ pattern ]
```
bmhatfield: Instead of `if has_wildcard` here, why not just set `listdir = [ pattern ]` on 81?

bmhatfield: Or rather, set `entries` in 74-81 and skip the `listdir` variable altogether.

bmhatfield:
This PR seems like a good idea - thanks for identifying a performance problem in Graphite and submitting a PR to resolve it.

I made a couple of comments that may be imprecise; my request is to perhaps take a second pass over the changes as submitted and see if they can be simplified somewhat.

Also, please add some tests for these changes, since future folks working on this code may not understand why it was changed and unintentionally remove your awesome improvement!

mechairoi:
@bmhatfield Thanks for your reviews! I fixed them and added some tests. But latest Sphinx breaks the CI build.

bmhatfield:
This looks good to me! Thanks for tidying it up and adding tests!

@obfuscurity happy to have a double check here, otherwise I'll merge this later this afternoon.

PR #1500, opened 2016-04-10
Update aggregation-rules.conf documentation

sarg:
Update aggregation-rules.conf documentation with examples from aggregation-rules.conf.example

### DIFF
--- a/docs/config-carbon.rst
+++ b/docs/config-carbon.rst
@@ -180,8 +180,23 @@ As an example, if the following metrics are received:
   prod.applications.apache.www05.requests
 
 They would all go into the same aggregation buffer and after 60 seconds the
-aggregate metric 'prod.applications.apache.all.requests' would be calculated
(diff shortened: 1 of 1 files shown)


gwaldo:
Can any @graphite-project/committers comment on the accuracy of this doc?

obfuscurity:
Yes, it appears accurate. The relevant bits are [here](https://github.com/graphite-project/carbon/blob/9bc35a040a4283615240b58af87f264b7b00bfd6/lib/carbon/aggregator/rules.py#L104) and [here](https://github.com/graphite-project/carbon/blob/9bc35a040a4283615240b58af87f264b7b00bfd6/lib/carbon/aggregator/rules.py#L113). The latter (traditional template field) has an explicit match for the dot, the former (`<<expanding>>` field) lacks this delimiter.

At least, that's _my_ hot take on this code. :wink:

Issue #1495, opened 2016-04-05
countSeries() errors if passed an empty list


ckolbeck:
Expected: 0
Got:

```
<omitted top of stack>
File "/mnt/services/graphite/lib/graphite/render/functions.py", line 2328, in countSeries
    (seriesList,start,end,step) = normalize(seriesLists)
  File "/mnt/services/graphite/lib/graphite/render/functions.py", line 129, in normalize
    step = reduce(lcm,[s.step for s in seriesList])
TypeError: reduce() of empty sequence with no initial value
```

We're running a 0.10.0 alpha which is a bit behind tip, but I can't tell exactly which commit it's based on. Looking at master, it looks like the behavior in this edge case has changed to instead return an empty series, but I'd contend that that's also not the expected behavior in this case.   

Context: Attempting to construct a graph of something like "Percent of \<thing\> over \<threshold\>".

Thanks a bunch

PR #1507, opened 2016-04-18
Make countSeries([]) return constantLine(0) instead of an empty series

ckolbeck:
Issue here: https://github.com/graphite-project/graphite-web/issues/1495

Given that no one else commented on the above the current behavior may be desired, but it surprised me.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2839,11 +2839,16 @@ def countSeries(requestContext, *seriesLists):
     &target=countSeries(carbon.agents.*.*)
 
   """
-  (seriesList,start,end,step) = normalize(seriesLists)
-  name = "countSeries(%s)" % formatPathExpressions(seriesList)
-  values = ( int(len(row)) for row in izip(*seriesList) )
-  series = TimeSeries(name,start,end,step,values)
-  series.pathExpression = name
+  if seriesLists:
+    (seriesList,start,end,step) = normalize(seriesLists)
+    name = "countSeries(%s)" % formatPathExpressions(seriesList)
+    values = ( int(len(row)) for row in izip(*seriesList) )
+    series = TimeSeries(name,start,end,step,values)
+    series.pathExpression = name
+  else:
+    series = constantLine(requestContext, 0).pop()
+    series.pathExpression = "countSeries()"
+
   return [series]
 
 def group(requestContext, *seriesLists):


obfuscurity:
It sounds like this does effectively the same thing as `transformNull()` or `drawNullAsZero()`. In that case I'm :-1:.

ckolbeck:
@obfuscurity I don't believe it's the same, though I could be wrong. For instance consider something like `countSeries(currentAbove(<query>, <threshold>))` Since `currentAbove` here will return an empty list the issue isn't replacing nulls, but dealing with the absence of any series.

bitprophet:
Upon review, I think the root bug here is actually in `normalize()`, and given the stacktrace in #1495, it looks like it has received a fix sometime after the HEAD that I, being an asshole, apparently left Cory with: 9da1d8a896ae08deaa36c2c97ea5a5710ae24c0d - with that in play I'm guessing `countSeries` works better.

(Which is odd because I am 95% sure I updated the Graphite install he's using, well after 2013...but maybe something something master isn't release branch something? I'll hash it out with him.)

----

That said, if there's additional bugs in `countSeries` re: the empty-list-of-`seriesList`s scenario, I would argue it needs fixing and is not subject to `transformNull` et al, as those appear to be solely fixated on null values within single `seriesList` objects, which isn't the problem here.

ckolbeck:
@bitprophet Between the version we're running and HEAD the behavior when `normalize()`ing and empty series list went from throwing to returning an empty series list. that seems totally reasonable in almost every case, but `countSeries()` seems like an exception.

bitprophet:
@ckolbeck hm, so is something in `countSeries` still being poopy with an empty list (after the `normalize`)? That sounds like it falls under "additional bugs" to me. (In which case - I think this changeset here is still pretty cool. the count of an empty series list certainly is 0.

ckolbeck:
@bitprophet As it stands the empty series list will propagate out to the requester and you'll get no data. That makes perfect sense for most functions, so I wouldn't call it a bug in `normalize()` or even in `countSeries()`. Just that `countSeries()` is semi-unique in that there's a sensible alternative to return.

It's also worth noting that it's possible to get this same behavior be combining other functions, for instance: `fallbackSeries(countSeries(server*.requests_per_second), constantLine(0))` (a small modification to an example in the `fallbackSeries()` docs), but this doesn't play all that nicely with some dashboard systems such as Grafana.

ckolbeck:
@obfuscurity Sorry to prod, but has the above discussion changed your mind at all?

obfuscurity:
@ckolbeck Yeah, seems reasonable and I'm happy to merge. Would you consider adding a test for this?

1337newbee:
I'm glad to see this one approved. Which version of Graphite version that contains this change?

obfuscurity:
It's not merged yet. See my previous comment/question about adding a test.

ckolbeck:
@obfuscurity @1337newbee I'm just getting back from vacation and this is on my list. I'm pretty backlogged, but I should have something this week.

obfuscurity:
@ckolbeck No worries, hope you had a nice vacation. Whenever you have time.

1337newbee:
@ckolbeck let me know if there's any testing I can help with. We are currently blocking by this issue, so the seyren alert won't be triggered downstream. 

Also, I'm wondering if there's any equivalent function I can use for now. I tried fallbackSeries(countSeries(server*.requests_per_second), constantLine(0)), it threw out some exceptions as well. Not sure if that's because it requested the latest version of Graphite.

Thanks

obfuscurity:
Here's a possible test I've been working on for this but I can't seem to get it working. It keeps returning the full original expression instead of the altered one. Browser (`format=json`) works as expected so I'm probably overlooking something stupid.

@cbowman0 see anything obvious?

```
diff --git a/webapp/tests/test_functions.py b/webapp/tests/test_functions.py
index c50342c..06a46e7 100644
--- a/webapp/tests/test_functions.py
+++ b/webapp/tests/test_functions.py
@@ -1559,6 +1559,22 @@ class FunctionsTest(TestCase):
         result = functions.countSeries(request_context, seriesList)
         self.assertEqual(result, expectedResult)

+    def test_empty_countSeries(self):
+        seriesList = [
+            TimeSeries('removeAboveValue(constantLine(1),0)',0,600,60,[None,None,None,
+        ]
+        for series in seriesList:
+            series.pathExpression = series.name
+        expectedResult = [
+            TimeSeries('countSeries(removeAboveValue(1, 0))',0,600,60,[0,0,0,0,0,0,0,0
+        ]
+        for series in expectedResult:
+            series.options = {}
+
+        request_context = {}
+        result = functions.countSeries(request_context, seriesList)
+        self.assertEqual(result, expectedResult)
+
     def test_group(self):
         seriesList = [
             TimeSeries('collectd.test-db1.load.value',0,600,60,[1,2,3,4,5,6,7,8,9,10])
```

```
FAIL: test_empty_countSeries (tests.test_functions.FunctionsTest)
  File "/usr/local/src/graphite-web/webapp/tests/test_functions.py", line 1576, in test_empty_countSeries
    self.assertEqual(result, expectedResult)
AssertionError: Lists differ: [TimeSeries(name=countSeries(r... != [TimeSeries(name=countSeries(r...

First differing element 0:
TimeSeries(name=countSeries(removeAboveValue(constantLine(1),0)), start=0, end=600, step=60)
TimeSeries(name=countSeries(removeAboveValue(1, 0)), start=0, end=600, step=60)

- [TimeSeries(name=countSeries(removeAboveValue(constantLine(1),0)), start=0, end=600, step=60)]
?                                               ------------- -

+ [TimeSeries(name=countSeries(removeAboveValue(1, 0)), start=0, end=600, step=60)]
?                                                 +


Ran 609 tests in 3.392s
```

cbowman0:
Based on how the changes to countSeries() is written, this is what will work for a test:

```
--- a/webapp/tests/test_functions.py
+++ b/webapp/tests/test_functions.py
@@ -1559,6 +1559,18 @@ class FunctionsTest(TestCase):
         result = functions.countSeries(request_context, seriesList)
         self.assertEqual(result, expectedResult)
 
+    def test_empty_countSeries(self):
+        expectedResult = [
+            TimeSeries('0',0,600,300,[0,0,0]),
+        ]
+
+        request_context = {
+                  'startTime': datetime(1970, 1, 1, 0, 0, 0, 0, pytz.timezone(settings.TIME_ZONE)),
+                  'endTime': datetime(1970, 1, 1, 0, 10, 0, 0, pytz.timezone(settings.TIME_ZONE)),
+        }
+        result = functions.countSeries(request_context)
+        self.assertEqual(result, expectedResult)
+

```

cbowman0:
To expand, it expects there are NO passed in seriesList.  Also, without resetting name on the new series, it'll keep the name from constantLine, hence the '0' and constantLine() uses the startTime and endTime of the request context to build the step.

obfuscurity:
@cbowman0 Thanks, that makes sense. Replacing this PR with #1608 so we can add the test.

PR #1513, opened 2016-05-04
Small documentation update for 0_9_15.rst

Pallieter:
Added the command needed to check the Django version from the CLI.


obfuscurity:
I'm not sure I agree with this addition. There are other locked dependency versions (e.g. `django-tagging`) and we have the mechanisms in place (e.g. `requirements.txt`) to help enforce these. I appreciate the contribution but I'm going to politely decline this one for now. 

P.S. FWIW I think a better approach would be to use something like `pip freeze`.

PR #1515, opened 2016-05-09
Set timeout in RemoteReader.fetch to be REMOTE_FETCH_TIMEOUT instead of REMOTE_FIND_TIMEOUT

cbowman0:
Commit 3d378b2f528d0040256055b2af85839813006411 changed it to REMOTE_FIND_TIMEOUT which I believe to be an error.


obfuscurity:
Thanks @cbowman0 👍

PR #1518, opened 2016-05-18
The SI prefix for thousands is abbreviated "k", not "K"

mhagger:
See, for example, [the wikipedia article](http://en.wikipedia.org/wiki/SI_prefix) referenced in [the docstring for `format_units`](https://github.com/mhagger/graphite-web/blob/df6c510b7967fea6b86cea0d85fc63121d6e4185/webapp/graphite/render/glyph.py#L1705-L1712).


bmhatfield:
It's true: http://physics.nist.gov/cuu/Units/prefixes.html

Please note that you have failing tests that are looking for 'K' and now seeing 'k'. Please go through and update any/all tests that would be affected here :-)

gwaldo:
I'm down as long as the tests are updated.

mhagger:
Sorry about the test failure. I just fixed it.

bmhatfield:
Looks great, thanks.

Issue #1521, opened 2016-05-19
movingAverage movingMedian very slow in cluster


arielnh56:
These seem to re-pull the data using the _fetchWithBootstrap() function.

For a specific query that was inserted into one of our grafana dashboards, this was mind-bogglingly slow.

In our environment we have 5 storage nodes. The original data query uses a wildcard which in this case matches 1411 metrics, (which are munged then combined with averageSeries) so each query to the storage nodes returns about 282 metrics. This returns pretty quick.
Someone encapsulated one of these queries with a movingAverage(x,100). This generated an additional 1411 separate data queries to the storage nodes (logged in rendering.log) each time. As the result took more than the grafana refresh time to generate (60s - actually took >200s sometimes) and is displayed in multiple locations, something like 90% of the queries to the storage nodes ended up being this one thing, and the whole cluster responds like a slug.

We are working around it by not using movingAverage on large queries, but it wouldn't it  be better if _fetchWithBootstrap() could match the wildcard originally used? 

I'm going to stare at it a bit and see if I can figure a fix, but my python foo may not be up to it.

arielnh56:
or maybe pre-process the query to look for the need for a bootstrap and pull the correct data the first time?

arielnh56:
this seems to apply whenever movingAverage is applies to a series that already has a function applied

for instance 

movingAverage(services.real-time-bidder.pnc.*.Heap.heap.percent,11)

generates two retrievals

Thu May 19 15:40:51 2016 :: Retrieval of services.real-time-bidder.pnc.*.Heap.heap.percent took 0.182778
Thu May 19 15:40:52 2016 :: Retrieval of services.real-time-bidder.pnc.*.Heap.heap.percent took 0.130155


while 

movingAverage(nonNegativeDerivative(services.real-time-bidder.pnc.*.Heap.heap.percent),11)

gets the initial wildcard, then a hit for each node, and takes an age. As most of our interesting metrics are counters that reset to zero at daemon restart, not taking nonNegativeDerivative() on them first would generate silly numbers in a movingAverage

arielnh56:
I dug deep and I think I get it now - and I think I see the fix - and I have tested it narrowly.

It appears to be an inconsistent way that TimeSeries.pathExpression is handled.

the field is initially set in fetchData as follows:

    series.pathExpression = pathExpr #hack to pass expressions through to render functions
 
some functions (all of the combination functions) pass this uphill nicely.

[...]

PR #1523, opened 2016-05-24
replace _fetchWithBootstrap

arielnh56:
I discovered that movingAverage() on data that was already retrieved by
a non-summary function like perSecond() was going back to retrieve
earlier data, which it needs to do to get the first points in the graph
correct. However in retrieving this data it was not using the original
wildcard but polling each datapoint individually. On queries with over a
thousand datapoints in a clustered environment this was incredibly
resource intensive and slow.

This change passes the original tokens to each function as an element
'args' in the resourceContext. So if a function needs to retrieve the
original data with a time offset it can do so, and retrieve the data in
the same way as it was originally retrieved with wildcards intact. This
is efficient.

Four functions used _fetchWithBootstrap, so I updated all of them to use
the new mechanism. There are a bunch of other functions that also re-pull
data but do not use _fetchWithBootstrap. I believe they could also
benefit from this, but I have not addressed them.

I verified the output against the original function results and did find
some discrepancies, but I believe these are bugs in the old code, so the
results should now be better as well as faster. These discrepancies
were:

In movingMedian _fetchWithBootsstap sometimes inserted an incorrect Null
before the time window, changing the result for the first few points

holtWinters functions old version had a discontinuity around a 7 day
time selection. 167 hours significantly different from 169 hours. This
discontinuity is now gone, but the results for shorter time windows are
now significantly changed. I'm not a statistician but I suspect these
are intended for long time windows and the discontinuity was a bug
brought on by mixing two whisper ranges together.


bmhatfield:
I like this change. However, I think 0.9.x might have a "no new merges" EOL policy? I am not sure @obfuscurity could say for certain. As someone who is also on 0.9.x, I'd personally be happy to have them.

In the case where this is not an official EOL, would it be possible to update/add tests for these changes?

deniszh:
IIRC, it's more like "no more releases from 0.9.x". IMO we can still accept valuable patches to 0.9.x.
But tests are really needed, though.
And maybe this patch need to be ported to master too?

arielnh56:
I would port to 10, but we're still on 0.9.x on our clusters, so I don't have a basis for testing. This is tested and running on our production clusters, found while figuring out why things were so slow.

This is also my first contrib, so I am unclear on the process for adding/updating tests. I ran these in parallel locally with a 'New" suffix on the name, and eyeballed the output for comparison, and dug into some of the tabular data to be sure.

obfuscurity:
Merging this because why not, right? :wink: 

But seriously, we still don't intend to cut another 0.9.x release but I don't see any harm in getting it in the branch, especially since we're likely to merge it into master too via https://github.com/graphite-project/graphite-web/pull/1529.

obfuscurity:
fixes #1521

PR #1524, opened 2016-05-26
Update who-is-using.rst

wonmo01:


gwaldo:
Do you work at Pandora and use Graphite yourself?

wonmo01:
Yes, I work at Pandora, and we are using graphite.

gwaldo:
Thank you! Is there any chance you'd be willing to participate in a case study (or refer someone there who is to me)?

PR #1530, opened 2016-06-03
Properly format floats in a handful of functions

drcapulet:


obfuscurity:
Thank you for adding these tests. Dumb question but why not just use `%f` instead of `%g`?

obfuscurity:
Disregard, `%g` definitely looks more suitable here.

obfuscurity:
LGTM. @deniszh @SEJeff @cbowman0 Any concerns?

cbowman0:
I have no concerns with it.

JeanFred:
Looks good to me as well.

deniszh:
:+1:

PR #1533, opened 2016-06-06
Add test coverage to /metrics/

cbowman0:
I am attempting to understand how to create test cases in Graphite.  This adds coverage to graphite.metrics.

### DIFF
--- a/webapp/graphite/metrics/views.py
+++ b/webapp/graphite/metrics/views.py
@@ -28,8 +28,11 @@ except ImportError:
 
 
 def index_json(request):
-  jsonp = request.REQUEST.get('jsonp', False)
-  cluster = request.REQUEST.get('cluster', False)
+  queryParams = request.GET.copy()
+  queryParams.update(request.POST)
+
+  jsonp = queryParams.get('jsonp', False)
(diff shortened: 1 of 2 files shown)


deniszh:
Same here. :+1:
Need to do the same, maybe will learn that too from your code. Much appreciated!

cbowman0:
I have something going for this, but to really test the edge cases more will be necessary.

Also, some views in graphite.metrics don't seem to work with Django 1.9.  I'll be pushing a fix after I transfer trains.

JeanFred:
Same as others − very much appreciate all this work! Will try to have a look later :)

cbowman0:
*grumble* returned json isn't sorted.

cbowman0:
I think this is as far as I'm going to get for now.  I'm ready to merge after someone else has had a chance to look over it.

There's a few cases (get/set metadata) where I don't know how to invoke a valid request.  I have deduced a valid key from a short amount of digging.

Overall, I think I have covered the majority of the requests variations, but that's where the coverage may be lacking so if someone sees something that was missed let me know.

obfuscurity:
:heart:

PR #1534, opened 2016-06-08
Add coverage to util.py for IPv6 and write_index

cbowman0:
Just a couple small test cases.


gwaldo:
I love that you're putting some effort into testing!  I'm not really competent with python, but having useful tests makes me more confident to make changes.

SEJeff:
Tests are always appreciated, thanks!

PR #1540, opened 2016-06-12
Add more test coverage to graphite/dashboard

cbowman0:
Add more test coverage to graphite/dashboard/
    
    Fix bug in email routine
    Fix bug in loading a template directly from URL
    Fix two bugs in template save and delete.

### DIFF
--- a/webapp/tests/test_dashboard.py
+++ b/webapp/tests/test_dashboard.py
@@ -1,10 +1,26 @@
+import copy
 import json
+
+from django.conf import settings
 
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 
 class DashboardTest(TestCase):
     def test_dashboard(self):
+
+        # Test when settings.DASHBOARD_CONF has a missing file 
+        url = reverse('dashboard')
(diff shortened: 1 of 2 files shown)


bmhatfield:
You are amazing. I gave this a skim but I think you are GTG at a high level, especially since the tests you added found bugs...

webapp/tests/test_dashboard.py
```diff
@@ -1,9 +1,226 @@
...
+
+from django.conf import settings
+
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 
-
 class DashboardTest(TestCase):
     def test_dashboard(self):
+
```
JeanFred: Generally speaking, I think it is better practice to keep test methods short, with explicit names on what they do. There are 50 assertions in this one test method! This implies that laters assertions might be dependant on the state left by previous test assertions − if that is the case, it should be done as part of a `setUp` method.

The comments left actually make good basis for methods like `test_append_dashboard_name_to_url_returns_200()`

PR #1544, opened 2016-06-16
Clean up all test files created in tests/test_util.py

cbowman0:
Not cleaning up some test files resulted in other newly created tests failing.

PR #1546, opened 2016-06-17
Fix new errors from newer pyflakes

cbowman0:
* Define a few variables before first use within conditionals.
* Remove extraneous define of CLUSTER_SERVERS
* Correct usage of join in test/test_settings.py


cbowman0:
This fixes issues seen in #1544 with newer versions of flask8/pyflakes.

obfuscurity:
LGTM. 👍

PR #1548, opened 2016-06-18
Adding test coverage for render/glyph.py

cbowman0:
Bug Fixes:
* Add some defensive input checks in find_x_times and format_units
* Explicitly define self.span and self.binary in class _LinearAxisTics.  It was being defined in method calls.

Standalone functions:
* Tests for format_units()
* Tests for find_x_times()
* Tests for toSeconds()
* Tests for sort_stacked()
* Tests for safeArgs()
* Tests for safeMin()
* Tests for safeMax()
* Tests for safeSum()
* Tests for any()

Classes:
* Tests for the _LinearAxisTics Class
* Tests for the _LogAxisTics Class

### DIFF
--- a/webapp/tests/test_render_glyph.py
+++ b/webapp/tests/test_render_glyph.py
@@ -1,0 +1,1002 @@
+import copy
+
+#from django.conf import settings
+from django.test import TestCase
+
+from graphite.render import glyph
+from graphite.render.datalib import TimeSeries
+
+class glyphStandaloneFunctionTest(TestCase):
+    from datetime import datetime, timedelta
+    dt = datetime(2016, 6, 16, 15, 55, 38)
+    dt2 = datetime(2016, 6, 17, 15, 55, 38)
+
+    # Convenience function
+    def _generate_series_list(self, config=None):
+        seriesList = []
+        if not config:
+            config = [range(101), range(101), [1, None, None, None, None]]
+
+        for i, c in enumerate(config):
+            name = "collectd.test-db{0}.load.value".format(i + 1)
+            seriesList.append(TimeSeries(name, 0, 1, 1, c))
+        return seriesList
+
+    #
+    # Testing toSeconds
+    #
+    def test_toSeconds_invalid_inputs(self):
+      with self.assertRaises(AttributeError):
+        glyph.toSeconds(None)
+      with self.assertRaises(AttributeError):
+        glyph.toSeconds('')
+      with self.assertRaises(AttributeError):
+        glyph.toSeconds(1)
+
+    def test_toSeconds(self):
+      self.assertEqual(glyph.toSeconds(self.dt2 - self.dt), 86400)
+
+    #
+    # Testing safeArgs()
+    #
+    def test_safeArgs_None(self):
+      with self.assertRaises(TypeError):
+        glyph.safeArgs(None)
+
+    def test_safeArgs_number_not_list(self):
+      with self.assertRaises(TypeError):
+        glyph.safeArgs(1)
+
+    def test_safeArgs_all_strings(self):
+      with self.assertRaises(TypeError):
+        list(glyph.safeArgs(['a','b','c']))
+
+    def test_safeArgs_list_with_string(self):
+      with self.assertRaises(TypeError):
+        self.assertEqual(list(glyph.safeArgs(['a',5.1,6])), ['a',5.1,6])
+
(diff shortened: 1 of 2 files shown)


cbowman0:
@mhagger Do you have feedback on these unit tests of some of the changes you recently made to render/glyph.ph?

@graphite-project/committers Feedback?

bmhatfield:
Hi,

I took a look at these. While the tests look fine, I agree with what seems to be a general unease around understanding the implications of the changes to glyph.py. What kinds of things can you do (screenshots?) to identify before/after changes or the lack thereof?

webapp/tests/test_render_glyph.py
```diff
@@ -0,0 +1,1002 @@
...
+
+    def test_safeSum_empty_string(self):
+      self.assertEqual(glyph.safeSum(''), 0)
+
+    def test_safeSum_list_with_string(self):
+      with self.assertRaises(TypeError):
+        glyph.safeSum(['a',10,30])
+
+    def test_safeSum_list_all_numbers(self):
+      self.assertEqual(glyph.safeSum([1,5.1,6,]), 12.1)
```
mhagger: Tests involving floating-point numbers are often subject to floating-point rounding errors, because `assertEqual()` insists that the numbers are identical, whereas they might, say, differ in the 13th decimal place due to rounding. `assertAlmostEqual()` is more robust in situations like this.

webapp/tests/test_render_glyph.py
```diff
@@ -0,0 +1,1002 @@
...
+        seriesList = self._generate_series_list(config)
+        self.assertEqual(glyph.dataLimits(seriesList, True, True), (10, 390))
+
+    def test_dataLimits_drawNull_ymin_positive_missing_data(self):
+        config = [range(10, 101), range(10, 101), [1, None, None, None, None]]
+        seriesList = self._generate_series_list(config)
+        self.assertEqual(glyph.dataLimits(seriesList, True, False), (0.0, 100))
+
+    def test_dataLimits_drawNull_ymax_negative_missing_data(self):
+        config = [range(-10, -101), range(-10, -101), [-50, None, None, None, None]]
```
mhagger: `range(-10, -101)` is `[]` (or the iterator equivalent in Python 3+). If that's what you want, it's probably clearer to write `[]` here. Alternatively, maybe you meant to add a third parameter?: `range(-10, -101, -1)`.

gwaldo:
As far as changes to visual elements, we could incorporate a perceptual-diff tool (idea cribbed from https://www.youtube.com/watch?v=1wHr-O6gEfc)

webapp/tests/test_render_glyph.py
```diff
@@ -0,0 +1,1002 @@
...
+      with self.assertRaises(glyph.GraphError):
+          y.applySettings(axisMin=1000, axisMax=10, axisLimit=None)
+
+    #
+    # Testing _LinearAxisTics.reconcileLimits()
+    #
+
+    def test_LinearAxisTics_reconcileLimits_defaults(self):
+      y = glyph._LinearAxisTics(0.0, 100.0, unitSystem='binary')
+      self.assertEqual(y.reconcileLimits(), None)
```
mhagger: The `AxisTics` classes were meant to be used by constructing an object and then calling the instance's `applySettings()`, `chooseStep()`, and `chooseLimits()` methods in sequence. Why? Because that was the smallest change from the previous code.

`reconcileLimits()` is really meant to be called only by `applySettings()`, and only in the case that `not (self.minValue < self.maxValue)`. So I think it is inappropriate for the tests in this section to be calling `reconcileLimits()` directly.

OTOH I think it would make sense to move the `not (self.minValue < self.maxValue)` check into `reconcileLimits()` to make it a little bit more self-contained. We could even make the method private, though when I was coding this I think I checked and didn't find many private methods, so I thought they were disfavored in this codebase.

I'm working on some patches to clear up minor points like this.

mhagger:
> @mhagger Do you have feedback on these unit tests of some of the changes you recently made to render/glyph.ph?

I made some comments here, and I also just submitted [a PR against this PR](cbowman0/graphite-web#1). I'm afraid that's all that I will have time for in the near future.

cbowman0:
Thank you, @mhagger.

I'm going to close this pull request and will open a new one later with the suggested changes and some visual confirmations.

PR #1550, opened 2016-06-22
Update install.rst

thijslemmens:
Latest versions of Django do not create account_profile. Also the syncdb command is replaced by migrate.


obfuscurity:
We don't want to lock Graphite to an old version of Django. If there are problems with newer versions we should fix those issues. Thank you for the proposed change though.

PR #1552, opened 2016-06-23
Don't stomp on configs

johnseekins:
Similar to https://github.com/graphite-project/carbon/pull/564, we shouldn't be stomping config values if they already exist at install time.


obfuscurity:
:zap::boom::sparkles::ice_cream:

Issue #1541, opened 2016-06-14
django-tagging 0.4 does not work with graphite, be more specific


jblaine:
django-tagging 0.4+ does not work with graphite 0.9.15 and Django 1.4.x or 1.5.x

version 0.3.6 is required

- [ ] pin django-tagging==0.3.6 in requirements.txt for 0.9.x branch for a 0.9.16 release?
- [ ] document?

deniszh:
We have https://github.com/graphite-project/graphite-web/blob/0.9.x/requirements.txt
We could change `django-tagging==0.3.1` to `django-tagging>=0.3.1,<0.4` but not sure if needed.
We also have `django-tagging<0.4` in https://github.com/graphite-project/graphite-web/blob/0.9.x/.travis.yml

obfuscurity:
Note that we're no longer evolving the `0.9.x` branch except in rare circumstances. Django-tagging has been pinned to version 0.4.3 in the `master` branch and it works as intended. Closing this issue.

Issue #1514, opened 2016-05-04
Creating Initial Schema Fails...


rsvancara:
Hi,

Attempting to initialize the database schema for graphite-web.  I have tried so many things and all them ask for the file graphite.settings.  I am not a django developer and the documentation does not appear to be correct either.

http://graphite.readthedocs.io/en/latest/config-database-setup.html
PYTHONPATH=$GRAPHITE_ROOT/webapp django-admin.py syncdb --settings=graphite.settings

I have tried:

source /opt/graphite/bin/activate
PYTHONPATH=/opt/graphite/webapp /opt/graphite/bin/django-admin.py syncdb

I have tried different versions of Django.  

Here is my error:
```
PYTHONPATH=/opt/graphite/webapp /opt/graphite/bin/django-admin.py syncdb
Traceback (most recent call last):
  File "/opt/graphite/bin/django-admin.py", line 5, in <module>
    management.execute_from_command_line()
  File "/opt/graphite/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/opt/graphite/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 377, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/graphite/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 238, in fetch_command
    klass = load_command_class(app_name, subcommand)
  File "/opt/graphite/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 41, in load_command_class
    module = import_module('%s.management.commands.%s' % (app_name, name))
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/opt/graphite/local/lib/python2.7/site-packages/django/core/management/commands/syncdb.py", line 5, in <module>
    from django.contrib.auth import get_user_model
  File "/opt/graphite/local/lib/python2.7/site-packages/django/contrib/auth/__init__.py", line 7, in <module>
    from django.middleware.csrf import rotate_token
  File "/opt/graphite/local/lib/python2.7/site-packages/django/middleware/csrf.py", line 14, in <module>
    from django.utils.cache import patch_vary_headers
  File "/opt/graphite/local/lib/python2.7/site-packages/django/utils/cache.py", line 26, in <module>
    from django.core.cache import caches
[...]

Issue #1509, opened 2016-04-21
Install docs / pip installation is broken


jvanasco:
I tried installing from PIP into a virtualenv using the docs.

This method is very broken.

pypi installs `0.9.15`, however this doesn't contain any information on dependencies -- so they're not automatically attempted.

the actual dependencies aren't listed in the docs or the pypi download.

the latest versions of django or django-tagging both seem to be incompatible with graphite in different ways as well.

the closest thing I found was the `requirements.txt` for the 0.9.15 tag (and master) both all list non-dependencies, re-list things like the whisper client (which is pegged to a git version, not the pypi version that the installation instructions give), and list things that are not compatible.

Installing these via pip should work:

    pip install Django==1.4
    pip install django-tagging==0.3.1
    pip install pytz

I was not able to find a version of whitenoise that would work.  the 3.0 branch wants a django with `django.utils.six`, the 2.x branch raises errors on a url component split.

The docs should reflect an install procedure that could actuall work.

There should be an actual functional set of requirements for a given release as well.

obfuscurity:
@jvanasco If you're still interested, I'd appreciate your feedback with the latest versions of the install documents and updated setuptools for pip.

http://graphite.readthedocs.io/en/latest/install.html
http://graphite.readthedocs.io/en/latest/install-virtualenv.html
http://graphite.readthedocs.io/en/latest/install-pip.html

This process has been updated so that anyone installing from these ("latest") docs should get the proper versions of each Graphite component (currently on version `0.10.0-rc1`). It should also handle all *required* Python package dependencies, and where it doesn't, should document the needed vendor packages (e.g. Python or Cairo headers).

It's certainly possible there are still some rough edges, but I've run through this a couple times now and knocked out all of the obvious bugs.

If you're *not* interested in testing this out or you've moved on to a different installation process (or entirely different TSDB), that's ok too, just let me know so we can close this issue. Sincerely, thanks for your feedback.

jvanasco:
I'm deploying a new Ubuntu image next week -- I'll put this on my task list!

obfuscurity:
Awesome, thanks @jvanasco. I hope you don't mind, I'm going to close this issue for now but I'm looking forward to hearing back from you.

jvanasco:
I think this seems to be working.

glad to see pycairo replaced.  fwiw, these were my old build notes to patch pycairo to get around errors:

	cd ~/build
[...]

Issue #1425, opened 2015-12-07
Unable to syncdb or migrate dumpdata


obfuscurity:
I'm updating [Synthesize](https://github.com/obfuscurity/synthesize) to be compatible with Graphite master but I'm running into some issues with the database migration. I typically perform a manual setup of the graphite.db and then `dumpdata` to create an `initial_data.json`, which I then use in an automated fashion (`--noinput`). This has worked well with all 0.9.x releases, but I'm hitting some issues with master.

I've tried a few different `dumpdata` settings, but `syncdb` or `migrate` always complains about missing tables. Here is one example.

```
vagrant@vagrant:/opt/graphite/webapp/graphite$ sudo PYTHONPATH=/opt/graphite/webapp django-admin.py migrate --noinput --settings=graphite.settings
Operations to perform:
  Synchronize unmigrated apps: account, render, staticfiles, whitelist, metrics, url_shortener, dashboard, composer, events, browser
  Apply all migrations: admin, contenttypes, tagging, auth, sessions
Synchronizing apps without migrations:
  Creating tables...
    Creating table account_profile
    Creating table account_variable
    Creating table account_view
    Creating table account_window
    Creating table account_mygraph
    Creating table dashboard_dashboard
    Creating table dashboard_template
    Creating table events_event
    Creating table url_shortener_link
    Running deferred SQL...
  Installing custom SQL...
Traceback (most recent call last):
  File "/usr/local/bin/django-admin.py", line 5, in <module>
    management.execute_from_command_line()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 338, in execute_from_command_line
    utility.execute()
...
...
  File "/usr/local/lib/python2.7/dist-packages/django/db/backends/sqlite3/base.py", line 318, in execute
    return Database.Cursor.execute(self, query, params)
django.db.utils.OperationalError: Problem installing fixture '/opt/graphite/webapp/graphite/initial_data.json': Could not load contenttypes.ContentType(pk=1): no such table: django_content_type
```

I'm not what you'd consider a Django expert, so any feedback or suggestions would be greatly appreciated. The data is too large to paste in this report, so I've uploaded it to a gist [here](https://gist.github.com/obfuscurity/d46240979b3bebb22ccc).

/cc @brutasse @mleinart @deniszh

obfuscurity:
[...]

PR #1555, opened 2016-06-24
Pin Django <1.9

johnseekins:
django 1.9 removes the "syncdb" option that graphite's db deploy requires. Django 1.9 also no longer works with django-tagging 0.3.6, which is also a pinned requirement.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,7 +35,7 @@
 #   deactivate
 #
 
-Django>=1.4
+Django>=1.4,<1.9
 python-memcached==1.47
 txAMQP==0.4
 simplejson==2.1.6


johnseekins:
Example issues:
https://github.com/graphite-project/graphite-web/issues/1541
https://github.com/graphite-project/graphite-web/issues/1514
https://github.com/graphite-project/graphite-web/issues/1509

obfuscurity:
@drawks @brutasse @SEJeff Thoughts on the long-term viability of this?

johnseekins:
I suppose alternatively, we could work on making the package compatible with Django >= 1.9...

obfuscurity:
That would certainly be my preference but I'm not sure how ugly that's going to get. Knowing Django, probably very.

drawks:
The march of time will continue to beyond the horizon. Pinning to <1.9 seems the best option for now, but certainly working towards forward compatibility with newer Django should be the goal.

johnseekins:
So...obviously we just throw Django out for something like Bottle or Flask, right? 
:trollface:

deniszh:
I would prefer to replace Django with Flask or Bottle, indeed. But I thinks it's target for e.g Graphite 2.0 version. 
IMO we need to support at least LTS releases  and version in current popular LTS distros - e.g. Ubuntu 12.04 / 14.04, CentoOS 6 / 7 etc. - but now pinning to <1.9 is fine.

obfuscurity:
@johnseekins Thanks for this. I've opened #1573 to make sure we track a long-term fix for this issue.

johnseekins:
Awesome solution. Thanks!

I guess I'll just get started on that flask migration...

obfuscurity:
Looks like @wawrzek's tip in https://github.com/graphite-project/graphite-web/issues/1425#issuecomment-194905689 may allow us to use 1.9 again.

brutasse:
Django is required primarily for database interaction for users management and storing events. A flask migration is pointless if no solution is found for these, since it will still require a dependency such as sqlalchemy & schema migrations handling. For events, there is https://github.com/brutasse/graphite-api/pull/182 which sounds like an interesting approach (pluggable events backend). For user management, I chose to drop that part completely from graphite-api…

Happy to bring graphite-api under the graphite-project umbrella if that's desired, it could share a `libgraphite` dependency with graphite-web for core functionality & graphite-web/graphite-api would implement the HTTP layer, plus some additional persistence features for graphite-web.

obfuscurity:
@brutasse There's so much volatility in the Django release cycles that I think we're going to have to decouple ourselves from it sooner rather than later. It's become a nightmare to try and support all of the different versions >= 1.4 and <= 1.9.

I would love to see graphite-web evolve to something like graphite-api (rendering) with pluggable support for events & users, and the UI stuff (composer, dashboard) broken off into their own projects. I think bringing graphite-api into the main project would be a great step forward but there's still a ton of work to break out the other bits, make them compatible, merge in graphite-web enhancements back into graphite-api, etc.

I don't have a conclusive opinion here but I'd love to have the conversation continue.

PR #1556, opened 2016-06-24
graceful failure on bad remote store request

johnseekins:
An old PR that I'd never actually fixed. But still valid.


johnseekins:
https://github.com/graphite-project/graphite-web/pull/1560 does this cleaner. Closing.

PR #1560, opened 2016-06-26
Gracefully handle offline cluster backends

nyerup:
When finding remote metrics, a single unavailable backend causes the
entire tree to be unavailable for two consecutive requests, as
graphite/metrics/views.py's find_view() raises an AttributeError rather
than marking the backend as failed, and continuing with others.

This is ultimately because remote_storage.py's get_results() fails to
check for errors before trying to work with the response from httplib.

This patch enables all other backends' metrics to be shown immediately
after a backend fails.

### DIFF
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -81,6 +81,10 @@ class FindRequest:
       response = self.connection.getresponse(buffering=True)
     except TypeError:  # Python 2.6 and older
       response = self.connection.getresponse()
+    except AttributeError:
+      self.store.fail()
+      if not self.suppressErrors:
+        raise
 
     try:
       assert response.status == 200, "received error response %s - %s" % (response.status, response.reason)


deniszh:
Nice catch, @nyerup ! 
LGTM 👍

obfuscurity:
:+1: Thanks!

nyerup:
Hi,


I found that article a couple of days ago and thought you may be really interested in such info, just read it  here <http://guest.xtmai.com/e4zdgl>

Best Wishes, jesper@dahlnyerup.dk

nyerup:
Yo! 


The stuff I just found is unbelievable, I couldn't imagine that  something like that can exist, take a look  here <http://welcome.boltonmahomes.com/e4yetk>

Warm regards, jesper

bmhatfield:
@nyerup heads up that your email account has posted spam to this issue a couple of times. You may want to check for breach/malware/etc.

PR #1564, opened 2016-06-29
Remote reader locking fix

### LINKED ISSUE #1353: Race condition for http connection in remote_storage.py
Running latest master, making numerous concurrent requests for metrics which hit the same remote, I am seeing a lot of exceptions in the log like this:

    Exception: Failed after 0 retry! See: Passive remote fetch failed to find cached results
    Thu Oct 15 18:04:45 2015 :: Error requesting http://10.203.11.21:8888/render/?target=%2A.me1-pre.interface-%2A.if_errors.tx&format=pickle&local=1&noCache=1&from=1444921484&until=1444932284
      File "./webapp/graphite/remote_storage.py", line 182, in wait_for_results
     AttributeError: 'NoneType' object has no attribute 'getresponse'

I believe this is the result of a race condition:
* Thread 1 acquires the request lock
* Thread 1 blocks attempting to establish an http connection
* Thread 2 attempts to acquire the request lock but is unable to
* Thread 2 returns a FetchInProgress
* Thread 2 via the FetchInProgress object later calls wait_for_results()
* Thread 2 attempts to call self.connection.getresponse() on self.connection which is None
* Thread 1 successfully creates the http connection and assigns to self.connection

<del>
The effect can be mitigated by spinning to wait for self.connection to be assigned:

    179     def wait_for_results():                                           
    180       if wait_lock.acquire(False): # the FetchInProgress that gets waited on waits for the actual completion
    181         try:
    182           # spin waiting on connection        
    183           while not self.connection:
    184             pass        
    185           response = self.connection.getresponse()

But this has the undesirable effect of the waiting thread never completing if the requesting thread raises an exception.
</del>

Increasing MAX_FETCH_RETRIES may also prevent the effect from being visible to the user, at the expense of extraneous logging.

This issue may be a duplicate of #987.

EDIT: the spinning on the connection mentioned above does not work; I misunderstood how the RemoteReader class works.  I believe a recursive lock is the simplest solution (see comment below).

redbaron:
Fixes #1353 

Ensures that RemoteReader() instance which initiated requests is the one which is reading them back

### DIFF
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -122,8 +122,11 @@ class RemoteReader(object):
   __slots__ = ('store', 'metric_path', 'intervals', 'query', 'connection')
   cache_lock = Lock()
   request_cache = {}
-  request_locks = {}
   request_times = {}
+
+
+  inflight_requests = {}
+  inflight_lock = Lock()
 
   def __init__(self, store, node_info, bulk_query=None):
     self.store = store
@@ -164,9 +167,9 @@ class RemoteReader(object):
     # Despite our use of thread synchronization primitives, the common
     # case is for synchronizing asynchronous fetch operations within
     # a single thread.
-    (request_lock, wait_lock, completion_event) = self.get_request_locks(url)
+    (request_initiator, completion_event) = self.get_inflight_requests(url)
 
-    if request_lock.acquire(False): # we only send the request the first time we're called
+    if self is request_initiator: # we only send the request the first time we're called
       try:
         log.info("RemoteReader.request_data :: requesting %s" % url)
         connector_class = connector_class_selector(settings.INTRACLUSTER_HTTPS)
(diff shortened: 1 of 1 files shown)


deniszh:
Looks good at first sight, but I do not running master in PROD.
Anyone else want to check @graphite-project/committers ?

cbowman0:
I'll check it out.

cbowman0:
This looks good to me.

redbaron:
This is a total mess. Single threaded graphite keeps calling `wait_for_result` on passive readers first, which obviously never going to succeed, for whole that `RemoteReader` thing to work, order of calls to `RemoteReader.fetch` must match order of calls to `wait_for_result`, at least first call in both sequences must be the first one, if it isn't it all deadlocks for FETCH_TIMEOUT.

I have a suspicion active call might never be called at all in some cases. 

Needs to be reworked ,so that it doesn't depend on order of calls

PR #1565, opened 2016-06-30
Fix RemoteReader races (second attempt to fix #1353)

### LINKED ISSUE #1353: Race condition for http connection in remote_storage.py

redbaron:
Fixes: #1353

We tie all RemoteReaders reading from the same url to a single
ReadResult instance which memoizes response.

Whover calls `ReadResult.get` for the first time, will make read from
connection and unpickle data, rest are waiting for that
process to finish and then using a memoized result.

`ReadResult.get` also calls `done_cb` unconditionally to deterministically cleanup
`RemoteReaders.inflight_requests`.

There is no more caching of results on this level, all datapoint caching already
done elsewhere, so only puprose of `ReaderResult.inflight_requsts` is to coalesce
multiple requests to same URL, not to cache.

It's been running for few hours on fairly busy server with memcache caching disabled and for the first time
*exceptions.log* is clean


deniszh:
Hi @redbaron,
lint is not happy about your change:
` /home/travis/build/graphite-project/graphite-web/webapp/graphite/remote_storage.py:4:1: F401 'threading.Event' imported but unused
/home/travis/build/graphite-project/graphite-web/webapp/graphite/remote_storage.py:163:93: F821 undefined name 'url'
/home/travis/build/graphite-project/graphite-web/webapp/graphite/remote_storage.py:169:45: F821 undefined name 'url'
ERROR: InvocationError: '/home/travis/build/graphite-project/graphite-web/.tox/lint/bin/flake8 /home/travis/build/graphite-project/graphite-web/webapp/graphite`

redbaron:
It has all the right to be so :) Fixed missing `url` var and unused `Event` import

deniszh:
Did you see any performance regressions w/o cache, @redbaron ?

redbaron:
There is slight regression,  but I'd say it because instead of failing on a buggy code it actually fetches and processes data now.


Unpatched code, cache disabled, histogram for 3 minutes:

```
cat /var/log/httpd24/graphite-web-access.log | grep '/render' |grep -v 'local' | awk '/26\/Jun\/2016:10:4[5-7]/ {print $11}'  |histogram.py  -m 1.0 -x 5000 -b 10 -p
# NumSamples = 288; Min = 1.00; Max = 5000.00
# Mean = 89.565972; Variance = 112033.259537; SD = 334.713698; Median 3.000000
# each ∎ represents a count of 3
    1.0000 -   500.9000 [   276]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎ (95.83%)
  500.9000 -  1000.8000 [     7]: ∎∎ (2.43%)
 1000.8000 -  1500.7000 [     2]:  (0.69%)
 1500.7000 -  2000.6000 [     0]:  (0.00%)
 2000.6000 -  2500.5000 [     1]:  (0.35%)
 2500.5000 -  3000.4000 [     1]:  (0.35%)
 3000.4000 -  3500.3000 [     1]:  (0.35%)
 3500.3000 -  4000.2000 [     0]:  (0.00%)
 4000.2000 -  4500.1000 [     0]:  (0.00%)
 4500.1000 -  5000.0000 [     0]:  (0.00%)
```

With this PR applied, cache disabled:

```
cat /var/log/httpd24/graphite-web-access.log | grep '/render' |grep -v 'local' | awk '/26\/Jun\/2016:10:1[3-5]/ {print $11}'  |histogram.py  -m 1.0 -x 5000 -b 10 -p
# NumSamples = 365; Min = 1.00; Max = 5000.00
# 1 value outside of min/max
# Mean = 233.926027; Variance = 422460.643843; SD = 649.969725; Median 7.000000
# each ∎ represents a count of 4
    1.0000 -   500.9000 [   320]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎ (87.67%)
  500.9000 -  1000.8000 [    22]: ∎∎∎∎∎ (6.03%)
 1000.8000 -  1500.7000 [    11]: ∎∎ (3.01%)
 1500.7000 -  2000.6000 [     5]: ∎ (1.37%)
 2000.6000 -  2500.5000 [     1]:  (0.27%)
 2500.5000 -  3000.4000 [     2]:  (0.55%)
 3000.4000 -  3500.3000 [     0]:  (0.00%)
 3500.3000 -  4000.2000 [     1]:  (0.27%)
 4000.2000 -  4500.1000 [     1]:  (0.27%)
 4500.1000 -  5000.0000 [     1]:  (0.27%
```

cbowman0:
I gave this a test drive on my cluster and saw no issues with it.  I sporadically have the errors described and agree with the logic here to fix them.

deniszh:
OK, let's give it a try.

obfuscurity:
Huge 👍 for removing an unnecessary knob.

PR #1529, opened 2016-06-01
Fix data rereads

cbowman0:
Cherry-picked the commits to fix issue #1521 against the master branch.

My peers were experiencing this exact problem yesterday.  We are testing this fix now.

PR #1568, opened 2016-06-30
Linear complexity for movingAverage function

kamaradclimber:
The previous version recomputed the whole sum at each iterations.
When the moving average window is large (a few days), it makes computation too long.

This change make computation time linear instead.

Change-Id: I15196441aefb35a2748eb24fb432d965a79ce435

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -967,9 +967,17 @@ def movingAverage(requestContext, seriesList, windowSize):
     newSeries = TimeSeries(newName, series.start + previewSeconds, series.end, series.step, [])
     newSeries.pathExpression = newName
 
-    for i in range(windowPoints,len(series)):
-      window = series[i - windowPoints:i]
-      newSeries.append(safeAvg(window))
+    window_sum = safeSum(series[:windowPoints])
+    count = safeLen(series[:windowPoints])
+    newSeries.append(safeDiv(window_sum, count))
+    for n, last in enumerate(series[windowPoints:-1]):
+      if series[n] is not None:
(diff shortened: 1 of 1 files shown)


deniszh:
LGTM, but maybe someone more knowledgeable in movinAverage() function could take a look? Maybe @arielnh56 ? Or anybody else? @bmhatfield @cbowman0 ?

drawks:
Looks good to me as well. My only niggling complaint would be that I'd like to see the test and the refactor done as two different pulls. Since this is a refactor id' be nice to see the test green on the current code path before replacing it with the refactored one.

I mostly just watch from the cheap seats, but that seems (to me) a tad bit saner as far as a flow for refactoring.

kamaradclimber:
I'll split the PR in half no pb

kamaradclimber:
done

deniszh:
Cool! Thanks, @kamaradclimber!
Any objections to merging it? @drawks ?

obfuscurity:
@kamaradclimber Sorry, we merged some other submissions. Would you mind rebasing?

kamaradclimber:
@obfuscurity done

@drawks I've removed the commit with the test since more evolved tests have been introduced in the meantime.

obfuscurity:
@kamaradclimber Looks like the test commit got swallowed somewhere? I'm only seeing the refactor commit.

cbowman0:
I believe the tests for movingAverage() that were added as part of #1529 cover these changes.

obfuscurity:
Looks good. I ran some basic benchmarks at https://repl.it/C9kE/2 and can confirm the new algorithm works as advertised.

PR #1570, opened 2016-07-03
Replace charged language "whitelist/blacklist"

gwaldo:
Addressing https://github.com/graphite-project/graphite-web/issues/1569

Related to https://github.com/graphite-project/carbon/issues/567 / https://github.com/graphite-project/carbon/pull/568

**EDIT:** This isn't tested, but is ready for review.

Issue #1569, opened 2016-07-02
Replace racially charged language "whitelist/blacklist"


gwaldo:
I'd like to deprecate the "whitelist/blacklist" language in Graphite, replacing them with "safelist/blocklist".  (Not a new or original idea; saw this at https://github.com/kickstarter/rack-attack/pull/181.)

graphite-web and carbon are impacted.

I'm thinking that the internals could be incorporated immediately, the safe/block method interfaces become the implementation, where the white/black names log a deprecation warning (to be removed at a later date), and we read from both sets of files.

I'm guessing that the worst part will be reading in both white/black and safe/block lists and deduping them.

What do the other contributors think about this?  I certainly don't own Graphite, and don't consider this a personal crusade, but I do see this as an opportunity to be more inclusive / less exclusionary.

gwaldo:
I've opened https://github.com/graphite-project/carbon/issues/567 for the Carbon PR, but I'd like the main discussion to be here.  The graphite-web PR will be linked against this Issue.

bmhatfield:
I agree with updating antiquated terms, but I am curious: Are safelist / blocklist clear enough in this context? Could we use simpler language such as `Allow` and `Deny`?

gwaldo:
Heh, I'd been thinking about those terms, too.  :-)

I think allow/deny makes a lot more sense.  What is thought of `allowed_metrics.conf` and `denied_metrics.conf`

obfuscurity:
I'm :100: on this although the PR should aim to be backwards compatible with the old naming, at least as far as the code (not docs). As far as the new names, I prefer `block` over `deny` but I don't have any strong argument for it. I just think `blocked_metrics.conf` sounds better than `denied_metrics.conf`. 😝

gwaldo:
Yes, I have no intention of ripping the carpet out from anyone.  The idea is that the old files will still be allowed, but perhaps throw a deprecation message out to the logs?

I intend for the docs to reflect the change.  Whitelist/Blacklist will be listed as pending deprecation (something to that effect), and directing people to use the new names.

Does `allowed_metrics.conf` + `blocked_metrics.conf` make sense?

obfuscurity:
Cool, yeah a startup message to the logs would be good.

Yes, those names make sense to me.

gwaldo:
For reference, I'm replacing the general term (env vars, etc) `USE_WHITELIST` with `USE_METRICSLIST`.

[...]

PR #1572, opened 2016-07-05
Coverage for graphite.whitelist

cbowman0:
I had this mostly together, so in light of issue #1569 and pull request #1570, I submit this for test cases before the wording changes.

* Fix graphite.whitelist.load_whitelist()
* Add test cases for show, add and remove

Of note, the load_whitelist() function was not working before this change.  If the preferred method is to read the blob in and use unpickle.loads(), then that can be done.

### DIFF
--- a/webapp/graphite/util.py
+++ b/webapp/graphite/util.py
@@ -164,7 +164,7 @@ if USING_CPICKLE:
   class SafeUnpickler(object):
     PICKLE_SAFE = {
       'copy_reg': set(['_reconstructor']),
-      '__builtin__': set(['object', 'list']),
+      '__builtin__': set(['object', 'list', 'set']),
       'collections': set(['deque']),
       'graphite.render.datalib': set(['TimeSeries']),
       'graphite.intervals': set(['Interval', 'IntervalSet']),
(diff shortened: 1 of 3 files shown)


gwaldo:
Thank you for this, @cbowman0

cbowman0:
I adjusted the testing to be deterministic.  I would like someone else from @graphite-project/committers to confirm the changes to load_whitelist() is ok.

Once that is merged, I will help @gwaldo with adjusting the test cases to work with his wording changes, if required.

obfuscurity:
@cbowman0 why the change from `unpickle` to `pickle`? The latter is intended as a safe alternative to the insecure pickle.

https://github.com/graphite-project/graphite-web/blob/af7c0c13889c59a192debcb3833643df3d394597/webapp/graphite/util.py#L159-L212

cbowman0:
@obfuscurity Because it currently doesn't work.  Calls to load_whitelist() error with:
`AttributeError: type object 'SafeUnpickler' has no attribute 'load'`

obfuscurity:
Ah right, we're loading from a file, not a string. Looks like a typo from https://github.com/graphite-project/graphite-web/commit/c198e5836970f0970b96498fcbe6fa83d90110cf. Yeah, this change looks good.

👍

cbowman0:
Alternatively, we can do:
```
   buffer = open(settings.WHITELIST_FILE, 'rb').read()
   whitelist = unpickle.loads(buffer)
```
and add set to the PICKLE_SAFE set.  I can do that quickly, if that's preferred.

obfuscurity:
@cbowman0 Yeah, good idea.

cbowman0:
Done.  Once automated checks go green again, I'll merge it.

PR #1574, opened 2016-07-06
Offical project exists here. Brightcove is deprecated.

esacteksab:
From their site: https://diamond.readthedocs.io/en/latest/#repos

Historically Diamond was a brightcove project and hosted at BrightcoveOS. However none of the active developers are brightcove employees and so the development has moved to python-diamond. We request that any new pull requests and issues be cut against python-diamond. We will keep BrightcoveOS updated and still honor issues/tickets cut on that repo.

Wasn't sure if the docs (exists in this PR) or their Github repo (https://github.com/python-diamond/diamond) was more appropriate. Happy to update the PR if you'd like it to point to the repo instead.


gwaldo:
This is perfect!  Thanks, @esacteksab

PR #1409, opened 2015-11-25
Update attime to use timezones properly

bmhatfield:
**The problem**: `parseTimeReference` is not time-zone aware, preventing users from rendering graphs with their requested time zone on the x-axis properly. See #1407 for more.

**The yak**: `parseTimeReference` does some ... *timezone unfriendly* ... manipulations of a naive time object, preventing us from simply passing `tzinfo` into it in lots of edge cases.

**The solution**: make `parseTimeReference` internally work in UTC, and then have `parseATTime` localize the output into the requested (or default) timezone. (as recommended by http://pythonhosted.org/pytz/#problems-with-localtime)

PR #1576, opened 2016-07-06
Add tests to attime for when current date is Feb 29th.

cbowman0:
Tests show no issues when the current date is on or around Feb 29th in Master.

### DIFF
--- a/webapp/tests/test_attime.py
+++ b/webapp/tests/test_attime.py
@@ -363,6 +363,57 @@ class getUnitStringTest(TestCase):
         with self.assertRaises(Exception):
             result = getUnitString(1)
 
+class LeapYearMockedDateTime(datetime):
+    def __new__(cls, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


bmhatfield:
I did a bit of work on attime on 0.9.x that probably should be carried forward if they don't already exist; notably, I got rid of all the wonky TZ code and instead did everything in UTC, then converted to the requested TZ at render time via built-in libraries.

See if anything in here resonates: https://github.com/graphite-project/graphite-web/pull/1409

cbowman0:
I cherry-picked those changes to get us consistency between 0.9.x and master in #1579 .  If those get merged, then I'll rebase this against that and verify this still runs clean.  Although, the differences between 0.9.10 and master is minimal, so I'm not sure where some issue would occur with leap-year...

obfuscurity:
LGTM 👍

PR #1579, opened 2016-07-06
ATTime: Make TimeZone Aware

cbowman0:
This is cherry picked from the 0.9.x branch with changes made to the tests to accommodate the changes to graphite/render/attime.py

### DIFF
--- a/webapp/graphite/render/attime.py
+++ b/webapp/graphite/render/attime.py
@@ -39,11 +39,11 @@ def parseATTime(s, tzinfo=None):
     offset = '-' + offset
   else:
     ref,offset = s,''
-  return tzinfo.localize((parseTimeReference(ref) + parseTimeOffset(offset)), daylight)
+  return tzinfo.normalize(parseTimeReference(ref).astimezone(tzinfo) + parseTimeOffset(offset))
 
 
 def parseTimeReference(ref):
-  if not ref or ref == 'now': return datetime.now()
+  if not ref or ref == 'now': return datetime.now(pytz.utc)
 
   #Time-of-day reference
   i = ref.find(':')
@@ -66,7 +66,7 @@ def parseTimeReference(ref):
     hour,min = 16,0
     ref = ref[7:]
 
-  refDate = datetime.now().replace(hour=hour,minute=min,second=0)
+  refDate = datetime.now(pytz.utc).replace(hour=hour,minute=min,second=0)
 
   #Day reference
   if ref in ('yesterday','today','tomorrow'): #yesterday, today, tomorrow
--- a/webapp/tests/test_attime.py
+++ b/webapp/tests/test_attime.py
@@ -15,8 +15,8 @@ class MockedDateTime(datetime):
     def __new__(cls, *args, **kwargs):
         return datetime.__new__(datetime, *args, **kwargs)
     @classmethod
-    def now(cls):
-        return cls(2015, 3, 8, 12, 0, 0)
+    def now(cls, tzinfo=None):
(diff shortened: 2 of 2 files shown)


cbowman0:
There are additional differences between master and the 0.9.x branch.  Most are innocuous, but there are a few that should be analyzed.

```
$ git diff upstream/0.9.x graphite/render/attime.py | cat
diff --git a/webapp/graphite/render/attime.py b/webapp/graphite/render/attime.py
index 10405e3..9f9845e 100644
--- a/webapp/graphite/render/attime.py
+++ b/webapp/graphite/render/attime.py
@@ -12,30 +12,24 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 
+import pytz
 from datetime import datetime,timedelta
 from time import daylight
-from django.utils import timezone
-
-try: # See if there is a system installation of pytz first
-  import pytz
-except ImportError: # Otherwise we fall back to Graphite's bundled version
-  from graphite.thirdparty import pytz
-
 
 months = ['jan','feb','mar','apr','may','jun','jul','aug','sep','oct','nov','dec']
 weekdays = ['sun','mon','tue','wed','thu','fri','sat']
 
-def parseATTime(s, tzinfo=None, now=None):
+def parseATTime(s, tzinfo=None):
   if tzinfo is None:
-    tzinfo = timezone.get_current_timezone()
-
+    tzinfo = pytz.timezone(settings.TIME_ZONE)
   s = s.strip().lower().replace('_','').replace(',','').replace(' ','')
   if s.isdigit():
     if len(s) == 8 and int(s[:4]) > 1900 and int(s[4:6]) < 13 and int(s[6:]) < 32:
       pass #Fall back because its not a timestamp, its YYYYMMDD form
     else:
       return datetime.fromtimestamp(int(s),tzinfo)
-  elif ':' in s:
+  elif ':' in s and len(s) == 13:
     return tzinfo.localize(datetime.strptime(s,'%H:%M%Y%m%d'), daylight)
   if '+' in s:
     ref,offset = s.split('+',1)
@@ -50,6 +44,7 @@ def parseATTime(s, tzinfo=None, now=None):
 
 def parseTimeReference(ref):
   if not ref or ref == 'now': return datetime.now(pytz.utc)
+
   #Time-of-day reference
   i = ref.find(':')
   hour,min = 0,0
@@ -109,7 +104,7 @@ def parseTimeReference(ref):
     elif ref[-1:].isdigit():
       refDate = refDate.replace(day= int(ref[-1:]))
     else:
-      raise Exception, "Day of month required after month name"
+      raise Exception("Day of month required after month name")
   elif ref[:3] in weekdays: #DayOfWeek (Monday, etc)
     todayDayName = refDate.strftime("%a").lower()[:3]
     today = weekdays.index( todayDayName )
@@ -118,8 +113,7 @@ def parseTimeReference(ref):
     if dayOffset < 0: dayOffset += 7
     refDate -= timedelta(days=dayOffset)
   elif ref:
-    raise Exception, "Unknown day reference"
-
+    raise Exception("Unknown day reference")
   return refDate
 
 
@@ -164,4 +158,4 @@ def getUnitString(s):
   if s.startswith('w'): return 'weeks'
   if s.startswith('mon'): return 'months'
   if s.startswith('y'): return 'years'
-  raise Exception, "Invalid offset unit '%s'" % s
+  raise Exception("Invalid offset unit '%s'" % s)
```

bmhatfield:
Thanks for doing this! This all makes sense to me, and looks like it merged well!

PR #1584, opened 2016-07-11
Add suport in timezones

zacharya19:


cbowman0:
I believe this duplicates #1482. Would you agree?

obfuscurity:
@cbowman0 Agreed, and let's get that other one merged. :)

PR #1585, opened 2016-07-11
Cumulative() function has incorrect doc string

cbowman0:
* Change doc string to be correct
* Add test for cumulative()
* Add test for consolidateBy()

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -983,13 +983,16 @@ def movingAverage(requestContext, seriesList, windowSize):
 
   return result
 
-def cumulative(requestContext, seriesList, consolidationFunc='sum'):
+def cumulative(requestContext, seriesList):
   """
-  Takes one metric or a wildcard seriesList, and an optional function.
+  Takes one metric or a wildcard seriesList.
 
-  Valid functions are 'sum', 'average', 'min', and 'max'
-
-  Sets the consolidation function to 'sum' for the given metric seriesList.
+  When a graph is drawn where width of the graph size in pixels is smaller than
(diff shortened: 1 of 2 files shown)


webapp/graphite/render/functions.py
```diff
@@ -983,11 +983,9 @@ def movingAverage(requestContext, seriesList, windowSize):
 
   return result
 
-def cumulative(requestContext, seriesList, consolidationFunc='sum'):
+def cumulative(requestContext, seriesList):
   """
-  Takes one metric or a wildcard seriesList, and an optional function.
-
-  Valid functions are 'sum', 'average', 'min', and 'max'
+  Takes one metric or a wildcard seriesList
```
obfuscurity: Add a period here?

webapp/graphite/render/functions.py
```diff
@@ -983,11 +983,9 @@ def movingAverage(requestContext, seriesList, windowSize):
...
 
   Sets the consolidation function to 'sum' for the given metric seriesList.
```
obfuscurity: Do you think we should expand on this for people who might not read ahead to what `consolidateBy` does?

webapp/graphite/render/functions.py
cbowman0: Sure.

webapp/graphite/render/functions.py
cbowman0: I can copy in some of the verbiage about what consolidateBy does.  That would make it cleaner.

obfuscurity: 💯

cbowman0:
OK now?

obfuscurity:
Thanks!

Issue #1589, opened 2016-07-12
Validity of unit tests that perform assertEqual(TimeSeries, TimeSeries)


cbowman0:
I may be using the wrong words in Python for this, so please work with me on this.

TimeSeries is a subclass of list but no comparison operators are explicitly defined in it.  Because of that, I believe that the unit tests are not complete.  I suspect the assertEqual ends up comparing the list values OR the private variables but never both.

Thoughts?

SEJeff:
Comparison is generally `__cmp__()` in Python I think the terminology you want is equality. I like the correctness of a `TimeSeries.__eq__()` and see it as the right approach from a "pythonic" perspective.

I'd say fix the tests and let's merge your #1590

cbowman0:
Ok.  I pushed changes to that pull request to clean up and/or fix the existing unit tests.

PR #1590, opened 2016-07-12
Implement __eq__ for TimeSeries.

cbowman0:
Without this testing equality of two TimeSeries will compare the list contents but not the private variables

Potential fix for #1589.

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -32,6 +32,13 @@ class TimeSeries(list):
     self.options = {}
 
 
+  def __eq__(self, other):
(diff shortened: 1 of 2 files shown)


cbowman0:
Rebased against master.

I fixed the failing unit tests and cleaned up some more that were doing extra or useless tests now that there is an equality operator for TimeSeries.

cbowman0:
I believe this is ready to go.

obfuscurity:
Ok with me if it's ok with @SEJeff.

SEJeff:
Yup, I love how much crud this cleans up in the tests.

PR #1588, opened 2016-07-12
Applying #1560 fix to python 2.6 as well

roeezab:

PR #1593, opened 2016-07-12
Lower the default REMOTE_FETCH_TIMEOUT

obfuscurity:
While testing #1588 I noticed that rendering took a "long" time when `CLUSTER_SERVERS` was broken, even though the metrics were local. Our default `REMOTE_FETCH_TIMEOUT` value was set to **6.0**, an excessively high setting for most environments. Lowering this to a more sane default value of **3.0**.

@bmhatfield @drawks You guys see any reason this shouldn't go in?


obfuscurity:
Gonna go ahead and merge this but if anyone has objections please let me know and we can reconsider.

PR #1597, opened 2016-07-13
Allow setColor to recognize unquoted integers as hex values (master)

obfuscurity:
This is a port of #1596 (based on #1053) to the master branch. Please see the aforementioned for a full explanation.

P.S. It was cute of me to think I was clever enough to write any tests for `setColor` in glyph.py. /cc @cbowman0 in case he's feeling up to the challenge.


cbowman0:
I have an idea on how to do unit tests for glyph.cc, but it pretty much involves static image comparisons...

PR #1601, opened 2016-07-14
Coverage render/datalib.py

cbowman0:
DO NOT MERGE.

I'm attempting to write some unit tests explicitly for TimeSeries.  I found two things that need clarification before I continue.

1. Return value of __eq__ when `isinstance(other, TimeSeries)` returns false is NotImplemented, which appears to cause Python to fall back to using List.__eq__.  This means that `[]` == `TimeSeries('a', 0, 1, 1, [])`, which just feels wrong.

2. Is setting `valuesPerPoint > 1` known to work when iterating over a TimeSeries?  I'm seeing this error:
```
  File "/home/cbowman/git/graphite-web/webapp/graphite/render/datalib.py", line 65, in __consolidatingGenerator
    if buf: yield self.__consolidate(buf)
  File "/home/cbowman/git/graphite-web/webapp/graphite/render/datalib.py", line 76, in __consolidate
    return float(sum(usable)) / len(usable)
TypeError: unsupported operand type(s) for +: 'int' and 'list'
```

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -36,7 +36,7 @@ class TimeSeries(list):
     if isinstance(other, TimeSeries):
       return ((self.name, self.start, self.end, self.step, self.consolidationFunc, self.valuesPerPoint, self.options) ==
               (other.name, other.start, other.end, other.step, other.consolidationFunc, other.valuesPerPoint, other.options)) and list.__eq__(self, other)
-    return NotImplemented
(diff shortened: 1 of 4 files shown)


cbowman0:
Ignore point 2.  I was testing a nested list by accident ([range(0,100)] doesn't do what I thought.  Maybe we should check for that in TimeSeries?  The values *should* always be a list of numbers?

PR #1604, opened 2016-07-18
Add more render/function.py tests

cbowman0:
* Add tests for countSeries()
* Add tests for group()
* Add tests for exclude()
* Add tests for grep()
* Expand test for applyByNode to not pass newName
* Add tests for smartSummarize()
* Add tests for hitcount()
* Add test for summarize()


obfuscurity:
Yes, please! 👍

PR #1605, opened 2016-07-18
Dygraph and Rickshaw format types

obfuscurity:
This PR supersedes #1119, fixing some bugs and adding documentation for the new format types. Tested on my local Synthesize/master system.

### DIFF
--- a/docs/render_api.rst
+++ b/docs/render_api.rst
@@ -263,6 +263,8 @@ Examples:
   &format=json
   &format=svg
   &format=pdf
+  &format=dygraph
+  &format=rickshaw
 
 png
 ^^^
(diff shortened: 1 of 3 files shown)


JeanFred:
This `renderView` method is quite a beast. What would you think of refactoring it by extracting methods for each render format − at least for the new ones introduced here? (Ultimately it could be nice to have one class per format and use polymorphism?).

obfuscurity:
@JeanFred I agree that we should definitely look to refactor `views.py` (and frankly, `glyph.py` needs a lot of TLC). But for now I'd like to get these in and we can refactor as time permits.

@cbowman0 @SEJeff Do these tests look adequate to you? They're basically a copy of the existing `format=json` tests, but they already uncovered some bugs in the original implementation.

obfuscurity:
Merging this in but please feel free to correct me if the tests are insufficient.

cbowman0:
Tests look sufficient to me!

PR #1610, opened 2016-07-21
Fix regression in movingAverage

iain-buclaw-sociomantic:
#1568 caused a regression

```
    window_sum += last
TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'
```
Where for a given window + preview was half None, half values.


obfuscurity:
Good catch, thanks! /cc @kamaradclimber

kamaradclimber:
thanks!

PR #1612, opened 2016-07-21
Option to filter null datapoints from json results

obfuscurity:
This supersedes #1172, introducing the `noNullPoints` param for removing null results in `format=json` output. The original work was submitted by @kokje. I've made a small change to the parameter name and documented the feature.


obfuscurity:
Note that I'm while not enthusiastic about implementing this in the render view, I don't see a better place for it since the output formats are so diverse in the way they handle timestamps. It also makes it more challenging to devise tests.

Anyways, I think this is safe to go in. If there are any objections please let me know.

PR #1614, opened 2016-07-22
Prefer cairocffi, fall back to pycairo

obfuscurity:
Earlier commits led us to prefer installing cairocffi over pycairo, but `glyph.py` would still use pycairo first if it was available. Let's fix this oversight.

/cc @SEJeff @anotherthomas (via #341) and @brutasse who were involved in the earlier Cairo discussions.

### DIFF
--- a/check-dependencies.py
+++ b/check-dependencies.py
@@ -28,12 +28,13 @@ except ImportError:
     required += 1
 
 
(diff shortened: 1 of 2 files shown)


obfuscurity:
ref #1110 since we definitely want to prefer cairocffi if we're going to merge that too.

PR #1626, opened 2016-07-28
Interpolate across Nones in perSecond()

cbowman0:
Expanding upon pull request #717 from @bruce-lyft 

* Add tracking of the step to reflect when we're skipping over a None value.
* Add test case.


cbowman0:
That test failure smells of a timing issue.  It's not in code that this PR touched.

obfuscurity:
Yeah, restarting the job cleared it up.

obfuscurity:
It was hurting my brain for a bit so I ran through it manually with some sample sets. LGTM 👍

PR #1629, opened 2016-07-28
Support for cactiStyle unit suffix

### LINKED ISSUE #671: How can I use CactiStyle and/or get human readable Legend values ?

obfuscurity:
This PR supersedes #682, resolving a merge conflict and fixing the original change to pass tests. Fixes #671.

@toni-moreno I believe this accomplishes the same goal with less code. Please review and let me know if I left anything out.

Example:
```
target=cactiStyle(scale(collectd.graphite.filecount-whisper.files,20),"binary","b")
```

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -1407,11 +1407,12 @@ def cactiStyle(requestContext, seriesList, system=None, units=None):
   Takes a series list and modifies the aliases to provide column aligned
   output with Current, Max, and Min values in the style of cacti. Optionally
   takes a "system" value to apply unit formatting in the same style as the
-  Y-axis.
+  Y-axis, or a "unit" string to append an arbitrary unit suffix.
 
   .. code-block:: none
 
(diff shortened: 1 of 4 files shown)


cbowman0:
The legend output with units is nice.

It looks like the PR is missing test cases for the changes to both cactiStyle() and format_units(), though.  Are those planned?

obfuscurity:
@cbowman0 I thought the existing tests were adequate but if you have suggestions I'll take a stab at it.

cbowman0:
I sent you a PR to your branch for some unit tests for this.  I hope that's ok.

The cactiStyle() ones were straightforward.

The format_units() ones were written with a rather limited scope, so I expanded them some more.  Those should be sufficient for this change.

PR #1631, opened 2016-08-01
Add verticalLine function

obfuscurity:
This PR supersedes #493, introducing the `verticalLine` function by @nharkins. I've resolved merge conflicts and enhanced the function documentation a bit. Took a stab at a unit test but I'm hitting problems with the attime stuff.

/cc @cbowman0 for his usual testing brilliance

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2864,11 +2864,6 @@ def verticalLine(requestContext, ts, label=None, color=None):
   return [series]
 
 
-def horizontalLine(**kwargs):
-  # suggest deprecating constantLine
-  threshold(**kwargs)
-
-
 def threshold(requestContext, value, label=None, color=None):
   """
   Takes a float F, followed by a label (in double quotes) and a color.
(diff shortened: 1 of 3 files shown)


nharkins:
thanks! :)

cbowman0:
@jdixon I'll send you a pull request with working tests within a couple hours.

obfuscurity:
Awesome, thanks.

Jason Dixon

cbowman0:
PR sent.

Does the horizontalLine() function work?  I tried to invoke it from a test case with the same arguments as threshold() and received an error: `TypeError: horizontalLine() takes exactly 0 arguments (2 given)`

obfuscurity:
@cbowman0 Nope, looks like the evaluator might need patching to support this pattern. I'm guessing @nharkins has some changes to evaluator that weren't pushed up. Gonna remove the `horizontalLine()` bits from my branch.

cbowman0:
Another thing I noticed was that the start and end times were returned as floats. I don't remember other functions returning non integers.

obfuscurity:
Good catch. That doesn't surprise me, there are a zillion different approaches in there. I'll `int()` those and push up.

obfuscurity:
@cbowman0 Not sure what's going on now, looks like TravisCI is running builds against a different commit.

obfuscurity:
Oh never mind, something ate my `int()` fixes.

PR #1637, opened 2016-08-02
Make Timeseries equality test check for color

cbowman0:
The color attribute is not expected to always be there.  If it does exist, then check for equality.

I'm not 100% satisfied on the pythonic way of doing this sort of logic, so code cleanup suggestions is welcome.


obfuscurity:
Seems fine. Should we do something similar for `label`? 👍

obfuscurity:
Er, never mind. Yeah this is fine to merge.

PR #1638, opened 2016-08-03
Ability to GET single event in json

obfuscurity:
First small step towards REST-ifying our Events support. There are no native consumers for this but it drives me nuts that you'd have to dump all events (`/events/get_data`) to inspect a single event from the command-line.

```
$ curl -s -k -H "Accept: application/json" "https://127.0.0.1/events/4/" | json_pp
{
   "id" : 4,
   "tags" : [
      "deploy",
      "stupid"
   ],
   "what" : "Event - deploy",
   "data" : "",
   "when" : 1469714975
}
```

Need to add a test for this before we merge. Will try to get this done soonish unless @cbowman0 beats me to it.

### DIFF
--- a/webapp/graphite/events/views.py
+++ b/webapp/graphite/events/views.py
@@ -6,7 +6,9 @@ try:
 except ImportError:  # Django < 1.9
     from django.contrib.sites.models import RequestSite
 
+from django.core.exceptions import ObjectDoesNotExist
 from django.forms.models import model_to_dict
+from django.http import HttpResponseNotFound
 from django.shortcuts import render_to_response, get_object_or_404
 from django.utils.timezone import now, make_aware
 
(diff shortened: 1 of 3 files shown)


webapp/graphite/events/views.py
```diff
@@ -34,9 +34,17 @@ def view_events(request):
 
 
 def detail(request, event_id):
-    e = get_object_or_404(Event, pk=event_id)
-    context = {'event': e}
-    return render_to_response("event.html", context)
+    if request.META['HTTP_ACCEPT'] == 'application/json':
+        e = Event.objects.get(id=event_id)
```
brutasse: The 404 case isn't covered for JSON requests. It'd be nice to return a 404 JSON response if `Event.DoesNotExist` is raised.

webapp/graphite/events/views.py
```diff
@@ -34,9 +34,17 @@ def view_events(request):
...
+        e.tags = e.tags.split()
+        response = HttpResponse(
+            json.dumps(model_to_dict(e), cls=EventEncoder),
+            content_type="application/json")
+        return response
```
brutasse: `Django>=1.7` has a built-in JSON response, perhaps it'd be useful to have it in graphite-web to avoid calling json.dumps and passing content-type every time JSON is returned.

obfuscurity:
@brutasse Thanks for the feedback, I think these have been addressed now. We should definitely aim to replace other uses of `json.dumps` as we come across them going forward.

obfuscurity:
@brutasse @cbowman0 Now that master is pinned to `Django==1.9`, would you agree that we're ok pulling out the other tox envs? The following are complaining due to the use of `JsonResponse` which wasn't introduced until Django 1.7.

```
py27-django14-pyparsing2
py27-django15-pyparsing2
py27-django16-pyparsing2
py27-django16-pyparsing1
```

brutasse:
@obfuscurity one option to keep requirements loose would be to add `JsonResponse` in Graphite directly, it's only 10-15 lines.

obfuscurity:
@brutasse Do we **want** to keep requirements loose?

obfuscurity:
Humbly presented as evidence: https://github.com/graphite-project/graphite-web/pull/1555

brutasse:
@obfuscurity it's a maintenance pain… but it also allows packagers to avoid patching if the Django required isn't the Django version provided with e.g. Debian stable.

Answering also in #1555 since I missed that discussion :)

obfuscurity:
@brutasse In order to preserve sanity for 1.0.0 I'll go ahead and introduce `JsonResponse` into graphite-web for now.

PR #1639, opened 2016-08-03
Add support for globstars in standard finder

cbowman0:
Changes from @mark-5 from #1178 

* Manually merge into finders/standard.py.
* Adjust the test cases.

### DIFF
--- a/webapp/content/js/completer.js
+++ b/webapp/content/js/completer.js
@@ -41,7 +41,9 @@ MetricCompleter = Ext.extend(Ext.form.ComboBox, {
   },
 
   prepareQuery: function (queryEvent) {
-    queryEvent.query += '*';
+    if (queryEvent.query.substr(-1) != '*') {
+      queryEvent.query += '*';
+    }
   },
 
   onSpecialKey: function (field, e) {
--- a/webapp/graphite/finders/standard.py
+++ b/webapp/graphite/finders/standard.py
@@ -1,4 +1,5 @@
 import os
+import operator
 from os.path import isdir, isfile, join, basename
 from django.conf import settings
 
@@ -70,6 +71,7 @@ class StandardFinder:
     patterns = patterns[1:]
 
     has_wildcard = pattern.find('{') > -1 or pattern.find('[') > -1 or pattern.find('*') > -1 or pattern.find('?') > -1
+    using_globstar = pattern == "**"
 
     if has_wildcard: # this avoids os.listdir() for performance
       try:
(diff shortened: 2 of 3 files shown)


cbowman0:
I tested on my backup cluster and this appears to work as expected.  I did have some timeouts when I searched across all 9 of the backend servers in the cluster for something that would return a very large result set, but I think that's to be expected.

@obfuscurity do you want to try this out before we merge?

webapp/content/js/completer.js
```diff
@@ -41,7 +41,9 @@ MetricCompleter = Ext.extend(Ext.form.ComboBox, {
   },
 
   prepareQuery: function (queryEvent) {
-    queryEvent.query += '*';
+    if (queryEvent.query.substr(-1) != '*') {
```
obfuscurity: nice

webapp/graphite/finders/standard.py
```diff
@@ -80,8 +82,15 @@ def _find_paths(self, current_dir, patterns):
     else:
       entries = [ pattern ]
 
-    subdirs = [entry for entry in entries if isdir(join(current_dir, entry))]
-    matching_subdirs = match_entries(subdirs, pattern)
+    if using_globstar:
+        matching_subdirs = map(operator.itemgetter(0), os.walk(current_dir))
```
obfuscurity: Shouldn't we check for `isdir` here too?

cbowman0: I tested independently and only saw directories show up.

Output is only directories:
```
$ python2.7 test.py | xargs -L1 -I{}  stat -c%F {} | uniq -c
  27177 directory
```

It did take 1.1 seconds to just run the osdir on this host, but it is at about 65% IO utilization all the time.
```
$ time python2.7 test.py  | wc -l
27177

real	0m1.149s
user	0m0.675s
sys	0m0.478s
```

Test script:
```
$ cat test.py
import os
import operator

current_dir='/opt/graphite/storage/whisper/carbon/'

matching_subdirs = map(operator.itemgetter(0), os.walk(current_dir))

for i in matching_subdirs:
  print i
```

obfuscurity:
👍

webapp/graphite/finders/standard.py
mark-5: According to the [docs](https://docs.python.org/3/library/os.html#os.walk), os.walk returns a list of tuples like `(dirpath, dirnames, filenames)`, so I figured that first item we are getting is always a dir, and doesn't need an `isdir` check.

obfuscurity: Ah ok, missed that.

PR #1643, opened 2016-08-08
Tagging mixin to support set return types

### LINKED ISSUE #1210: event get_data return all events when tags are combined from not existing and existing ones.
For example when I make query like:
`/events/get_data?from=-7d&until=now&tags=foo` i got nothing, but:
`/events/get_data?from=-7d&until=now&tags=existing_tag%20foo` returns multiple results.
I think that expected behaviour would be also returning nothing.

obfuscurity:
This introduces two new filtering options for searching events+tags. Turns out that graphite-web has used django-tagging's `with_all()`, returning a "lazy union" that is surprisingly unintuitive (see #1210).

With this change we're adding a mixin to expose django-tagging's `get_intersection_by_model`, and extending our Events model to accept a `set` param that lets the user select among the different set types:

* union
* intersection
* "lazy union" (legacy, default)

Fixes #1210.

### DIFF
--- a/webapp/tests/test_events.py
+++ b/webapp/tests/test_events.py
@@ -87,6 +87,57 @@ class EventTest(TestCase):
         self.assertEqual(error, {'error': '"tags" must be an array'})
         self.assertEqual(Event.objects.count(), 0)
 
+    def test_tag_sets(self):
+        creation_url = reverse('graphite.events.views.view_events')
+        events = [
+            {
+                'what': 'Something happened',
+                'data': 'more info',
+                'tags': ['foo'],
+            },
+            {
+                'what': 'Something else happened',
+                'data': 'even more info',
+                'tags': ['bar'],
+            },
+            {
+                'what': 'A final thing happened',
+                'data': 'yet even more info',
+                'tags': ['foo', 'bar'],
+            },
+        ]
+
+        for event in events:
+            response = self.client.post(creation_url, json.dumps(event),
+                                        content_type='application/json')
+            self.assertEqual(response.status_code, 200)
+
+        url = reverse('events_get_data')
+
+        # should match two events using old set logic
+        response = self.client.get(url, {'tags': 'foo'})
+        self.assertEqual(response.status_code, 200)
+        events = json.loads(response.content)
+        self.assertEqual(len(events), 2)
+
+        # should match one event using set intersection
+        response = self.client.get("%s%s" % (url, '?tags=foo%20bar&set=intersection'))
+        self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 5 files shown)


webapp/tests/test_events.py
```diff
@@ -87,6 +87,57 @@ def test_tag_as_str(self):
...
+        self.assertEqual(len(events), 2)
+
+        # should match one event using set intersection
+        response = self.client.get("%s%s" % (url, '?tags=foo%20bar&set=intersection'))
+        self.assertEqual(response.status_code, 200)
+        events = json.loads(response.content)
+        self.assertEqual(len(events), 1)
+
+        # should match zero events using set intersection and unknown tag
+        response = self.client.get("%s%s" % (url, '?tags=foo%20bar%20nope&set=intersection'))
```
obfuscurity: @cbowman0 Any idea why this keeps returning one match instead of zero? The `curl` equivalent of this test works as expected so I'm not sure why it keeps returning an unwanted result.

torkelo:
Interesting, what http api changes would this introduce? Also thought the plan was to remove the Django dependency post 1.0

webapp/tests/test_events.py
cbowman0: It's returning this data `[{"what": "A final thing happened", "tags": ["foo", "bar"], "id": 3, "data": "yet even more info", "when": 1470740187}]`

The tags `[u'nope', u'foo', u'bar']` are being passed all the way down into the find_events method.  I haven't checked past there, but it appears that the with_intersection call isn't working as expected.

cbowman0: https://github.com/flebel/django-tagging/issues/157 ?

obfuscurity:
@torkelo This would add a `set` parameter to the events `get_data` endpoint. The parameter would support two new filter sets: `union` and `intersection`. Without specifying `set`, the behavior will not change.

Yes, the plan is to remove Django from the scenario post-1.0 but that doesn't preclude us from adding new functionality now to fix [bugs](https://github.com/graphite-project/graphite-web/issues/1210).

webapp/tests/test_events.py
obfuscurity: You're right, my CLI tests were invalid. It does indeed return a set even with unknown tags. I'll look deeper into this today, may require another patch to tagging.

obfuscurity:
How stupid am I to use a reserved function name for a variable. Fixed that. Also, I feel dirty but this change to implement manual filtering for intersections was the best non-yucky way to fix tagging's broken implementation.

obfuscurity:
This feels ready to merge. Any last minute objections?

PR #1652, opened 2016-08-18
Add a refresh button to Metrics tree

### LINKED ISSUE #723: Allow updating the tree / list of metrics
When I start submitting a new metric from a host, I have to refresh graphite to update the tree. This means I loose the graph I'm composing. It would be nice if there was a asynchronous way to update the tree.

obfuscurity:
This PR supersedes #431, resolving conflicts and squashing one minor bug. Fixes #723.


obfuscurity:
Note that this works well for branch nodes, but it doesn't seem to pick up any additions to the root nodes. Investigating.

PR #1659, opened 2016-08-22
powSeries function

leoleovich:
I needed it to analyze `input - output` chain of my services.
Example

| IN | OUT | OUT^IN | Conclusion |
| --- | --- | --- | --- |
| 1 | 1 | 1 | OK |
| 0 | 0 | 1 | OK |
| 2 | 0 | 0 | BAD |
| 3 | 2 | 8 | OK |

At the end I have graph like this

Before 12 you can see that something was comming, but noting was going out, but later on at 12 something got fixed and data went out off the chain.
We are using it for alerting.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -872,6 +872,35 @@ def pow(requestContext, seriesList, factor):
       series[i] = safePow(value,factor)
   return seriesList
 
+def powSeries(requestContext, *seriesLists):
+  """
+  Takes two or more series and pows their points. A constant line may be
+  used.
+
+  Example:
+
+  .. code-block:: none
+
+    &target=powSeries(Server.instance01.app.requests, Server.instance01.app.replies)
+
+
+  """
+
+  try:
+    (seriesList,start,end,step) = normalize(seriesLists)
+  except:
+    return []
+  name = "powSeries(%s)" % formatPathExpressions(seriesList)
+  values = []
+  for row in izip(*seriesList):
+    tmpVal = 1
+    for element in row:
+      tmpVal = safePow(tmpVal, element)
+    values.append(tmpVal)
+  series = TimeSeries(name,start,end,step,values)
+  series.pathExpression = name
+  return [series]
+
 def squareRoot(requestContext, seriesList):
   """
   Takes one metric or a wildcard seriesList, and computes the square root of each datapoint.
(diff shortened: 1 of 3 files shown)


obfuscurity:
The function needs to be exposed to the Composer UI in https://github.com/graphite-project/graphite-web/blob/master/webapp/content/js/composer_widgets.js#L1067 and it would be great to see a test included.

obfuscurity:
@leoleovich Just fyi in case you haven't noticed but the test is failing.

leoleovich:
@obfuscurity thanks, I noticed :) I will fix it.

leoleovich:
I will write when I am fully sure it is fixed :)

leoleovich:
Now I have `OverflowError: math range error` sometimes.

leoleovich:
@obfuscurity so:
I had problems with overfloating:

for integers
```
math.pow(200, 134)
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-28-4a942691ecaa> in <module>()
----> 1 math.pow(200, 134)

OverflowError: math range error
```

for floats
```
math.pow(200.1, 134.1)
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-38-3dcd5b78efc0> in <module>()
----> 1 math.pow(200.1, 134.1)

OverflowError: math range error
```

with `**` is it kinda better
```
In [29]: 200 ** 134
Out[29]: 217780714829400616616559748756331655331840000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000L
```
But only for integers. Floats broken the same way
```
200.1 ** 134.1
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-39-48bd160a010d> in <module>()
----> 1 200.1 ** 134.1

OverflowError: (34, 'Numerical result out of range')
```

But these numbers are very small and having broken graph is not the best option.
So far I put `None` if we catch exception in `safePow`.

webapp/graphite/render/functions.py
```diff
@@ -65,7 +65,7 @@ def safePow(a, b):
   if b is None: return None
   try:
     result = math.pow(a, b)
-  except ValueError:
+  except:
```
leoleovich: Would be very nice if you can preserve this part.

leoleovich: Other option would be something like
```
qqq = Decimal(323.4500000000001) ** Decimal(372.85)
print(qqq)
6.047806945385690914856988084E+935
print(float(qqq))
inf
```

deniszh:
Looks like should be rebased for having working tests...

deniszh:
Will try to carefully merge this

deniszh:
Tests are looks green

PR #1662, opened 2016-08-23
Sanitize graph urls for XSS

obfuscurity:
This PR supersedes #1470. Thanks to @Sinkmanu for finding this and submitting the initial patch. I've reworked it a bit so that it handles multiple `target` params.

### DIFF
--- a/webapp/graphite/browser/views.py
+++ b/webapp/graphite/browser/views.py
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License."""
 
 import re
+
 from django.conf import settings
 from django.shortcuts import render_to_response
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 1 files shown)


obfuscurity:
Note that this is a bit of a band-aid imho. I would rather move this logic to a custom model, which would also allow us to add a real test case. But for now I'd like get this merged and we can circle back later.

obfuscurity:
Note that what I mean is that by inserting the graph directly via the ORM, we bypass thhe view logic, e.g.:

```
user.profile.mygraph_set.create(name=u'test_graph', url='target=secondYAxis(carbon.agents.test-db1.memUsage)')
response = self.client.get(url, {'path': ''})
self.assertEqual(response.status_code, 200)
[leaf] = json.loads(response.content)
self.assertEqual(leaf['graphUrl'], u'target=secondYAxis%28carbon.agents.test-db1.memUsage%29')
```

The result comes back without encoding since we added it directly via the model.

```
[   {   u'allowChildren': 0,
        u'expandable': 0,
        u'graphUrl': u'target=secondYAxis(carbon.agents.test-db1.memUsage)',
        u'id': u'2a812540041e8cef62b37dec5f8c6196',
        u'leaf': 1,
        u'text': u'test_graph'}]
```

Note that I've also tried adding it with the test client `get` but it fails to save the graph for an unknown reason.

```
response = self.client.get(url, {'action': 'save', 'graphName': 'test_graph', 'url': 'target=secondYAxis(carbon.agents.test-db1.memUsage)'})
```

```
[   {   u'allowChildren': 0,
        u'expandable': 0,
        u'id': u'no-click',
        u'leaf': 1,
        u'text': u'No saved graphs'}]
```

This is all just FYI for later.

obfuscurity:
Proposed test:

```
diff --git a/webapp/tests/test_browser.py b/webapp/tests/test_browser.py
index f271720..5e9c2c8 100644
--- a/webapp/tests/test_browser.py
+++ b/webapp/tests/test_browser.py
@@ -62,3 +62,19 @@ class BrowserTest(TestCase):
         self.assertEqual(response.status_code, 200)
         [leaf] = json.loads(response.content)
         self.assertEqual(leaf['text'], u'fòo')
+
+    def test_encoded_graph_name(self):
+        url = reverse('browser_my_graph')
+        user = User.objects.create_user('test', 'test@example.com', 'pass')
+        self.client.login(username='test', password='pass')
+        response = self.client.get(url, {'path': ''})
+        self.assertEqual(response.status_code, 200)
+
+        response = self.client.get(url, {'action': 'save', 'graphName': 'test_graph', 'url': 'target=secondYAxis(carbon.agents.test-db1.memUsage)'})
+        self.assertEqual(response.status_code, 200)
+
+        response = self.client.get(url, {'path': ''})
+        self.assertEqual(response.status_code, 200)
+        [leaf] = json.loads(response.content)
+        self.assertEqual(leaf['graphUrl'], u'target=secondYAxis%28carbon.agents.test-db1.memUsage%29')
+
```

PR #1668, opened 2016-09-01
setup.py: make it works with python3 based tox.

iksaif:
Make tox work if it has been installed to use python3 (like
it is in recent ubuntu).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,15 +3,13 @@
 from __future__ import with_statement
 
 import os
-import ConfigParser
+try:
+    import configparser as ConfigParser
+except ImportError:
+    import ConfigParser
 
 from glob import glob
 from collections import defaultdict
-
-try:
-    from io import BytesIO
-except ImportError:
-    from StringIO import StringIO as BytesIO
 
 # Graphite historically has an install prefix set in setup.cfg. Being in a
 # configuration file, it's not easy to override it or unset it (for installing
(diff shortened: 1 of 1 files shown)


deniszh:
👍

obfuscurity:
I don't see any obvious regressions for Python 2.x but I wonder why we'd _want_ this in there? Are you running Graphite successfully with Python 3.x?

iksaif:
I'm not running Graphite with Python 3, but my `tox` command seems to use python3 to run setup.py

obfuscurity:
Maybe something wonky with your python environment? IMHO you should be [developing and running your tests in a virtualenv](http://graphite.readthedocs.io/en/latest/development.html#setting-up-a-development-environment) anyways.

obfuscurity:
Regardless I don't see any problems with this going in for now with the expectation that One Day :tm: we'll support Python 3. 😜

obfuscurity:
@iksaif I'm tempted to merge it, but given that it's largely an unnecessary change and that the same hasn't been submitted for the other projects, I'm going to pass on this for now. 

@deniszh You have any strong opinion on this?

deniszh:
@obfuscurity - I also have no idea why this change is needed for @iksaif 
So - ¯\_(ツ)_/¯

obfuscurity:
Ok I'm closing this one now but I'm gonna hold onto it like that last piece of bubble gum for when we really need it. :wink:

iksaif:
To add more context: this change is needed to simply be able to run "tox" on modern systems (such as my ubuntu box where python3 is default). Running tox in a virtualenv works, but is kind of counter-intuitive because tox will then install more virtualenv to run the tests.

obfuscurity:
Modern systems or not, Python 2.x is still a requirement for Graphite. Our own tox.ini requires Python 2.7 as the `basepython`.

Issue #1671, opened 2016-09-02
Sporadic query results - data returned may sometimes be the data for times prior to the actual query.


iliapolo:
**Scenario**

In our system, we insert data to graphite in an hour based manner.
That is, a periodic flush procedure inserts data for ``14:00-15:00``, ``15:00-16:00``, and so on...
Concurrently, queries are being performed against graphite-web, again in an hour based manner.

The querying process will frequently query a time frame that has yet to be inserted, so for example it might query data for ``16:00-17:00``, even though that window still hasn't been inserted. (it might never will be) 

**Expected result**

The expected result in such a case is for graphite-web to return ``None`` values for the all data points in the query.

**Witnessed result**

Sometimes, the result of such a query will return the data points for the ``15:00-16:00`` window instead.

This is a major issue in my opinion since it deals with correctness of results.

**Environment**

We are using the [hopsoft container](https://hub.docker.com/r/hopsoft/graphite-statsd/) (latest version - 0.9.15). 

Digest: 

``sha256:a14e56af7f6193d05cc6d5b3dcaa08109a32e7c3c12222897dc9a30dbfb03277``

iliapolo:
This issue is also present in the master branch.

PR #1672, opened 2016-09-02
Fix datalib mergeResults function to properly handle time frames that are in the future compared to the cached results

iliapolo:
This PR resolves issue #1671 

The problem was that if it takes a long time writing data to disk, the carbon-cache will hold the results in memory. When graphite-web asks carbon for its data, it gives him the entire cache, which may hold older data than the query's time frame. The ``mergeResults`` function doesn't handle this scenario properly, and will mistakenly return the cache data, even though it is irrelevant.

Note that this issue exists in [master](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/readers.py#L144) as well, however, only in the ``CeresReader``.

Should i create a PR against master as well?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,6 @@ dropin.cache
 *.egg-info
 docs/_build
 *.log
+*.iml
+storage
 *.pyc
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -496,6 +496,13 @@ def mergeResults(dbResults, cacheResults, lowest_step):
 
     try:
       i = int(interval - start) / step
+      if i < 0:
+          # cached data point is earlier then the requested data point.
+          # meaning we can definitely ignore the cache result.
+          # note that we cannot rely on the 'except'
+          # in this case since 'values[-n]=' is
+          # is equivalent to 'values[values.length - n]='
+          continue
       values[i] = value
     except:
       pass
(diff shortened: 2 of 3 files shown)


deniszh:
Nice catch! Yes, please create same for the master, much appreciated!

iliapolo:
Cool, ill do that shortly.

webapp/graphite/render/datalib.py
```diff
@@ -496,6 +496,13 @@ def mergeResults(dbResults, cacheResults, lowest_step):
 
     try:
       i = int(interval - start) / step
+      if i < 0:
+          # cached data point is earlier then the requested data point.
+          # meaning we can definitely ignore the cache result.
+          # note that we cannot rely on the 'except'
+          # in this case since 'values[-n]=' is
+          # is equivalent to 'values[values.length - n]='
```
obfuscurity: Great comment here, thanks.

.gitignore
```diff
@@ -7,4 +7,6 @@ dropin.cache
 *.egg-info
 docs/_build
 *.log
+*.iml
+storage
```
obfuscurity: Did you mean to pull these in too?

obfuscurity: Oops, disregard. I see your explicit commit. :smile:

iliapolo:
@obfuscurity thanks for the quick merge! :)

@deniszh regarding the PR against master, should i create a corresponding issue? or just edit #1671 and reference it?

.gitignore
iliapolo: :+1:

deniszh:
@iliapolo - up to you. Issue edit is fine.

PR #1685, opened 2016-09-06
Update tools.rst

strajansebastian:
Added a link to the URL for more info, as suggested. The graphite feature can be seen as an embedded Graphios script into the icinga2 software that automatically feeds performance data to the carbon daemon.


obfuscurity:
The url link and section are still wrong, fixed in #1686.

PR #1694, opened 2016-09-14
Add install_requires for pip

obfuscurity:
When installing via pip we should handle these dependencies automatically, like Carbon does with Twisted.


obfuscurity:
refs #1509

Issue #1690, opened 2016-09-08
Webapp cluster performance issue


Kenterfie:
**Configuration**
OS: Ubuntu 14.04
Webserver: nginx
WSGI: uwsgi
Graphite: graphite (latest git)

2 Graphite storage server + 1 Graphite Webapp Render Host (only json output)

**Issue**
We save our data on two different graphite storage hosts over statsite + carbon-c-relay . Everything is working well, but one problem we notice is, that every request needs very long on webapp render host.

Rendering performance
Small request needs >1000ms on webapp render host, but the real rendering on the storage host needs only ~15ms. What does graphite webapp the other 980ms on the render host?

When we call the same render request over wget from the render host it needs only 180ms not 1000ms. Someone else notice render performance issues like this? I havnt found any reason in graphite currently, what cause it.

deniszh:
When you say 'latest git' is it really latest, i.e. 2aee2bc ? And it includes 2cc2a2a ?

Kenterfie:
I have everything updated 3 hours ago, so yes. The latest commit is 2aee2bc, but the problem was with 0.9.15 the same.

deniszh:
>Small request needs >1000ms on webapp render host
>When we call the same render request over wget from the render host it needs only 180ms not 1000ms. 

Sorry, looks really confusing, do not really understand which request takes what. Could you maybe give exact queries?

deniszh:
>I have everything updated 3 hours ago, so yes. The latest commit is 2aee2bc, but the problem was with 0.9.15 the same.

Ah, then maybe it's something else.  Did you have `REMOTE_STORE_MERGE_RESULTS` enabled on 0.9.15 ?

deniszh:
Another (probably stupid) question - are you sure that you really running on master code now, and not on still 0.9.15 ?

Kenterfie:
In my case not necessary. I don't use replication between the two graphite hosts. Both hosts holding different data. I use the render node to have one api endpoint.

And yes, im very sure i use the latest version. I have compared my webapp folder with the latest git master.

Kenterfie:
On render host
Thu Sep 08 17:27:57 2016 :: Retrieval of sortByMaxima(summarize(..)) took 0.900179
Thu Sep 08 17:27:57 2016 :: Retrieval of sortByMaxima(summarize(..)) took 0.893999
[...]

PR #1699, opened 2016-09-15
Implement threading for all fetches.

cbowman0:
Since there has been lots of talk of parallel fetches, this implements it for all fetches.  This code isn't 100% clean and needs testing, but that's why I put it here for others to test too as we attempt to prove cluster performance issues.
#1063
#1690


deniszh:
Retrigger test, green now.
@Kenterfie @reiyjar @bmhatfield - please check it out.

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
...
+        temp_fetches_running = []
+        for fetch_thread in fetches_running:
+          if fetch_thread.is_alive():
+            temp_fetches_running.append(fetch_thread)
+
+        # Induce some sort of sleep?
+          # Would be nice to wait on a semaphore...
+        if len(fetches_running) >= settings.MAX_FETCH_THREADS:
+          for fetch_thread in fetches_running:
+            fetch_thread.join(10)
```
bmhatfield: This timeout should be a setting, and not hardcoded.

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
...
+        fetches_running = temp_fetches_running
+
+      for fetch_thread in fetches_running:
+        fetch_thread.join(10)
```
bmhatfield: Same

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
   startTime = int( epoch( requestContext['startTime'] ) )
   endTime   = int( epoch( requestContext['endTime'] ) )
 
+  def _fetchDataWrapper(node, startTime, endTime, result_queue):
+    result = node.fetch(startTime, endTime)
+    result_queue.put((node, result))
+
   def _fetchData(pathExpr,startTime, endTime, requestContext, seriesList):
+    result_queue = Queue.Queue()
```
bmhatfield: I'm a little confused about the pre-existing threading model and these changes. Are these fetches queued up / threaded in some other place? Or was this a missed opportunity that you're now fixing?

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
...
+
+        # Rebuild list of running fetches
+        temp_fetches_running = []
+        for fetch_thread in fetches_running:
+          if fetch_thread.is_alive():
+            temp_fetches_running.append(fetch_thread)
+
+        # Induce some sort of sleep?
+          # Would be nice to wait on a semaphore...
+        if len(fetches_running) >= settings.MAX_FETCH_THREADS:
```
bmhatfield: Why would this be the case if we only start `settings.MAX_FETCH_THREADS`?

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
...
+        fetches_running = temp_fetches_running
+
+      for fetch_thread in fetches_running:
+        fetch_thread.join(10)
+
+    while not result_queue.empty():
+      try:
+        (node, results) = result_queue.get(False)
+      except:
+        log.exception("result_queue not empty, but unable to retrieve results")
```
bmhatfield: Probably worth logging the actual exception in this case?

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
...
     matching_nodes = STORE.find(pathExpr, startTime, endTime, local=requestContext['localOnly'])
-    fetches = [(node, node.fetch(startTime, endTime)) for node in matching_nodes if node.is_leaf]
 
-    for node, results in fetches:
+    if settings.MAX_FETCH_THREADS < 1:
+      for node in matching_nodes:
+        if node.is_leaf:
+          _fetchDataWrapper(node, startTime, endTime, result_queue)
+    else:
+      fetches = []
```
bmhatfield: This variable is probably more clearly named `pending_fetches`

webapp/graphite/render/datalib.py
```diff
@@ -111,11 +114,60 @@ def fetchData(requestContext, pathExpr):
...
+        if len(fetches_running) < settings.MAX_FETCH_THREADS:
+          fetch_thread = fetches.pop()
+          fetch_thread.start()
+          fetches_running.append(fetch_thread)
+
+        # Rebuild list of running fetches
+        temp_fetches_running = []
+        for fetch_thread in fetches_running:
+          if fetch_thread.is_alive():
+            temp_fetches_running.append(fetch_thread)
```
bmhatfield: Why not just track the index in `fetches_running` and delete from the existing array, rather than building up a duplicate?

webapp/graphite/render/datalib.py
cbowman0: The existing model is asynchronous fetches to remote stores (cluster servers) and serial fetches for local (whisper files).  The asynchronous responses were then serially handled in-line with the local fetches, if any existed.

I put this patch together because I was looking at long fetch times for serially opening whisper files + carbonlink over thousands of datafiles.  This is an extreme case, but something I was looking to see if this helped out.  I do not have number yet to confirm if it does.

If this does help, then we should probably remove the async callback from remote fetches and just use threading, if threading is enabled, and not mix the two styles.

webapp/graphite/render/datalib.py
cbowman0: Sure.

webapp/graphite/render/datalib.py
cbowman0: I started with `for i, fetch_thread in enumerate(fetches_running)`, but when doing a remove on the index, but that's broken logic.  So, I went with this, but am not tied to it.

webapp/graphite/render/datalib.py
cbowman0: It should never be greater than `settings.MAX_FETCH_THREADS`, so that's just defensive coding.  If it's equal, then I was thinking it was better to not spin the main process waiting for a worker thread to finish.

Was there another question?

webapp/graphite/render/datalib.py
cbowman0: Yes.  The question that I hadn't yet answered was what value to use.  I didn't want to create another variable, but to use an existing it'd be dependent on the type of fetch (local or remote) that it was.

webapp/graphite/render/datalib.py
cbowman0: Agreed.

obfuscurity:
Any useful results from your tests?

obfuscurity:
I also just ran a basic test against a 1:1 cluster. Like @Kenterfie, I'm also seeing even worse performance with this patch applied.

deniszh:
I think https://github.com/graphite-project/graphite-web/pull/1818 superseding this. Closing, I hope you do not mind, @cbowman0 ?

cbowman0:
I do not mind.  #1818 superceded it.

PR #1705, opened 2016-09-26
average/current/max/min Above/Below fails if there is no data.

iain-buclaw-sociomantic:
- Adds check for None before integer comparisons.
- Converts list comprehension code into loop.
- Add tests for averageAbove/Below()
- Add tests for currentAbove/Below()
- Add tests for maximumAbove/Below()
- Add tests for minimumAbove/Below()

N.B: These functions only fail when using Python 3, not 2.7.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -1695,7 +1695,8 @@ def maximumAbove(requestContext, seriesList, n):
   """
   results = []
   for series in seriesList:
-    if safeMax(series) > n:
+    val = safeMax(series)
+    if val is not None and val > n:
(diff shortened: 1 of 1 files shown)


deniszh:
Running tests on python3 and app itself on 2.7... ?

iain-buclaw-sociomantic:
Hmm, I didn't check to see that there are already a test for all `None` series.  I guess I should tweak the behaviour then.

@deniszh - Running graphite-api, but the same problem is present here also.

iksaif:
any news ?

obfuscurity:
LGTM 👍

Issue #1645, opened 2016-08-11
Feature request: support for jump_fnv1a_ch hashing algorithm


piotr1212:
The consistent hashring sharding is very unbalanced as discussed here:
https://github.com/graphite-project/carbon/issues/485
@grobian implemented jump_fnv1a_ch in https://github.com/grobian/carbon-c-relay which is better balanced. 

It would be nice if graphite-web would support jump_fnv1a_ch.

obfuscurity:
Please re-open this issue under the https://github.com/graphite-project/carbon project.

piotr1212:
My request is for graphite-web to support jump_fnv1a so I can use graphite-web in combination with carbon-c-relay in jump_fnv1a_ch mode. Now graphite-web's carbonlink queries fail because they query the wrong cache. 

From what I can tell the hashring (used for carbonlink queries) is implemented here:
https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/hashing.py

I could open a second issue on carbon to request it implemented in carbon-relay, but I personally don't need it as I don't use carbon-relay.  I'm not sure if I understand why this issue is closed.

obfuscurity:
My apologies, I misunderstood your request.

piotr1212:
Thinking more about it, it would make sense to have the ConsistentHashRing class not duplicated in the two projects. ConsistentHashRing could be defined in carbon and imported in graphite-web (unless you want to prevent dependencies between the projects). 

A quick check between https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/hashing.py#L45-L100 and https://github.com/graphite-project/carbon/blob/master/lib/carbon/hashing.py#L8-L62 shows that there are slight differences. I haven't looked into details so don't know if there are issues, but I would expect it to be identical.

deniszh:
Both carbon and graphite-web has support of user keyfunc for hashing, not sure if that enough for proper fnv1a support, suspecting that's not...

bemeyert:
We really could use this feature. We use the `fnv1a_ch` from Carbon-C-Relay to make our cluster easy to scale out, since we are not dependent on hostnames and ports.

deniszh:
I added FNV1a support in PR https://github.com/graphite-project/graphite-web/pull/1723

piotr1212:
Amazing!! thanks

obfuscurity:
Fixed by #1723.

Civil:
[...]

PR #1723, opened 2016-10-14
Support for fnv1a_ch hashing

deniszh:
Adding support for FNV1a_ch hashing from https://github.com/grobian/carbon-c-relay for graphite-web. Except from its docs:
>"The fnv1a_ch cluster is a identical in behavior to carbon_ch, but it uses a different hash technique (FNV1a) which is faster but more importantly defined to get by a limitation of carbon_ch to use both host and port from the members. This is useful when multiple targets live on the same host just separated by port. The instance that original carbon uses to get around this can be set by appending it after the port, separated by an equals sign, e.g. 127.0.0.1:2006=a for instance a. When using the fnv1a_ch cluster, this instance overrides the hash key in use. This allows for many things, including masquerading old IP addresses, but mostly to make the hash key location to become agnostic of the (physical) location of that key. For example, usage like 10.0.0.1:2003=4d79d13554fa1301476c1f9fe968b0ac would allow to change port and/or ip address of the server that receives data for the instance key. Obviously, this way migration of data can be dealt with much more conveniently"

Jump_fnv1a_ch is better with metrics distribution, but it's more complicated, will add it later, if needed.

Fixing https://github.com/graphite-project/carbon/issues/602 and partially - https://github.com/graphite-project/graphite-web/issues/1645


deniszh:
I can add more tests if needed - now I added only for changed functions. Tests data is taken from `carbon-c-relay` output, so, it should be compatible.

docs/config-local-settings.rst
```diff
@@ -363,6 +363,11 @@ CARBONLINK_TIMEOUT
 
   Timeout for carbon-cache cache queries in seconds.
 
+CARBONLINK_HASHING_TYPE
+  `Default: carbon_ch`
+
+  Type of metric hashing function. Graphite-web has limited support of "fnv1a_ch" hashing from carbon-c-relay. Please note, that fnv1a_ch hashing will work here only if instance is defined.
```
obfuscurity: Would be useful to link to the external project/docs. Also, what does `instance` refer to here?

obfuscurity: Can we get an explicit list of supported values?

obfuscurity: `s/support of/support for/`

webapp/graphite/local_settings.py.example
```diff
@@ -285,6 +285,10 @@
 #CARBONLINK_HOSTS = ["127.0.0.1:7002:a", "127.0.0.1:7102:b", "127.0.0.1:7202:c"]
 #CARBONLINK_TIMEOUT = 1.0
 #CARBONLINK_RETRY_DELAY = 15 # Seconds to blacklist a failed remote server
+#
+# Graphite-web has limited support of "fnv1a_ch" hashing from carbon-c-relay
+# Please note, that fnv1a_ch hashing will work here only if instance is defined.
+#CARBONLINK_HASHING_TYPE = 'carbon_ch'
```
obfuscurity: We should document all supported values.

webapp/graphite/render/hashing.py
```diff
@@ -43,25 +61,32 @@ def compactHash(string):
...
+      small_hash = int(big_hash[:4], 16)
     return small_hash
 
   def add_node(self, key):
     self.nodes.add(key)
     self.nodes_len = len(self.nodes)
     for i in range(self.replica_count):
       replica_key = "%s:%d" % (key, i)
+      if self.hash_type == 'fnv1a_ch':
+        replica_key = "%d-%s" % (i, key[1])
```
obfuscurity: I would prefer an explicit `if else` here unless this is considered more pythonic.

deniszh:
Tried to fix the docs, @obfuscurity 
Not sure if good now. Suggestions are welcome. :)

docs/config-local-settings.rst
```diff
@@ -363,6 +363,14 @@ CARBONLINK_TIMEOUT
 
   Timeout for carbon-cache cache queries in seconds.
 
+CARBONLINK_HASHING_TYPE
+  `Default: carbon_ch`
+
+  `Possible values: carbon_ch, fnv1a_ch`
+
+  The default is "carbon_ch" - usual Graphite hashing function. Also, you can use "fnv1a_ch" hashing from `carbon-c-relay relay <https://github.com/grobian/carbon-c-relay>`_. Fnv1a_ch cluster uses instance name as the solely source of routing information. As described in "carbon-c-relay" documentation:
```
obfuscurity: I would propose this instead, also removing the 2nd paragraph:

```
The default `carbon_ch` is Graphite's traditional consistent-hashing implementation. Alternatively, you can use the `fnv1a_ch`, which supports the Fowler–Noll–Vo hash function (FNV-1a) hash implementation offered by the `carbon-c-relay relay <https://github.com/grobian/carbon-c-relay>`_ project.
```

webapp/graphite/local_settings.py.example
```diff
@@ -279,12 +279,32 @@
...
+# "fnv1a_ch" hashing from carbon-c-relay relay (https://github.com/grobian/carbon-c-relay)
+# Fnv1a_ch cluster uses instance name as the solely source of routing information.
+# As described in "carbon-c-relay" documentation: "The fnv1a_ch cluster is identical
+# in behavior to carbon_ch, but it uses a different hash technique (FNV1a) which
+# defined to get by a limitation of carbon_ch to use both host and port from the members.
+# When using the fnv1a_ch cluster, instance value overrides the hash key in use.
+# This allows for many things, including masquerading old IP addresses, but mostly
+# to make the hash key location to become agnostic of the (physical) location of that key.
+# For example, usage like 10.0.0.1:2003:4d79d13554fa1301476c1f9fe968b0ac would allow
+# to change port and/or IP address of the server that receives data for the instance key."
```
obfuscurity: Let's take a similar approach here to what I suggested above.

deniszh:
Done, @obfuscurity !

obfuscurity:
@deniszh Did you see my full comment? I had replacement text with formatting for that paragraph.

obfuscurity:
Hint: it was inside the blockquote. 😄

deniszh:
Oops, was blind, sorry :)
Looks better now?

docs/config-local-settings.rst
```diff
@@ -363,6 +363,12 @@ CARBONLINK_TIMEOUT
 
   Timeout for carbon-cache cache queries in seconds.
 
+CARBONLINK_HASHING_TYPE
+  `Default: carbon_ch`
+
+  `Possible values: carbon_ch, fnv1a_ch`
+
+  The default `carbon_ch` is Graphite's traditional consistent-hashing implementation. Alternatively, you can use the `fnv1a_ch`, which supports the Fowler–Noll–Vo hash function (FNV-1a) hash implementation offered by the `carbon-c-relay relay <https://github.com/grobian/carbon-c-relay>`_ project.
```
obfuscurity: Sorry, one more fix (my fault)... s/you can use the/you can use/. Also in the settings example below. Thanks!

deniszh:
NP, fixed!

obfuscurity:
Thanks!

deniszh:
Will squash and merge that EOD

Issue #361, opened 2013-06-25
index.json doesn't return all metrics from storage


mihaibivol:
When requesting metrics/index.json the [index_json](https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/metrics/views.py#L36) view doesn't follow symlinks when calling os.walk.

I saw #80 but this does not work for .wsp files

Is this behavior intended?

PR #1725, opened 2016-10-19
return realpath to fix issues with symlinks

edlitmus:
trivial fix for a symlinked file issue


obfuscurity:
Seems fine, only affects RRD.

obfuscurity:
@edlitmus Do you know if this fixes https://github.com/graphite-project/graphite-web/issues/361?

edlitmus:
It might: in the relevant rrdtool code it uses realpath(3) which should resolve symlinks, but we found that adding this helped us with an issue where RRDs were not getting flushed properly before graphite queried them.

PR #1726, opened 2016-10-20
Use latest minor Django version

elsmorian:
Use the most up to date security / bugfix minor version of Django available instead of only 1.9.0.

Not using `Django>=1.9,<1.10` syntax due to https://groups.google.com/forum/#!msg/python-virtualenv/DvG1InRGdR0/14b9CISO2AcJ

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,7 +35,7 @@
 #   deactivate
 #
 
-Django==1.9
+Django>=1.9,<1.9.99
 python-memcached==1.47
 txAMQP==0.4
 simplejson==2.1.6


deniszh:
@elsmorian ,
Are you sure that we're compatible with Django 1.9. IIRC we had some issues with that.

elsmorian:
@deniszh It was pinned to 1.9 before this change- looks like it was changed in commit fd53fad51e1b4807d1da1c2c4cd23e7726a10397

obfuscurity:
Good catch, I was thinking of the _release **series**_ instead of the specific version(s). I didn't mean to lock it into 1.9(.0).

PR #1713, opened 2016-10-02
Master brace expansion

deniszh:
Port of https://github.com/graphite-project/graphite-web/pull/1531 to master branch.

PR #1730, opened 2016-10-22
Optimizing brace expansion

deniszh:
According to @iain-buclaw-sociomantic comments in https://github.com/graphite-project/graphite-web/pull/1713/files `expand_braces()` is not really optimized.
I tested that using `timeit` - https://gist.github.com/deniszh/1e7eb64cb372d83d9ecd46d11c813193
According to tests, optimized version is 20% faster, so, it's worth trying.

Also change indentation back to 2 spaces.

### DIFF
--- a/webapp/graphite/finders/__init__.py
+++ b/webapp/graphite/finders/__init__.py
@@ -62,13 +62,12 @@ def expand_braces(s):
   m = EXPAND_BRACES_RE.search(s)
   if m is not None:
     sub = m.group(1)
-    open_brace = s.find(sub)
-    close_brace = open_brace + len(sub) - 1
-    if sub.find(',') != -1:
+    open_brace, close_brace = m.span(1)
+    if ',' in sub:
       for pat in sub.strip('{}').split(','):
-        res.extend(expand_braces(s[:open_brace] + pat + s[close_brace + 1:]))
+        res.extend(expand_braces(s[:open_brace] + pat + s[close_brace:]))
     else:
-        res.extend(expand_braces(s[:open_brace] + sub.replace('}', '\\}') + s[close_brace + 1:]))
+        res.extend(expand_braces(s[:open_brace] + sub.replace('}', '\\}') + s[close_brace:]))
   else:
       res.append(s.replace('\\}', '}'))
 
--- a/webapp/graphite/finders/__init__.py
+++ b/webapp/graphite/finders/__init__.py
@@ -43,33 +43,33 @@ def extract_variants(pattern):
 
 
 def match_entries(entries, pattern):
-    # First we check for pattern variants (ie. {foo,bar}baz = foobaz or barbaz)
-    matching = []
+  # First we check for pattern variants (ie. {foo,bar}baz = foobaz or barbaz)
+  matching = []
 
-    for variant in expand_braces(pattern):
-      matching.extend(fnmatch.filter(entries, variant))
+  for variant in expand_braces(pattern):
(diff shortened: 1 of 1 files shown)


webapp/graphite/finders/__init__.py
```diff
@@ -42,37 +43,33 @@ def extract_variants(pattern):
...
-          res.extend(expand_braces(s[:open_brace] + sub.replace('}', '\\}') + s[close_brace + 1:]))
+def expand_braces(s):
+  res = list()
+
+  m = EXPAND_BRACES_RE.search(s)
+  if m is not None:
+    sub = m.group(1)
+    open_brace = s.find(sub)
+    close_brace = open_brace + len(sub) - 1
+    if sub.find(',') != -1:
```
iain-buclaw-sociomantic: Alternatively, `if "," in sub:`. Whichever reads better I suppose. :-)

webapp/graphite/finders/__init__.py
```diff
@@ -42,37 +43,33 @@ def extract_variants(pattern):
...
-      else:
-          res.extend(expand_braces(s[:open_brace] + sub.replace('}', '\\}') + s[close_brace + 1:]))
+def expand_braces(s):
+  res = list()
+
+  m = EXPAND_BRACES_RE.search(s)
+  if m is not None:
+    sub = m.group(1)
+    open_brace = s.find(sub)
+    close_brace = open_brace + len(sub) - 1
```
deniszh: Yes, that's better

webapp/graphite/finders/__init__.py
deniszh: You mean, `if not "," in sub:`.
Not sure which one is better. :)

webapp/graphite/finders/__init__.py
deniszh: No, `open_brace, close_brace = m.span(1)` it's not the same. :(

webapp/graphite/finders/__init__.py
deniszh: Oh, I'm blind. Of course, `if "," in sub:`, and it's much better.

webapp/graphite/finders/__init__.py
iain-buclaw-sociomantic: Well, you have the following below:
```
      for pat in sub.strip('{}').split(','):
        res.extend(expand_braces(s[:open_brace] + pat + s[close_brace + 1:]))
    else:
        res.extend(expand_braces(s[:open_brace] + sub.replace('}', '\\}') + s[close_brace + 1:]))
```

With `m.span(1)` you can drop the ` + 1` bit.
```
      for pat in sub.strip('{}').split(','):
        res.extend(expand_braces(s[:open_brace] + pat + s[close_brace:]))
    else:
        res.extend(expand_braces(s[:open_brace] + sub.replace('}', '\\}') + s[close_brace:]))
```
:-)

deniszh: Ah, in that sense. Yes, sorry, didn't get your point from start. Yes, it works, and more readable now.

iain-buclaw-sociomantic:
Looks OK to me.

Following the linked bug report, I found the following revision change that this should probably be attributed to.
https://hg.python.org/cpython/rev/ec897bb38708

I was surprised to see how similar this now looks to it after my suggestions. :-)

deniszh:
Wow, indeed, @iain-buclaw-sociomantic !
Thanks for suggestions!

deniszh:
Merging for master and 0.9.x

PR #1733, opened 2016-10-24
Forward HTTP request headers to CLUSTER_HOSTS

benburry:
When making requests to remote webapps (defined in CLUSTER_HOSTS), allow us to
forward a configurable set of HTTP headers along from the original request.

The REMOTE_STORE_FORWARD_HEADERS setting, defaulting to an empty list, takes an
iterable of header names that are forwarded from the originating request along
to remote webapps, when a remote store request is made.

In our setup, this allows us to forward cookies and so preserve SSO auth status
between all graphite servers in a render cluster


obfuscurity:
Wouldn't it make more sense to just source the setting directly in `remote_storage.py` rather than passing it from the view (it's not a query param)?

benburry:
@obfuscurity I agree. I was struggling with how to access the request (and so, have access to the header values) anywhere outside of the view

Is there an easy django/graphite-web way to do this that I missed?

obfuscurity:
Sorry, I had a total brain fart. I forgot we were pulling the headers out of the request in the view. As you were. :wink:

obfuscurity:
Any feedback from anyone else @graphite-project/committers?

iksaif:
This could potentially help us add support for X-Forwarded-For, LGTM

benburry:
Thanks @obfuscurity. I'll get working on porting this to master

PR #1736, opened 2016-11-01
Forward HTTP request headers to CLUSTER_HOSTS

benburry:
When making requests to remote webapps (defined in CLUSTER_HOSTS), allow us to
forward a configurable set of HTTP headers along from the original request.

The REMOTE_STORE_FORWARD_HEADERS setting, defaulting to an empty list, takes an
iterable of header names that are forwarded from the originating request along
to remote webapps, when a remote store request is made.

In our setup, this allows us to forward cookies and so preserve SSO auth status
between all graphite servers in a render cluster

Port of #1733 to master


benburry:
I'm generally a little less comfortable with this on master because of the addition of the `local` param to STORE.find - having both `local` and `headers` parameters feels odd, as `local=True` will mean `headers` is unused.

Regardless, this is a bare port of #1733 to the master branch

iksaif:
My issue with this is that not all backends use HTTP behind the scene (https://github.com/criteo/biggraphite for example calls Cassandra instead), and changing the API to add headers is kind of weird.

Would there be a better / more generic way to pass the current request context ?

benburry:
I hear your concerns, and had overlooked that master now has custom storage backend support.

Without adjusting the API, making the request context available is achievable either by using a threadlocal (which has enough negative/conflicting opinions to make it an undesirable approach) or ... I'll have a think

obfuscurity:
/cc @DanCech who's working on related

Jason Dixon

DanCech:
In my clustering branch I'm passing requestContext down to the readers and using it to implement forward headers.  I haven't updated the finders to use that also but I do intend to do that.  Reading from the request context is thread safe, there isn't any locking needed unless you're writing to the same key from multiple threads.

webapp/graphite/remote_storage.py
```diff
@@ -124,12 +134,13 @@ def get_results(self):
 class ReadResult(object):
   __slots__ = ('lock', 'store', 'has_done_response_read', 'result', 'done_cb', 'connection', 'urlpath')
```
DanCech: 'headers' should be added here

DanCech:
Personally I don't see an issue with making the headers available, as they could well be used by custom storage backends.  @benburry if you want to add that entry to `__slots__` and rebase this on master I'd support merging it.

DanCech:
Solving the issue of `node.fetch` getting access to the headers so it can pass them to `RemoteReader` is somewhat complicated.  In #1818 we're passing `requestContext` (and `now`) through since `RemoteReader` needs access to `requestContext`, but since this branch doesn't have `now` that complicates things slightly.

I can cut another branch to make a minimal patch for handling `now`, which should then let us cleanly update this branch and get `requestContext` down to `RemoteReader` so it can grab the headers in a way that won't later conflict with #1818

DanCech:
Ok, this should be pretty much ready for merge. @obfuscurity @deniszh you guys want to take a look?

benburry:
Thanks for running with this @DanCech

deniszh:
Merging!

PR #1738, opened 2016-11-03
Enhance get_real_path to handle symlinks in any part of the path

cbowman0:
The conditional in get_real_path only worked when the file was a symlink.

This changes it so that if any part of the path is a symlink, then obtain the real metrics path, so that carbonlink lookups work.


deniszh:
LGTM

PR #1742, opened 2016-11-08
Don't fetch data from offline backends.

nyerup:
I experienced that having a CLUSTER_NODE offline rendered Graphite
completely unresponsive, and found out that this was due to an uncaught
exception in HTTPConnectionWithTimeout.connect().

In addition to this, render/datalib.py always iterated through all
CLUSTER_NODEs, without considering their 'available' property, i.e. if
they'd been marked as offline.

render/datalib.py:
  When fetching remote data, bypass remote stores that aren't available.

remote_storage.py:
  Also mark stores unavailable if requests to them fail on network or
  transport level.

### DIFF
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -48,9 +48,6 @@ class FindRequest:
     if self.cachedResults:
       return
 
-    self.connection = HTTPConnectionWithTimeout(self.store.host)
-    self.connection.timeout = settings.REMOTE_STORE_FIND_TIMEOUT
-
     query_params = [
       ('local', '1'),
       ('format', 'pickle'),
@@ -59,6 +56,9 @@ class FindRequest:
     query_string = urlencode(query_params)
 
     try:
+      self.connection = HTTPConnectionWithTimeout(self.store.host)
+      self.connection.timeout = settings.REMOTE_STORE_FIND_TIMEOUT
+
       if settings.REMOTE_STORE_USE_POST:
         self.connection.request('POST', '/metrics/find/', query_string, headers)
       else:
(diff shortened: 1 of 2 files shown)


webapp/graphite/remote_storage.py
```diff
@@ -139,12 +139,17 @@ def fetch(self, startTime, endTime, now=None, result_queue=None, headers=None):
...
-    else:
-      connection.request('GET', '/render/?' + query_string, None, headers)
+    try:
+      connection = HTTPConnectionWithTimeout(self.store.host)
+      connection.timeout = settings.REMOTE_STORE_FETCH_TIMEOUT
+      if settings.REMOTE_STORE_USE_POST:
+        connection.request('POST', '/render/', query_string, headers)
+      else:
+        connection.request('GET', '/render/?' + query_string, None, headers)
+    except:
```
bmhatfield: Can this `except` be more specific?

webapp/graphite/remote_storage.py
```diff
@@ -139,12 +139,17 @@ def fetch(self, startTime, endTime, now=None, result_queue=None, headers=None):
...
+      self.store.fail()
+      return None
```
bmhatfield: I see there's a similar pattern above with the `fail()` and `return None`. Is there some additional information you determined in debugging this taht would be helpful context here? Is `None` the right thing to return? Is there a richer/more correct type?

webapp/graphite/remote_storage.py
```diff
@@ -139,12 +139,17 @@ def fetch(self, startTime, endTime, now=None, result_queue=None, headers=None):
...
     query_string = urlencode(query_params)
 
-    connection = HTTPConnectionWithTimeout(self.store.host)
-    connection.timeout = settings.REMOTE_STORE_FETCH_TIMEOUT
-    if settings.REMOTE_STORE_USE_POST:
-      connection.request('POST', '/render/', query_string, headers)
-    else:
-      connection.request('GET', '/render/?' + query_string, None, headers)
+    try:
+      connection = HTTPConnectionWithTimeout(self.store.host)
```
bmhatfield: What was the actual exception with instantiating this class?

webapp/graphite/render/datalib.py
```diff
@@ -322,7 +322,7 @@ def prefetchLookup(requestContext, node):
 
 def fetchRemoteData(requestContext, pathExpr, usePrefetchCache=settings.REMOTE_PREFETCH_DATA):
   (startTime, endTime, now) = _timebounds(requestContext)
-  remote_nodes = [ RemoteNode(store, pathExpr, True) for store in STORE.remote_stores ]
+  remote_nodes = [ RemoteNode(store, pathExpr, True) for store in STORE.remote_stores if store.available ]
```
bmhatfield: What's the difference between here and `storage.py:41`?

nyerup:
All those questions make good sense. I'll get back with answers as soon as I can.

My problem is that I haven't found a way to reliably reproduce this problem in my test setup – only in my production setup, which has >50 cluster backends, and in which outages disturb my colleagues.

iksaif:
You might also be interested in https://github.com/criteo-forks/graphite-web/commit/479958f2826c286e15f6d9182466b72aec75b7fd that I haven't had the time to send here.

iksaif:
Is finally merged mine. Is this one still relevant ?

deniszh:
I think no. @nyerup ?

DanCech:
It's definitely outdated by #1810

deniszh:
Closing then. Sorry, @nyerup
Thanks for your contribution, anyway!

nyerup:
All right, makes good sense. Sorry for not getting around to clearing out these questions sooner.
I'll look through #1818, and open a new merge request if I find anything meaningful to contribute.

PR #1745, opened 2016-11-11
Add more test coverage for standard finder

cbowman0:
Fix the GzippedWhisperReader get_interval method to work
Add a little bit of coverate for finder of gzipped whisper files.

### DIFF
--- a/webapp/graphite/readers.py
+++ b/webapp/graphite/readers.py
@@ -11,6 +11,11 @@ try:
 except ImportError:
   whisper = False
 
+# The parser was repalcing __readHeader with the <class>__readHeader
+# which was not working.
+if bool(whisper):
+  whisper__readHeader = whisper.__readHeader
+
 try:
   import rrdtool
 except ImportError:
@@ -194,7 +199,7 @@ class GzippedWhisperReader(WhisperReader):
   def get_intervals(self):
     fh = gzip.GzipFile(self.fs_path, 'rb')
     try:
-      info = whisper.__readHeader(fh) # evil, but necessary.
+      info = whisper__readHeader(fh) # evil, but necessary.
     finally:
       fh.close()
 
--- a/webapp/tests/test_finders.py
+++ b/webapp/tests/test_finders.py
@@ -1,3 +1,4 @@
+import gzip
 import os
 from os.path import join, dirname, isdir
 import random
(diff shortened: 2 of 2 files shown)


obfuscurity:
Always a fan of your tests. 👍

webapp/graphite/readers.py
```diff
@@ -11,6 +11,11 @@
 except ImportError:
   whisper = False
 
+# The parser was repalcing __readHeader with the <class>__readHeader
+# which was not working.
+if bool(whisper):
```
SEJeff: Slightly nicer but equally effective:

    if whisper is False:

webapp/tests/test_finders.py
```diff
@@ -62,11 +63,15 @@ class StandardFinderTest(TestCase):
...
-    def create_whisper(self, path):
+    def create_whisper(self, path, gz=False):
         path = join(self.test_dir, path)
         if not isdir(dirname(path)):
             os.makedirs(dirname(path))
         whisper.create(path, [(1, 60)])
+        if gz:
+          with open(path, 'rb') as f_in, gzip.open("%s.gz" % path, 'wb') as f_out:
+             shutil.copyfileobj(f_in, f_out)
+          os.remove(path)
```
SEJeff: Nice! :+1:

webapp/tests/test_finders.py
```diff
@@ -153,6 +158,32 @@ def listdir_mock(d):
             os.listdir = self._original_listdir
             self.wipe_whisper()
 
+    def test_standard_finder_gzipped_whisper(self):
+        def listdir_mock(d):
+            self._listdir_counter += 1
+            return self._original_listdir(d)
+
```
SEJeff: This is entirely reasonable, but stupid question. Why not just use `mock.patch()` with a context manager?

cbowman0: This was how it was done in the other similar test.  Other than that, I don't know of a reason off hand to choose this way over `mock.patch()`.

I think we're pretty inconsistent with how we mock things in the tests.  It may be useful to define some standards ¯\\_(ツ)_/¯

webapp/graphite/readers.py
SEJeff: It tends to be bad python, but fair enough, that's an excellent reason.

SEJeff:
:boom:

webapp/tests/test_finders.py
SEJeff: More test coverage is better than perfectly consistent tests. Thanks for your help!

PR #1752, opened 2016-11-23
WIP update 0.10.0 release notes

obfuscurity:
Updating the 0.10.0 release notes. This is currently a WIP document. If there are any Carbon-focused devs I'd love your input here because I'm far less familiar with the changeset over there than I am with Graphite-Web. /cc @mleinart @drawks @iksaif @iain-buclaw-sociomantic


docs/releases/0_9_14.rst
```diff
@@ -172,15 +172,15 @@ Carbon
...
-* Fix instrumentation (avishai-ish-shalom, jssr)
+* Move tests to tox (jssjr)
+* Add hup signal handler (jssjr)
+* Fix instrumentation (avishai-ish-shalom, jssjr)
 * Fix exception handling (steve-dave)
-* Fix CACHE_WRITE_STRATEGY (jssr)
+* Fix CACHE_WRITE_STRATEGY (jssjr)
 * Fix aggregated metrics (pgul, ctavan)
 * Logging fixes (obfuscurity, piotr1212)
 * Fix race condition for full queues (mleinart)
```
iksaif: Protocols and clients are now "pluginized", which will allow custom ones to be writtens.

obfuscurity:
I've added some Carbon highlights. I'm actually torn on whether any/which of those "bug fixes" are worth mentioning since release notes _should_ be relative to the previous release (in this case, 0.9.15). Some of those are undoubtedly fixes in relation to _very old_ `0.9.x` releases, while some are probably just fixes in relation to a much earlier version of `master`.

OTOH it's entire possible I'm overthinking this and should just STFU and move on. :wink:

deniszh:
@obfuscurity : Maybe we should mention that master has no working carbon bulk query implementation. I can try to port that to master but TBH I see no real improvements if it's enabled...

obfuscurity:
@deniszh I don't think it makes enough difference to warrant mentioning here. I've added a lot more, everything look ok otherwise?

PR #1756, opened 2016-11-25
aliasByMetric() display a trailing ')'

mcoolive:
- fix the aliasByMetric impl
- add unit tests on aliasByMetric


obfuscurity:
Yup, makes sense. Thanks!

PR #1627, opened 2016-07-28
Event tags creation and serialization

obfuscurity:
This PR supersedes #636, merging conflicts and fixes #635.

/cc @brutasse @cbowman0

PR #1757, opened 2016-11-25
fixes event documentation

ctaintor:
I noticed there were a few problems with the event documentation, so I updated it to be more complete/correct.

### DIFF
--- a/docs/events.rst
+++ b/docs/events.rst
@@ -23,13 +23,17 @@ Events can be submitted via HTTP POST using command-line tools such as ``curl`` 
 .. code-block:: none
 
     $ curl -X POST "http://graphite/events/" 
-        -d '{ "what": "Event - deploy", "tags": ["deploy"], 
+        -d '{ "what": "Event - deploy", "tags": ["deploy"], "when": 1467844481,
         "data": "deploy of master branch happened at Wed Jul  6 22:34:41 UTC 2016" }'
+ 
+``when`` is an optional key which is set to the current Unix timestamp if ``when`` is not set.
+
+*Note*: Prior to 0.10.0, the value of ``tags`` is a string, with multiple tags being separated by a space.
 
 
 Querying Events
 ---------------
-Graphite allows you to query for tags associated with events. You can search for a single tag string, a combination of tags, or a simple ``*`` wildcard using the `events <functions.html#graphite.render.functions.events>`_ function.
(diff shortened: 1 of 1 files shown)


obfuscurity:
Your "fix" describes the old behavior that was changed in https://github.com/graphite-project/graphite-web/pull/1627. I assume you're using a 0.9 release.

The URL fix is correct, would you like to rework the patch for that only?

ctaintor:
Thanks for the info - I suspected as much but couldn't find the release notes (although I admit I didn't try that hard...).

I updated the PR

docs/events.rst
```diff
@@ -74,7 +78,9 @@ It's also possible to dump the raw events using the API.
        }
     ]
 
-The ``set`` parameter accepts an optional ``union`` or ``intersection`` argument to determine the behavior for filtering sets of tags (i.e. inclusive or exclusive). By default, Graphite uses a "lazy union" that will return any matching events for a given tag in a list of tags. This behavior is not intuitive and will therefore be deprecated in a future release.
+The ``set`` parameter accepts an optional ``union`` or ``intersection`` argument to determine the behavior for filtering sets of tags (i.e. inclusive or exclusive). By default, Graphite uses a "lazy union" that will return any matching events for a given tag in a list of tags. This behavior is not intuitive and will therefore be deprecated in a future release. When matching using multiple tags, separate the tags with a space.
+
+*Note*: Prior to 0.10.0, the value of ``tags`` will be a string, with multiple tags being separated by a space.
```
ctaintor: good point. In my case, I was talking about the output of get_data changing. I agree it's confusing so added what you suggested and clarified what I wrote.

obfuscurity: Ok, this still needs to be cleaned up. I really dislike that sentence being slapped onto the `set` paragraph, it doesn't belong there. And I don't think we need that *Note* since it's referring to the output. Nowhere else do we explicitly document the format of each key/value output. As long as we have an accurate example (which we do, [here](https://github.com/graphite-project/graphite-web/blame/master/docs/events.rst#L56-L75)), I don't see the point.

TL;DR please remove the changes in this snippet.

ctaintor: Sorry - I had meant to remove that sentence. I'll remove it. (the "When matching...")

About the output, I only have it there because it changed - in 0.9, the output doesn't match the example. It seemed worth noting. If you feel that has no value, I will remove it as well.

obfuscurity:
Thanks!

PR #1753, opened 2016-11-23
align django requirement specs & allow 1.9.x patch releases, remove u…

DanCech:
…nused tox envs

PR #1759, opened 2016-11-28
Revert "update django version specifier syntax"

DanCech:
This reverts commit be7fca6e436a42b0bc4ee0f5c005395c6deb901c.

The `~=` requirement syntax isn't supported by older pip releases including but not limited to 1.5.4, see #1753 for further discussion.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -35,7 +35,7 @@
 #   deactivate
 #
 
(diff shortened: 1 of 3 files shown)


SEJeff:
Can we get older pip for distros we support in the Travis config?

PR #1710, opened 2016-09-30
Render infinite values to valid JSON values. Fixes #813

rehevkor5:
- The 'json' and 'jsonp' formats output float('inf') as 1e9999, float
  ('-inf') as -1e9999, and float('nan') as null
- The 'dygraph' format (a JSON-based format) outputs the same as
  Infinity, -Infinity, and null (see http://dygraphs.com/tests/gviz-infinity.html)
- Minor updates to documents to make sure commands can be run without errors or problems

PR #1768, opened 2016-12-11
improve json rendering performance

DanCech:
One of the major differences in performance between 0.9.x and master is JSON rendering speed, which took a major hit because of the switch in master to use the custom FloatEncoder class.  In my testing I found it can be up to 9x slower than the default json encoder.

After tweaking around a fair bit with the JSON encoder, I came to the conclusion that there isn't a good way to get custom encoding for None, NaN and +/- Infinity values in python.  I even went as far as to write a completely custom encoder building the strings manually and couldn't get better than 2x slower than the default encoder.

So far this is the best solution I've found, which is to use the default encoder and then use string manipulation to swap out the values for None, NaN & Infinity.  It's not the most elegant solution in the world, but it is much much faster than any other option I can come up with.

For my test I rendered a result set that produces 26MB of JSON:
```
best result master:
Sat Dec 10 21:28:28 2016 :: Total json rendering time 6.015758
worst result master:
Sat Dec 10 21:24:56 2016 :: Total json rendering time 6.638874

best result json-encoding:
Sat Dec 10 21:24:32 2016 :: Total json rendering time 0.652258
worst result json-encoding:
Sat Dec 10 21:29:21 2016 :: Total json rendering time 0.731274

raw json.dumps:
Sat Dec 10 22:15:29 2016 :: Total json rendering time 0.652526
```

This branch tweaks the log output slightly, so that the json rendering timer tarts only when rendering starts, rather than also including finding and fetching the data.  It also adds a test to make sure than NaN values are properly encoded.

To facilitate comparison, the `FloatEncoder` class is still present, and can be used by changing `useFloatEncoder = False` to `True` on `webapp/graphite/render/views.py:177`


obfuscurity:
@DanCech Are you saying this is related to the issues reported in #1690? Because the `FloatEncoder` changes weren't merged into master until https://github.com/graphite-project/graphite-web/pull/1710, 12 days ago.

DanCech:
All I'm saying is that the FloatEncoder change made JSON rendering an order of magnitude slower than using `json.dumps` without the custom encoder.  For people with large data sets it's quite possible with FloatEncoder for the JSON rendering phase to take the majority of the time spent servicing a request.  For reference, the time spent retrieving the result set (find & fetch) in my examples above was less than 0.4s.

I didn't really start looking at that part of the pipeline until after the change was merged, but it does jive with what I was seeing in initial testing (before the merge) where it didn't seem to be a bottleneck.  The way the time taken was reported previously (including the fetch time) also made it difficult to understand exactly how much time was spent rendering the JSON because the duration logged included the find & fetch duration.

obfuscurity:
LGTM, any concerns or feedback from @rehevkor5 or @torkelo?

torkelo:
9x slower json encoding sounds like a major issue. Was this due to the fix of non standard JSON Infinity values? That fix is not worth a 9x json encoding perf hit

obfuscurity:
@torkelo Yeah, fixing a perf regression like that is a no-brainer. I cc'd you more for any concern about compatibility with Grafana. 😺

rehevkor5:
@obfuscurity My apologies for the unintended side effect. My only comment is about the `replace`. It's not syntax-aware, so obviously you'd want to be reasonably sure that there wouldn't be something like `{title: "To Infinity and Beyond"}` in the JSON, otherwise it's going to show up as `{title: "To 1e9999 and Beyond"}`. I'm sure you all know this component well enough to be sure whether that's an actual problem or not. I don't.

It's possible there's something bad about the way I extended JSONEncoder. For example, does overriding `default` instead of `iterencode` make any difference?

I'm also wondering whether using https://simplejson.readthedocs.io/en/latest/ might enable customization without such a big impact on performance? (it gets `None` right but still does +-inf and NaN the same way). Btw ujson is not an option if infinity or NaN is present, "OverflowError: Invalid Inf value when encoding double" and "Invalid Nan value when encoding double".

Perhaps a better option would be to fix the data itself before it gets to the JSON encoder. You'd have to figure out how to iterate over all the values (though maybe there are already places in the code where it's iterating over them). That approach would be unambiguous compared to string replacement, and would enable use of any naive json encoder (including ujson).

Regardless of what happens, I hope we will be able to retain this fix eventually. The JSON spec from ECMA explicitly states "Numeric values that cannot be represented as sequences of digits (such as Infinity and NaN) are not permitted." In that regard, both built-in Python json and simplejson are setting poor examples.

DanCech:
@rehevkor5 thanks for detailed comment!

I has the same concern as you about unintentional replacements, which is why I first tried:
 - using `default()` doesn't help us, since it is only used to encode custom objects.  We could maybe wrap every value into a custom object but I haven't tried that yet, since we'd be adding a lot of overhead to build up the list of "value" objects to then pass them into the encoder.
 - a raw string manipulation solution, I got it down to 2x the speed of raw `json.dumps` but I'm not sure how much better that that I could make it.

As for `simplejson`, the code in `FloatEncoder` actually looks a lot like https://github.com/simplejson/simplejson/blob/master/simplejson/encoder.py#L299 so I doubt that would help much.

Replacing the values before encoding won't work because then you'd end up with `"null"` in the output instead of `null`, We could possibly pre-process the values array and replace None, NaN & +/-Infinity values with placeholders like `###NULL###` or similar, then do string replacement on those strings, but I feel that it would add complexity and processing time for little benefit.

What I did include to help prevent the incorrect replacement scenario you described in that the replacement isn't `None` but `None,` etc.  That should make it less likely that there will be an incorrect replacement.  We could take that a step further by doing `.replace('[None, ', '[null, ')` etc (with or without the trailing space), though we would require an extra call to replace `-Infinity` since right now the single infinity replacement matches both positive and negative infinity.

As far as generating a large dataset, you can just take a target that matches a good number of series and use a long time range or include it multiple times.

I absolutely agree that the JSON output should be correct, so I hope we can arrive at a solution people can support.

rehevkor5:
Please take a look at https://github.com/rehevkor5/graphite-web/tree/fix_json_serialization_perf_regression

If we are willing to accept a slight change in the behavior of JSON serialization, we can fix the performance reversion while still getting reliable JSON. Specifically, instead of serializing inf and -inf as +-1e9999, I replace the values with Python's `sys.float_info.max` and `sys.float_info.min`. Similarly, I replace NaN with `None` in the data before it gets to the serializer. This allows the JSON serializer to serialize them normally.

If you think it's a good approach, I can do some minor cleanup & make a PR.

deniszh:
LGTM @DanCech @obfuscurity ?

obfuscurity:
Is the intent to remove the `FloatEncoder` class at some point in the near future? Why not remove it now?

DanCech:
The concern I'd have with this approach would be that it's creating a seemingly-random (and long; an when you have a lot of them) proxy value for the Inf/-Inf values, and as far as I can tell it'll still output `None` rather than `null` for None and NaN values.

I initially left the FloatEncoder class in for benchmarking, it and the associated import and if/else should be removed before any merge.

rehevkor5:
@obfuscurity yes I will remove it as cleanup before I submit a PR.

@DanCech It will output `null`, not `None`. I'm not sure if there's a reasonable fix for the values seeming random though. Any choice other than `sys.float_info.max` seems arbitrary.

rehevkor5:
Created https://github.com/graphite-project/graphite-web/pull/1785 for your consideration

obfuscurity:
@DanCech Assuming it fixes the performance hit, I'm leaning towards #1785 since it fully reverts the FloatEncoder class. Any concerns?

DanCech:
Personally I'm not in favor of the `1.7976931348623157e+308` and `2.2250738585072014e-308` values for `Inf` and `-Inf`, which have the disadvantage that `json.loads('[2.2250738585072014e-308]')` returns `[2.2250738585072014e-308]` rather than `[-inf]` (as `json.loads('[-1e9999]')` does) as well as being ~3x as many bytes.  The actual values used are also dependent on `sys.float_info.max / sys.float_info.min` which are platform/implementation dependent and for cPython are based on the values for the c constants `DBL_MAX / DBL_MIN`.

I know that doing string manipulation on the JSON is also not a perfect solution, but I don't see a better one.  It lets us use a compact and platform-independent representation for these values, which will get correctly decoded by python.

I can update this PR to totally remove FloatEncoder very easily, it's only in there for comparison purposes.

obfuscurity:
@DanCech Makes sense to me, thank you for the detailed explanation. I'm :+1: in favor of your approach. Anyone else want to chime in before I merge this?

PR #1769, opened 2016-12-12
webapp/glyph.py: fix blank space below the legend

Marek77:
Don't add padding (5px) for each line of legend; only add it
        once to separate the graph area from the legend.

Before and after fix graphs:
1 line: http://tinypic.com/r/muabys/9
5 lines: http://tinypic.com/r/2wmniiv/9
6 lines: http://tinypic.com/r/23h7lgw/9
5 lines, with the fix: http://tinypic.com/r/2db2i6f/9
6 lines, with the fix: http://tinypic.com/r/2s8pvzn/9

### DIFF
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -727,7 +727,7 @@ class Graph:
       numberOfLines = max(len(elements) - numRight, numRight)
       columns = math.floor(columns / 2.0)
       if columns < 1: columns = 1
-      legendHeight = max(1, (numberOfLines / columns)) * (lineHeight + padding)
+      legendHeight = max(1, (numberOfLines / columns)) * lineHeight + padding
       self.area['ymax'] -= legendHeight #scoot the drawing area up to fit the legend
       self.ctx.set_line_width(1.0)
       x = self.area['xmin']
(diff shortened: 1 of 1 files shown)


Marek77:
I can't say I know what is that failed check about..?

deniszh:
It's codecov patch status - https://github.com/codecov/support/wiki/Commit-Coverage-Status
In your case you can ignore that.

obfuscurity:
@Marek77 The only image link that works for me is the first one. Instead of linking externally, please embed the images in your original comment using GitHub attachments (see the help at the bottom of the comment editor window).

Marek77:
You will probably need to download the images to see the bottom border getting bigger in the first three images, and their fixed status (no extra border) in the last three. The issue becomes really visible with lots of metrics (I saw the bug with 'hideLegend=false').

Border:





Patch applied:

2_fixed
5_fixed
6_fixed

obfuscurity:
Ok, I agree with the intent of the change. Would you put the parentheses back (in the right space) to remove any ambiguity in the math?

Marek77:
OK, noise added ;-)

obfuscurity:
Yeah, I know. I just really like the added clarity cuz I is stupid. :wink:

obfuscurity:
Thanks @Marek77!

PR #1771, opened 2016-12-13
Update the command to setup a new database.

petecheslock:
per this doc where its set as such.

https://github.com/graphite-project/graphite-web/blob/master/docs/config-database-setup.rst


obfuscurity:
Thanks! 🍰

PR #1773, opened 2016-12-14
Updated doc custom install location

deltaskelta:
With the previous custom install with `pip install carbon` or `pip install graphite-web` a django 1.9 incompatible version is installed


obfuscurity:
Agreed, we need to be consistent within the same page. Thanks!

PR #1774, opened 2016-12-14
Improved stability/performance for storage

iksaif:
- readers: be more resilient to the loss of a single backend
- storage: try to shuffle nodes
- finder: avoid calling get_intervals() when listing metrics

### DIFF
--- a/webapp/graphite/node.py
+++ b/webapp/graphite/node.py
@@ -23,11 +23,14 @@ class LeafNode(Node):
   def __init__(self, path, reader):
     Node.__init__(self, path)
     self.reader = reader
-    self.intervals = reader.get_intervals()
     self.is_leaf = True
 
   def fetch(self, startTime, endTime):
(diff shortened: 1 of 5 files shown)


obfuscurity:
Generally LGTM. What's the advantage to using `random.shuffle()`?

/cc @DanCech

iksaif:
The idea of the random.shuffle() is that it changes the first node that get contacted, this help spreading the load, on our setup we found out that one node would timeout way more than the others, and it was always the first node in the list.

DanCech:
code changes look reasonable to me, need to figure out the failing test though.

iksaif:
looks like it calls `listdir` in ceres less now because it doesn't try to get the intervals even when you don't need them, which is good :)

iksaif:
@obfuscurity test are fixed, do I get a final LGTM ? :)

obfuscurity:
👍

Issue #1779, opened 2016-12-24
Allow deduplicating by path


TvdW:
One can't always plot a hundred lines on a graph, as this would make the graph incomprehensible. Aggregations help, but sometimes you're really interested in the actual metrics. In this case, filters like lowestCurrent can rock, but even then you sometimes want to show a bit more data.

One of my most useful graphs plots a combination of `mostDeviant(5,server.*.disk_space)` and `lowestCurrent(server.*.disk_space,5)`. However, occasionally some metrics will be matched twice, and Graphite's render code will happily draw the line twice. The legend can be told to deduplicate, with uniqueLegend=true, but because all lines are still drawn, the color of the line may not actually match what is shown in the legend.

A simple uniq() function could mitigate this, by taking two or more seriesLists and returning only unique paths. In my case this would look like `uniq(mostDeviant(5,server.*.disk_space),lowestcurrent(server.*.disk_space))`.

PR #1780, opened 2016-12-25
Implement uniqueNames() for issue #1779

TvdW:
I guess the code for issue 1779 would look something like this. :)

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -3665,6 +3665,28 @@ def removeEmptySeries(requestContext, seriesList):
     """
     return [ series for series in seriesList if safeIsNotEmpty(series) ]
 
+def uniqueNames(requestContext, *seriesLists):
+  """
+  Takes an arbitrary number of seriesLists and returns unique series, filtered by name.
+
+  Example:
+
+  .. code-block:: none
+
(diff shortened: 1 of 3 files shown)


webapp/graphite/render/functions.py
```diff
@@ -3665,6 +3665,28 @@ def removeEmptySeries(requestContext, seriesList):
...
+
+  .. code-block:: none
+
+    &target=uniqueNames(mostDeviant(server.*.disk_free,5),lowestCurrent(server.*.disk_free,5))
+
+  Draws servers with low disk space, and servers with highly deviant disk space, but never the same series twice.
+
+  """
+  newList = []
+  seenNames = {}
```
yati-sagade: Can be a set (`seenNames = set()`) and then `seenNames.add(series.name)` :)

obfuscurity:
I like this in general, but I would prefer to call it `uniqueSeries` or simply `unique`.

Should the function attempt to do any reconciliation of series with different contents (or determine preference)?

obfuscurity:
refs #1779

deniszh:
Hello @yati-sagade,
Would you mind to rename your function to `unique`? It will have nice match with `grep`, `exclude` etc.

DanCech:
Closing in favor of #2063

PR #1784, opened 2017-01-03
webapp/glyph.py: fix incorrect display of 'title' URI parameter

Marek77:
When passing an URI-encoded string (e.g. '&title=Hello%20World%21'), display 'Hello World!' instead in graph title.

Graph before the fix:
test_uri_bad

Fixed version:
test_uri_good

### DIFF
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -1046,11 +1046,11 @@ class LineGraph(Graph):
     self.setColor( self.foregroundColor )
 
     if params.get('title'):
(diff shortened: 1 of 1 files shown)


Marek77:
As the fix changes the generated image, I admit I have hard time seeing what to test and how.

iksaif:
I guess you can mock  drawVTitle to validate that its being called with the correct argument

deniszh:
Not sure if a test is needed here. unquote_plus() in known function and fix is quite minor.

deniszh:
Thanks! Merging

PR #1785, opened 2017-01-03
Fix JSON performance regression caused by #1710

rehevkor5:
- Removed slow FloatEncoder, using json.dumps again instead
- Still getting good behavior with regard to inf, -inf, and NaN
by converting the values in the data to max & min float value and
None respectively. allow_nan=False prevents bad values from
accidentally being rendered in the JSON.


iksaif:
How much faster is the new version compared to the old one ?

webapp/graphite/render/views.py
```diff
@@ -155,37 +168,41 @@ def renderView(request):
...
                                 content_type='application/json')
 
       if useCache:
         cache.add(requestKey, response, cacheTimeout)
         patch_response_headers(response, cache_timeout=cacheTimeout)
       else:
         add_never_cache_headers(response)
-      log.rendering('Total json rendering time %.6f' % (time() - start))
+      log.rendering('JSON rendering time %.6f' % (time() - jsonStart))
+      log.rendering('Total time %.6f' % (time() - start))
```
obfuscurity: Would this be more clear as `Total rendering time`?

obfuscurity:
Closing this one in favor of #1768. Regardless, @rehevkor5 thank you for the original change and for your input on fixing the regression.

Issue #988, opened 2014-10-13
Make render functions pluggable/modularized


drawks:
Render functions should be refactored to allow for plugging in additional "site-local" functions. This should be implemented in such a way that user can add entire functions or groups of functions that would extend the current capabilities of graphite-web OR override the existing functions with different versions which may provide better performance or slightly different implementation details which are more suitable to a user's needs.

We should for example support a user dropping a file into a directory which defines a group of functions which implement some seasonal statistical analysis which would otherwise be unsuitable for general inclusion.  Or a user should be able to drop in functions which reimplement some of the current core rendering functions but use numpy for optimized performance.

DanCech:
This was implemented in #2146

PR #1789, opened 2017-01-13
add consecutiveNonZero and consecutiveSecondsNonZero functions

Roguelazer:
This pair of functions returns the number of buckets or seconds (depending on the function called) where at least one of the invocant series has had a non-zero, non-null value. This kind of metric is particularly good for looking at things like the number of seconds that a system has been queueing traffic.

### DIFF
--- a/webapp/content/js/composer_widgets.js
+++ b/webapp/content/js/composer_widgets.js
@@ -1184,6 +1184,8 @@ function createFunctionsMenu() {
         {text: 'Changed', handler: applyFuncToEach('changed')},
         {text: 'Transform Nulls', handler: applyFuncToEachWithInput('transformNull', 'Please enter the value to transform null values to')},
         {text: 'Count non-nulls', handler: applyFuncToAll('isNonNull')},
+        {text: 'Consecutive non-zeros', handler: applyFuncToEach('consecutiveNonZero')},
+        {text: 'Consecutive seconds non-zeros', handler: applyFuncToEach('consecutiveSecondsNonZero')},
         {text: 'Substring', handler: applyFuncToEachWithInput('substr', 'Enter a starting position')},
         {text: 'Group', handler: applyFuncToAll('group')},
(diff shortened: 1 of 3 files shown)


obfuscurity:
I'm probably being naive, but why not just convert nulls to zero, sum up all the series, and look for anything > 0 (or the actual value, if that's interesting)?

deniszh:
@Roguelazer - I'm with @obfuscurity here. Do you have some usecase where his proposal will not work?
I'm generally 👍 for adding more functions to Graphite but let's not bloat it over the top.

Roguelazer:
The particular metric I'm concerned with this is "time since queue X was empty". I guess I could divide the series by itself to normalize all non-zeros to one and then integrate that? Presuming 0/0 equals 0, which I don't think is strictly speaking true.



-- 
James Brown,
currently mobile

deniszh:
Hm. Looks legit. What do you think, @obfuscurity ?

Roguelazer:
Somewhat unrelatedlay, it would be neat if there were a way to add functions to graphite without patching the package; if you could, say, list other python packages in a config file and they'd be imported and searched via some standard protocol for functions.

Then I'd probably not buy y'all with things like this.

-- 
James Brown,
currently mobile

obfuscurity:
@Roguelazer Yes, we've talked about adding the concept of "pluggable functions" to Graphite. We'd all agree that it would be a fantastic addition, someone just has to find the time and resources to make it happen.

obfuscurity:
@deniszh I don't really have a strong opinion on it anymore.

deniszh:
@Roguelazer - yes, you're completely right, it should be implemented. We already have issue #988 for that, which is added to v1.1.0 milestone. Maybe we can try to add it in same way as [`graphite-api` does](https://graphite-api.readthedocs.io/en/latest/custom-functions.html)

Could you please rebase test file? IMO your functions can be useful for monitoring.
Thanks a lot!

PR #1792, opened 2017-01-17
Exponential Moving Average Function Implementation

cmaurer:


deniszh:
Cool! LGTM!
Anyone else wants to check? /cc @graphite-project/committers

obfuscurity:
I'm curious about the new `gitignore` entries. I like the added tests. Everything else looks generally good although I haven't scrutinized it yet.

cmaurer:
I added those entries while testing.  I can remove them and update the pull request.

iksaif:
As the function isn't trivial, the commit log and doctstring could eventually contain a rationale for the use of such function and/or some links.

cmaurer:
@iksaif I am not sure I understood your comment.  Is it related to .gitignore, or the EMA function itself?

iksaif:
The function itself :)

deniszh:
Not sure if explanation of the function is needed. Maybe link to wikipedia? https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average

deniszh:
Could you please rebase, @cmaurer ? I'm gonne merge it if no objections. :)

cmaurer:
Yes I can.  I will do it as soon as I can.

deniszh:
Did you finish, @cmaurer ? Can we review it already?

setup.py
```diff
@@ -94,7 +94,7 @@
         ['templates/*', 'local_settings.py.example']},
       scripts=glob('bin/*'),
       data_files=webapp_content.items() + storage_dirs + conf_files + examples,
-      install_requires=['Django>=1.9,<1.9.99', 'django-tagging==0.4.3', 'pytz', 'pyparsing', 'cairocffi'],
+      install_requires=['Django>=1.9,<1.9.99', 'django-tagging==0.4.3', 'pytz', 'pyparsing<2.1', 'cairocffi'],
```
deniszh: We already removed pyparsing version binding, could you please remove it too?

cmaurer: yep.  Sorry about that.  Doing it now.

cmaurer: @deniszh removed pyparsing version binding.

deniszh: Thanks!

deniszh:
I'll merge after another review.

webapp/graphite/render/functions.py
```diff
@@ -756,6 +756,88 @@ def weightedAverage(requestContext, seriesListAvg, seriesListWeight, *nodes):
...
+    if isinstance(windowSize, basestring):
+      newName = 'exponentialMovingAverage(%s,"%s")' % (series.name, windowSize)
+    else:
+      newName = "exponentialMovingAverage(%s,%s)" % (series.name, windowSize)
+
+    newSeries = TimeSeries(newName, series.start + previewSeconds, series.end, series.step, [])
+    newSeries.pathExpression = newName
+    window_sum = safeSum(series[:windowPoints]) or 0
+    count = safeLen(series[:windowPoints])
+    EMA = safeDiv(window_sum, count)
```
iksaif: uppercase variables are usually for constants. Simply `ema` here would be fine.

webapp/tests/test_functions.py
```diff
@@ -3864,3 +3862,138 @@ def test_summarize_1minute_alignToFrom(self):
               },
               seriesList, "1minute", func, True)
           self.assertEqual(result, expectedResults[func])
+
+    def test_exponentialMovingAverage_integerWindowSize(self):
```
iksaif: There is quite a bit of duplicated code in these tests ? Could you try refactoring that ?

Having a helper function called in the various tests and putting some of the values (like the dates) into global constants might help

iksaif: There is quite a bit of duplicated code in these tests ? Could you try refactoring that ?

Having a helper function called in the various tests and putting some of the values (like the dates) into global constants might help

webapp/graphite/render/functions.py
cmaurer: Cool.  I will change it.

webapp/tests/test_functions.py
cmaurer: Cool. Let me take a look at that this afternoon.

cmaurer: @iksaif I did some work on your second request yesterday, I would like to get your feedback to make sure it is in-line with what you are requesting before I refactor everything, and commit. 

Thanks,
Chris

- extracted gen_seriesList to global helper function
- created global function to create request contexts

```python
    def test_exponentialMovingAverage_evaluateTokens_returns_half_none(self):
        seriesList = self._gen_series_list_with_data(start=10)
        expectedResults = self._gen_series_list_with_data(
            key='exponentialMovingAverage(collectd.test-db0.load.value,10)',
            start=20,
            end=30,
            data=[0, 0.0, 0.182, 0.512, 0.965, 1.517, 2.15, 2.85, 3.604, 4.404, 5.239]
        )

        def mock_evaluateTokens(reqCtx, tokens, replacements=None):
            return self._gen_series_list_with_data(key='collectd.test-db0.load.value',start=10, end=30, data=([None] * 10 + range(0, 10)))

        with patch('graphite.render.functions.evaluateTokens', mock_evaluateTokens):
            result = functions.exponentialMovingAverage(self._build_request_context(endTime=datetime(1970, 1, 1, 0, 9, 0, 0, pytz.timezone(settings.TIME_ZONE))), seriesList, 10)

        self.assertEqual(result, expectedResults)

    def _build_request_context(self, startTime=datetime(1970, 1, 1, 0, 0, 0, 0, pytz.timezone(settings.TIME_ZONE)), endTime=datetime(1970, 1, 1, 0, 59, 0, 0, pytz.timezone(settings.TIME_ZONE)),data=[]):
        return {
            'template': {},
            'args': ({}, {}),
            'startTime': startTime,
            'endTime': endTime,
            'localOnly': False,
            'data': data
        }

    def _gen_series_list_with_data(self, key='collectd.test-db0.load.value', start=0, end=59, step=1, data=[]):
        seriesList = [
            TimeSeries(key, start, end, step, data)
        ]
        for series in seriesList:
            series.pathExpression = series.name

        return seriesList
```

iksaif: Looks good to me :)

cmaurer: cool.  I will refactor and commit.  It might take me until tonight (GMT-5), as I am at work now, etc.
Thanks,
Chris

cmaurer: @iksaif ok, checked in.  Please let me know what you think.

Thanks,
Chris

deniszh:
Cool! Thanks for reformatting test functions, @cmaurer !
Merging.

deniszh:
Reverting, tests are become real flacky now.
```
FAIL: test_find_view (tests.test_metrics.MetricsTester)
  File "/home/travis/build/graphite-project/graphite-web/webapp/tests/test_metrics.py", line 333, in test_find_view
    self.assertEqual(int(data[0]['intervals'][0]['start']), ts_minus_sixty_seconds)
AssertionError: 1487678717 != 1487678716
```

deniszh:
Tried 3 times, failed 3 of 3 attempts. :(

deniszh:
Reverted. Not sure why it's doing that. Do you have ideas, @cmaurer ?
Or @iksaif ?

DanCech:
@deniszh I suspect it's to do with the way that the whisper library use time.time() when determining the covered intervals.  We never saw it before because there was no test coverage, but itooks like those test are affected by how long it is between the whisper files being written and when they're queried.  For now I'd recommend commenting out those lines in the test that assert on 'intervals' values, and I'll try to figure out what a solution might be.  The reliance on time.time() in the whisper finder seems like a bug to me. In whisper itself time.time() is present but seems to only be used as a fallback if a "proper" value isn't provided.

cmaurer:
@deniszh, @DanCech I commented out the lines checking ts - 60, and committed.  I don't know if the changes will be picked up after a merge or not.

deniszh:
I think you should create new PR, @cmaurer :(
I see latest commit 15 hours ago. Sorry for that.

cmaurer:
No worries.  I can do that.

Chris

PR #1793, opened 2017-01-20
support whisper aggregation method "last" in readers.py

Roguelazer:
This method gets called with the rollup aggregation method specified in the whisper file; Whisper supports the "last" aggregation method (as per <http://graphite.readthedocs.io/en/latest/whisper.html#rollup-aggregation>). This patch fixes reading whisper files with their aggregationMethod set to "last".


deniszh:
LGTM
👍

obfuscurity:
Thanks!

PR #1797, opened 2017-01-25
requirements: unpin pyparsing

iksaif:
Looks like it works well with the latest version (and the old
one is problematic now)


deniszh:
LGTM,  merging

PR #1804, opened 2017-01-30
Allow to override Memcache options easily

mcoolive:


deniszh:
No examples in config? :(

obfuscurity:
Agreed, we should have at least a commented-out entry in `local_settings.py` so users know it exists.

mcoolive:
You're right. I am going to update local_settings.py with an example and some explanation.

iksaif:
LGTM

PR #1808, opened 2017-02-01
Introduce a movingSum() function

### LINKED ISSUE #1036: movingSum
I have a metric counting events per minute. It would be nice to see the trailing sum of those events over the last day. This is different than `summarize` because the `summarize` function collapses 1440 minutes of data points (a day) into a single data point and results in graphs that are very discontinuous.

I would like to have a `movingSum` function which is similar to the `movingAverage` and `movingMedian` functions.

Bonus points for a general-purpose `movingWindow` function which takes one of a set of supported aggregation functions. `sum`, `avg`, `max`, `min`, `max`, and `median` all come to mind.

benbc:
Fixes #1036


deniszh:
Maybe could be simplified by merging movingAvegare() and movingSum()... myeah, good enough for me. 
Thanks for the tests!
Any other of @graphite-project/committers wants to check?

obfuscurity:
:+1:

deniszh:
Thanks for your contribution, @benbc !

PR #1810, opened 2017-02-01
Remove pyparsing pinning requirement

ybizeul:
By using a Regex parser instead of building the rule we are able to get rid of the concurrency issue while loading multiple graphs.

From Paul McGuire, Pyparsing author :

> That KeyError is being raised inside the collections.OrderedDict code, and the key is extracted just a couple of lines prior to the place where the exception is raised, so we know it is a valid key. But there is a (small) window of time in which the same key could be popped by two different threads, and so the second thread will fail as the key will no longer be there.

> I've found in other parsers that building up numeric values with Combine, Word, etc. is a real slowdown compared to using a Regex, and there is not that much loss in readability or maintainability. These re's aren't that long or complicated.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -94,7 +94,7 @@ try:
         ['templates/*', 'local_settings.py.example']},
       scripts=glob('bin/*'),
       data_files=webapp_content.items() + storage_dirs + conf_files + examples,
-      install_requires=['Django>=1.9,<1.9.99', 'django-tagging==0.4.3', 'pytz', 'pyparsing<2.1', 'cairocffi'],
+      install_requires=['Django>=1.9,<1.9.99', 'django-tagging==0.4.3', 'pytz', 'pyparsing', 'cairocffi'],
       classifiers=[
           'Intended Audience :: Developers',
           'Natural Language :: English',
--- a/webapp/graphite/render/grammar.py
+++ b/webapp/graphite/render/grammar.py
@@ -1,23 +1,19 @@
 from pyparsing import (
     ParserElement, Forward, Combine, Optional, Word, Literal, CaselessKeyword,
     CaselessLiteral, Group, FollowedBy, LineEnd, OneOrMore, ZeroOrMore,
-    nums, alphas, alphanums, printables, delimitedList, quotedString,
+    nums, alphas, alphanums, printables, delimitedList, quotedString, Regex,
     __version__,
 )
 
-ParserElement.enablePackrat()
 grammar = Forward()
 
 expression = Forward()
 
 # Literals
-intNumber = Combine(
-  Optional('-') + Word(nums)
-)('integer')
 
-floatNumber = Combine(
-  Optional('-') + Word(nums) + Literal('.') + Word(nums)
-)('float')
+intNumber = Regex(r'-?\d+')('integer')
+
+floatNumber = Regex(r'-?\d+\.\d+')('float')
 
 sciNumber = Combine(
   (floatNumber | intNumber) + CaselessLiteral('e') + intNumber
--- a/webapp/graphite/render/grammar.py
+++ b/webapp/graphite/render/grammar.py
@@ -1,7 +1,7 @@
 from pyparsing import (
-    ParserElement, Forward, Combine, Optional, Word, Literal, CaselessKeyword,
+    Forward, Combine, Optional, Word, Literal, CaselessKeyword,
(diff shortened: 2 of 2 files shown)


deniszh:
Cool! Just remove unused imports, it breaks tests.

obfuscurity:
I haven't tested but I like the approach.

ybizeul:
There you go!

obfuscurity:
Would be nice to have someone benchmark reads after this change.

iksaif:
Would be easy to write a benchmark for grammer, but I strongly suspect that grammar parsing is blazing fast compared to the I/O made when rendering the target.

DanCech:
This looks good, but I'm not sure what the implications are of disabling packrat

ybizeul:
Please see #1828, this becomes a real issue as pinning doesn't work in some circumstances.

ybizeul:
@DanCech That I don't know. What I have is just words from pyparsing author that performance shouldn't be impacted. Is that your question?

DanCech:
@ybizeul do you mean on chat or something you can link us to?

ybizeul:
@DanCech Yes in my original post, last quote

DanCech:
@ybizeul doesn't appear to mention packrat?

ybizeul:
Look, I'm trying to pretend I know what we're talking about here. Obviously you got me :-D

I assumed "packrat" was the other chain parsing method, which I dropped in favor of regular expression. That's not what it is right?

DanCech:
After reading https://pyparsing-public.wikispaces.com/Frequently%20Asked%20Questions-What%20the%20heck%20is%20%22packrat%20parsing%22? I think that as long as performance isn't affected adversely we shouldn't see any other issues from not using packrat.

iksaif:
As said earlier, I don't think we should worry about performances here, most of the time is spent doing I/O anyway.

deniszh:
Agreed - parsing metrics' grammar is not most heavyweight operation on graphite. But I agreed with @obfuscurity - we should benchmark that, at least on separate test.

iksaif:
@ybizeul would you have time to write a small benchmark for that ?

ybizeul:
@iksaif Not sure how to do that?

iksaif:
@ybizeul I guess simply try to call grammar.parseString(target) ~10000 times with and without the patch and record the time :)

ybizeul:
Ok, both runs with 1000 iterations.

Legacy :
```
netapp.perf.MBP.clus1.node.clus1-01.vol_summary.total_data 2s
netapp.perf.MBP.clus1.node.clus1-01.aggr.disk_busy 3s
netapp.perf.MBP.clus1.node.clus1-01.vol_summary.read_latency 4s
netapp.perf.MBP.clus1.node.clus1-01.vol_summary.write_latency 4s
netapp.perf.MBP.clus1.node.clus1-01.processor.avg_processor_busy 6s
sumSeries(netapp.perf.MBP.clus1.node.clus1-01.vol_summary.total_ops) 7s
```

Patched:
```
netapp.perf.MBP.clus1.node.clus1-01.vol_summary.read_latency 1s
netapp.perf.MBP.clus1.node.clus1-01.vol_summary.total_data 1s
netapp.perf.MBP.clus1.node.clus1-01.vol_summary.write_latency 2s
netapp.perf.MBP.clus1.node.clus1-01.processor.avg_processor_busy 2s
sumSeries(netapp.perf.MBP.clus1.node.clus1-01.vol_summary.total_ops) 2s
netapp.perf.MBP.clus1.node.clus1-01.aggr.disk_busy 1s

**Comment by DanCech** [comment]
Very nice! Can you share the benchmarking script?

**Comment by ybizeul** [comment]
I didn't know how to isolate it, so I don't really have a script.

What I did is this in `graphite/render/evaluator.py` (and I reinstalled since then but you'll get the idea) :

```
def evaluateTarget(requestContext, target):

  sec_ = calendar.timegm(time.gmtime())

  for x in xrange(1, 1000):                                  # Loop
    tokens = grammar.parseString(target)
  result = evaluateTokens(requestContext, tokens)

  log target + " " + calendar.timegm(time.gmtime()) - sec_ # pseudo code
  if type(result) is TimeSeries:
    return [result] #we have to return a list of TimeSeries objects

  else:
    return result
```

Then just loaded a dashboard page

deniszh:
So, is that enough to merge? @graphite-project/committers ?

DanCech:
I think so :+1:

PR #1813, opened 2017-02-02
Make movingSum available for use

benbc:
My PR from yesterday (#1808) doesn't work, I'm afraid. The missing piece is here.


deniszh:
Yes, looks like it was overlooked. Thanks!

obfuscurity:
Dangit, I was looking specifically for that and missed it. Thanks for catching it in hindsight.

PR #1815, opened 2017-02-07
support 0.9.x backends

DanCech:
The structure of the pickled response from the find endpoints changed from 0.9.x to master, this patch allows a front-end instance running master to use back-end nodes running 0.9.x by accepting both result formats.

### DIFF
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -9,6 +9,7 @@ from graphite.readers import FetchInProgress
 from graphite.logger import log
 from graphite.util import unpickle
 from graphite.render.hashing import compactHash
+from graphite.intervals import Interval, IntervalSet
 
 def connector_class_selector(https_support=False):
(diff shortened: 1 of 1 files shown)


obfuscurity:
Haven't tested but LGTM. Anyone have time to test against both backend versions?

obfuscurity:
P.S. Great idea @DanCech. :smile_cat:

deniszh:
Absolutely great idea, indeed!
Same here, did not tested, but looks good.
Did you tried that code, @DanCech ?

DanCech:
@obfuscurity @deniszh yes, I've been using this code as part of the cluster performance testing so I can run both 0.9.x and master front-end nodes against the same set of 0.9.x back-end nodes to eliminate back-end node performance as a source of variation.

DanCech:
The alternative approach would be to change the master find output to match the 0.9.x format, since I'm not sure that there's any real benefit to the change in format.

deniszh:
I think that approach is fine. Merging.

PR #1819, opened 2017-02-10
default whisper reader - number of os.stat >> stat syscalls

olevchyk:
super small fix reduces the number of stat syscalls by using scandir libray:
https://github.com/benhoyt/scandir
scandir has been included in the Python 3.5 standard library as os.scandir()

context:
by tracing graphite-web workers on a system with slow disk sub-system, found out that end ts of whisper database query we get by calling os.stat for database, so even if get your data from local carbon-cache instance, you generate abnormal amount of stats calls.
In our case graphite workers spend 90% of time on stat calls on the storage subsystem with limited number of IOPS.
After fix, we spend only 25% of time

https://www.python.org/dev/peps/pep-0471/
This new function adds useful functionality and increases the speed of os.walk() by 2-20 times (depending on the platform and file system) by avoiding calls to os.stat() in most cases.

changed function calls only in most expensive places, without trying to change _find_paths() not be recursive


deniszh:
Nice improvement, thanks a lot! Let me (and others) review it?

olevchyk:
that would be awesome.  I already see that imports are not optimal and probably I would need to change the tests to use Mock >> call_count

deniszh:
In general, I'm very 👍 about this patch. Graphite was always quite hungry about disk operations and any optimizations in this area are quite valuable. OTOH this patch makes `scandir` module as mandatory dependence, which is probably fine because it's a master branch.
It's quite easy to make this patch to use `os` module if scandir is not present - but I'm not sure if that needed.
What other committers say? @obfuscurity @iksaif @DanCech 
/cc @graphite-project/committers

obfuscurity:
:+1:

webapp/graphite/finders/standard.py
```diff
@@ -3,6 +3,14 @@
 from os.path import isdir, isfile, join, basename
 from django.conf import settings
 
+try:
+    from os import scandir
+    del scandir
+    del os
+    import os as scandir
```
iksaif: I'm not sure why we need 
```
+    del scandir
+    del os
+    import os as scandir
```

Maybe add a comment ?

The canonical way to import scandir seems to be:
```
# Use the built-in version of scandir/walk if possible, otherwise
# use the scandir module version
try:
    from os import scandir, walk
except ImportError:
    from scandir import scandir, walk
```

olevchyk: Thank you. modified as suggested.

webapp/tests/test_finders.py
```diff
@@ -240,6 +239,7 @@ def test_terminal_globstar(self):
             self.wipe_whisper()
 
 
+
```
cbowman0: Is this extra line intentional?

olevchyk: no, removed , thank you

cbowman0:
Looks good to me.  👍

DanCech:
Looking good 👍

deniszh:
Thanks for reviews, guys! Merging.

And thank @olevchyk for your contribution!

PR #1820, opened 2017-02-10
add json format option for find api

DanCech:
This patch adds a lightweight json output format option for find requests.  It was initially implemented as part of the work on clustering.  The json format is pretty simple, supporting leaf and branch nodes and looks like this:
```
[
  {
    "is_leaf": true,
    "path": "carbon.agents.dev-a.cpuUsage",
    "intervals": [
      {"start": 1478989692.654787, "end": 1486765671.553646}
      ...
    ]
  },
  {
    "is_leaf": false,
    "path": "carbon.agents.dev-a"
  }
  ...
]
```
It also supports the jsonp parameter with the same semantics as the existing `completer` format and is roughly 20% more compact than the existing `pickle` encoding.

### DIFF
--- a/webapp/graphite/intervals.py
+++ b/webapp/graphite/intervals.py
@@ -19,6 +19,12 @@ class IntervalSet:
   def __iter__(self):
     return iter(self.intervals)
 
+  def __len__(self):
+    return len(self.intervals)
+
+  def __getitem__(self, i):
+    return self.intervals[i]
+
   def __nonzero__(self):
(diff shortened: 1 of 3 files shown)


deniszh:
Nice thing, quite clear.
LGTM 👍

DanCech:
Good call @deniszh, added tests for the json format and expanded the pickle tests

deniszh:
Cool, looks much better now!

PR #1821, opened 2017-02-11
supporting refactors for #1818

DanCech:
This patch implements support for the REMOTE_STORE_FORWARD_HEADERS config option, and adds some supporting changes for #1818.  The general approach is the same as #1736 and if desired we could apply that separately which (after some cleanup & conflict resolution) would reduce this patch to an update to  pass an explicit "now" value to backends, make the requestContext available in more places, and some other supporting refactor work to reduce the size of the #1818 patch.

### DIFF
--- a/webapp/graphite/local_settings.py.example
+++ b/webapp/graphite/local_settings.py.example
@@ -245,7 +245,7 @@
 # used.
 #CLUSTER_SERVERS = ["10.0.2.2:80", "10.0.2.3:80"]
 
-# This settings control wether https is used to communicate between cluster members
+# This setting controls whether https is used to communicate between cluster members
 #INTRACLUSTER_HTTPS = False
 
 # These are timeout values (in seconds) for requests to remote webapps
--- a/webapp/graphite/remote_storage.py
+++ b/webapp/graphite/remote_storage.py
@@ -4,12 +4,12 @@ from urllib import urlencode
 from threading import Lock
 from django.conf import settings
 from django.core.cache import cache
+from graphite.intervals import Interval, IntervalSet
 from graphite.node import LeafNode, BranchNode
-from graphite.readers import FetchInProgress
 from graphite.logger import log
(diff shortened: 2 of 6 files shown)


deniszh:
It's not really clear what relation of this patch to #1818 - is it required for it?
Also tests are broken. :(

DanCech:
@deniszh missed the tests, just needed a couple of tweaks from clustering_skip_find.  All sorted now.

So the relationship is that the functionality in #1818 requires that `requestContext` be passed down into the readers.  This patch implements that change with minimal changes to the core functionality itself, so we can reduce the overall size of the #1818 patch and keep things more "bite-sized".

I didn't realize when doing this that #1736 already existed, so at this point I'd suggest that we first look at merging that PR, then I can rebase this branch and the patch set here will be reduced to the structural changes needed to support #1818 as well as some assorted cleanups.

Also of note, the large change in `datalib.py` is due to moving `_fetchData` out of `fetchData`, the logic changes involved there are minimal and can be seen more clearly with whitespace changes disabled https://github.com/graphite-project/graphite-web/pull/1821/files?w=1#diff-9cfc4413865fc2a2f6942d9a215fc4e1

DanCech:
Actually I can open a separate PR just for moving `_fetchData`, which should clarify things here also.

DanCech:
@deniszh this should be a lot clearer now

obfuscurity:
I'm generally :+1: in favor of this but I'd also like to see #1736 go in first and then add the supporting changes for #1818.

DanCech:
@obfuscurity we can do that, but it'll need to be rebased and have the requestContext stuff wired up properly since right now it doesn't forward headers for render requests because of the missing link in node.fetch

DanCech:
This is now just various small refactors to support #1818

deniszh:
Strange, test was flaky, done after second run

PR #1822, opened 2017-02-11
move _fetchData out of fetchData

DanCech:
This patch simply moves the definition of `_fetchData` out of `fetchData` and makes it a separate top-level function.

The advantage of this change is that it makes it easier for a user to understand the logic flow in `fetchData` as it's no longer interrupted midway by the definition of `_fetchData`.


DanCech:
See https://github.com/graphite-project/graphite-web/pull/1822/files?w=1 to ignore the outdenting of `_fetchData`

PR #1824, opened 2017-02-12
pass consistent 'now' value to backend nodes

DanCech:
This patch sends a 'now' query parameter in requests to backend hosts in the same way as the 0.9.x branch.

It also passes `requestContext` through to `RemoteReader`, where it isn't used yet but will be by #1736 and #1818

### DIFF
--- a/webapp/graphite/node.py
+++ b/webapp/graphite/node.py
@@ -25,8 +25,13 @@ class LeafNode(Node):
     self.reader = reader
     self.is_leaf = True
 
-  def fetch(self, startTime, endTime):
-    return self.reader.fetch(startTime, endTime)
+  def fetch(self, startTime, endTime, now=None, requestContext=None):
+    try:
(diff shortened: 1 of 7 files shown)


webapp/graphite/node.py
```diff
@@ -25,8 +25,13 @@ def __init__(self, path, reader):
     self.reader = reader
     self.is_leaf = True
 
-  def fetch(self, startTime, endTime):
-    return self.reader.fetch(startTime, endTime)
+  def fetch(self, startTime, endTime, now=None, requestContext=None):
+    try:
```
deniszh: Is that `try..except` block really needed? I mean, when it could fail?

DanCech: It's needed to support readers that don't accept the new parameters, we could go through and update all the readers and tests to accept them, but that would still leave people with 3rd-party readers needing to update them for no good reason.

deniszh: Good, that's clear now.

PR #1826, opened 2017-02-14
Returned "target" field should reflect input target

mayurmahajan:


deniszh:
Hello @mayurmahajan,
Could you please elaborate this PR? I mean what you trying to achieve here.

Thanks!

PR #1829, opened 2017-02-21
perSecond is nonNegativeDeriviative, not Derivative

brian-brazil:
The difference in semantics is important, so clarify.


deniszh:
I would say that derivative here is name of math function, and not Graphite function. But semanically you're right ofc.
Merging.

PR #1832, opened 2017-02-21
Exponential Moving Average function implementation

cmaurer:


deniszh:
Merging because it's already reviewed in #1792.
Ran tests couple of times - looks stable now.

@DanCech - could you please check possible solutions for whisper tests? Thanks!

Thanks for your contribution, @cmaurer !

PR #1839, opened 2017-02-27
Feature/mon 160365 2

mayurmahajan:
This PR is to introduce a new field that reflects input target.

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -103,6 +103,8 @@ def renderView(request):
     else:
       cachedData = None
 
+    new_old_target_map = {}
+
     if cachedData is not None:
       requestContext['data'] = data = cachedData
     else: # Have to actually retrieve the data now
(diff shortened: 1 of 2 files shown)


deniszh:
Hello Mayur,
Could you please explain a bit why do you using this change?

mayurmahajan:
Hi Deniszh, 

Sorry, for late response and thanks for following up. This PR is to fix this issue: https://github.com/graphite-project/graphite-web/issues/248.  The response "target" field sometimes does not show what was sent as in target field in the URL. This field is not making change in the response target field (to maintain backwards compatibility), however adding a new field input_target that reflects what's been sent as an input target.  I will add another commit to get the unit test checks pass.

DanCech:
@mayurmahajan can you resolve the conflicts?

deniszh:
Any objections to merge this?

iksaif:
Could you add to the commit message an example of data before and after ?

PR #1845, opened 2017-03-02
Pre-load each graph image before updating the dashboard UI

davidhusselmann:
This stops the Dashboard screen from flashing blank images and being distracting to look at.

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -1070,7 +1070,14 @@ function updateGraphRecords() {
     if (!params.uniq === undefined) {
         delete params["uniq"];
     }
-    item.set('url', document.body.dataset.baseUrl + 'render?' + Ext.urlEncode(params));
+	
+	//Preload the image and set it into the UI once it is available.
+	var img = new Image();
+    img.onload = function() {
+      item.set('url',img.src);
+    };
+    img.src = document.body.dataset.baseUrl + 'render?' + Ext.urlEncode(params);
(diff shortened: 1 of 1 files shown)


deniszh:
I'm 🤷‍♂️ about that change. Any opinions? /cc @graphite-project/committers ?

davidhusselmann:
I just figured I'd make a quick change to make it easier for me to use (my graphite server is quite slow so the images take a few seconds to load up) - figured I'd PR it in case the community can also use it.

DanCech:
seems reasonable enough at first glance

deniszh:
OK, will merge tomorrow if looks good

webapp/content/js/dashboard.js
```diff
@@ -1070,7 +1070,14 @@ function updateGraphRecords() {
     if (!params.uniq === undefined) {
         delete params["uniq"];
     }
-    item.set('url', document.body.dataset.baseUrl + 'render?' + Ext.urlEncode(params));
+	
+	//Preload the image and set it into the UI once it is available.
+	var img = new Image();
```
iksaif: indentation seems off here

deniszh:
Ok, merging that. Ident is not so critical in JS, we can fix it later.

davidhusselmann:
Thanks @deniszh.  I might do another enhancement to this soon as it isn't too obvious when images have finished loading when you do a full screen reload (not automatic refresh).

PR #1847, opened 2017-03-09
Add movingMin and movingMax functions

AlexAkulov:
Add movingMin and movingMax functions
We use these functions in alerting system to avoid trigger flapping. I think it could be useful for others.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -1200,6 +1200,122 @@ def movingSum(requestContext, seriesList, windowSize):
 
   return result
 
+def movingMin(requestContext, seriesList, windowSize):
+  """
+  Graphs the moving minimum of a metric (or metrics) over a fixed number of
+  past points, or a time interval.
+
+  Takes one metric or a wildcard seriesList followed by a number N of datapoints
+  or a quoted string with a length of time like '1hour' or '5min' (See ``from /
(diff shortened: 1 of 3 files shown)


webapp/graphite/render/functions.py
```diff
@@ -1200,6 +1200,122 @@ def movingSum(requestContext, seriesList, windowSize):
...
+
+  Example:
+
+  .. code-block:: none
+
+    &target=movingMax(Server.instance01.requests,10)
+    &target=movingMax(Server.instance*.errors,'5min')
+
+  """
+  if not seriesList:
```
iksaif: A lot of code seems common between this, movingMin, and probably other moving* functions. Do you think you could try to refactor it ?

DanCech: It looks like the moving* functions share 99% of their code, differing only in the aggregation function applied (safeMin, safeMax, etc) and the name of the new series, so at first glance it seems like it should be possible to create a helper function that takes a seriesList, period, aggregation function and name template that all the moving* fructions can use.

deniszh:
Looks like everybody proposing to refactor moving* functions, but nobody really wants to do so. :)
(no offense intended, just fact of life 😈 ). So, I opened issue #1854 for that and added it to v1.1.0 milestone.
I think we should merge this PR now and proceed with refactor later. Any bjections?
@DanCech @iksaif @AlexAkulov ?

PR #1850, opened 2017-03-14
debian 8 at least needs build-essential

tahoe:
debian 8 needs build-essential or at least some of it's components to compile twisted


deniszh:
It will not hurt, I think.

PR #1681, opened 2016-09-06
Prevent repeated series evaluations for hitcount and smartSummarize

ellipses:
Changes smartSummarize and hitcount to use the original wildcard expression rather than evaluate each series individually similar to https://github.com/graphite-project/graphite-web/commit/53d1b3dee8f99c03e18d96909a12147f4a161030  .

Since smartSummarize is now using the original expression, queries such as summarize(metric.{a,b}) also work. 

Thomas (Hosted Graphite)

PR #1855, opened 2017-03-18
prevent repeated series evaluations for hitcount and smartSummarize

ellipses:
Changes smartSummarize and hitcount to use the original wildcard expression rather than evaluate each series individually similar to https://github.com/graphite-project/graphite-web/commit/53d1b3dee8f99c03e18d96909a12147f4a161030  .

Since smartSummarize is now using the original expression, queries such as summarize(metric.{a,b}) also work. 

Same pr as https://github.com/graphite-project/graphite-web/pull/1681 but this time without an insane diff from messed up rebase. @deniszh :)


deniszh:
Thanks, @ellipses !
Could I please get more reviews from @graphite-project/committers ?

DanCech:
Looks good

PR #1884, opened 2017-03-29
remote storage fixes

iksaif:


deniszh:
Cool!
👍

yunstanford:
I have a graphite cluster that has 2 shards (data are different on these two nodes based on consistent hashing). And we also have Master Graphite-Web talk with these two shards and merge the results. But..... there are some chance (very low, about 1/60) that Master graphite-web just return data from one shard. but i can't find any exception from log around this. Does this fix the issue ? any idea ?

deniszh:
They're counters, indeed. When you running script from command line it's
obviously not - but when running from collectd values should be
monotonically increased. You can enable debug and check all output in
system logs.
Do you have Graphite connected to Collectd or only rrd?

yunstanford:
we have collectd, statsite, monitoring system connected to the graphite cluster.

each shard includes a graphite-web, a carbon-relay-ng, 8 carbon-cache.

a master graphite-web fetch remote data from these two shards and merge them. (there is no data on master graphite-web host)

a carbon-relay-ng does consistent hashing in front of these two shards. (for writing)

writing part works fine based on consistent hashing. the thing is that our Master graphite-web sometimes(very unlikely, but i do observe) just return data from one shard, although most time it works fine. (basically it doesn't get data from the other shard).

iksaif:
Can you open a dedicated issue for that with the exact version that you are using ? 

@deniszh this patch was wrongly sent to 1.0.0 only. I'll spend some time on Monday to cherry-pick whatever we need on 1.0.0 and merge back everything that didn't make it to master.

PR #1888, opened 2017-04-03
settings: reduce default number of threads

iksaif:
See #1885

### DIFF
--- a/webapp/graphite/local_settings.py.example
+++ b/webapp/graphite/local_settings.py.example
@@ -248,18 +248,23 @@
 # Creates a pool of worker threads to which tasks can be dispatched. This makes
 # sense if there are multiple CLUSTER_SERVERS because then the communication
 # with them can be parallelized
+# The number of threads is equal to:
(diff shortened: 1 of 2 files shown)


deniszh:
`POOL_WORKERS_PER_BACKEND = 1`
But does reducing `POOL_WORKERS` to 1 making all pool thing useless?

iksaif:
I think that's fine, you end up with between 2 and N threads (with N-1 being the number of backends). That's usually way enough since you're not going to process requests in parallel most of the time.

Remember that the total number of threads is `POOL_WORKERS_PER_BACKEND * len(CLUSTER_SERVERS) + POOL_WORKERS`.

deniszh:
>Remember that the total number of threads is POOL_WORKERS_PER_BACKEND * len(CLUSTER_SERVERS) + POOL_WORKERS.

Cool! Maybe it's worth to mention that somewhere in https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/local_settings.py.example#L248-L262 ?

iksaif:
Updated :)

deniszh:
Cool! 👍

PR #1890, opened 2017-04-03
docs: add biggraphite, mark ceres as unmaintained.

iksaif:


deniszh:
👍 
Maybe it's worth to mention the same on https://github.com/graphite-project/ceres
Will fix release notes too.

Issue #1886, opened 2017-03-31
Importing graphite.carbonlink is hard


iksaif:
carbonlink.py does a bunch of stuff in the body of the module, including using settings, which make it very problematic to import in finder/reader plugins. Even worse it's being imported by readers.py

I'm trying to refactor this code to delay the initialization of CarbonLinkPool()

@graphite-project/committers , do you think it would be fine to break the API here, and make carbonlink.CarbonLink a callable instead of an instance ?

deniszh:
I'm not sure that an API officially, and I'm not sure that many people using that out there, so I'm OK with that.

PR #1891, opened 2017-04-03
carbonlink: make it easilly importable, delay initialization

iksaif:
Also make readers.py pull less dependency and add
graphite.singleton to make it easier to have singleton.
This could also be used for some of our other singletons
to allow easier lazy-initialization.

See #1014 , #1886

### DIFF
--- a/webapp/graphite/carbonlink.py
+++ b/webapp/graphite/carbonlink.py
@@ -1,13 +1,15 @@
 import time
 import socket
 import struct
-import errno
 import random
-from select import select
+
 from django.conf import settings
+
 from graphite.render.hashing import ConsistentHashRing
 from graphite.logger import log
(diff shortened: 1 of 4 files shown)


iksaif:
Also see https://github.com/criteo/biggraphite/pull/259

iksaif:
(wait before reviews, some things apparently don't work yet)

iksaif:
ok good now :)

iksaif:
@deniszh : I'll probably use the same singleton.py in carbon as there are a few singleton that could benefit from it

wdauchy:
:+1:

deniszh:
>I'll probably use the same singleton.py in carbon too as there are a few singleton that could benefit from it

Yep, that's a good idea ! 👍 
Thanks!

PR #1893, opened 2017-04-05
CarbonLink() must return singleton to be accesible

chriwill:
Sympton:
```
2017-04-05,17:45:51.784 :: Failed CarbonLink query 'carbon.agents.casar1_fyre_ibm_com-c.cache.queries'
  File "/opt/graphite/webapp/graphite/readers.py", line 211, in fetch
    cached_datapoints = CarbonLink().query(self.real_metric_path)
AttributeError: 'NoneType' object has no attribute 'query'
```

GlobalCarbonLinkPool.instance() is called to return global singleton, but the return function is missing. This causes carbon queries to fail with above message.


iksaif:
woops

PR #1907, opened 2017-04-20
correct install instructions

jvanasco:
I thought I was having a stroke because I couldn't understand the install instructions.  Then I realized they were just wrong.  I'm guessing this got mutilated by a find/replace, and my PR is somewhat closer to what was intended?

at this point in the install, `$GRAPHITE_ROOT/webapp` only has the graphite_web python package as `graphite` and `graphite_web-1.1.0-py2.7.egg-info`

`django-admin.py` is located in $GRAPHITE_ROOT/bin/

### DIFF
--- a/docs/config-local-settings.rst
+++ b/docs/config-local-settings.rst
@@ -86,7 +86,11 @@ STATIC_ROOT
 
   This directory doesn't even exist once you've installed graphite. It needs to be populated with the following command::
 
-      PYTHONPATH=$GRAPHITE_ROOT/webapp django-admin.py collectstatic --noinput --settings=graphite.settings
+      $GRAPHITE_ROOT/bin/django-admin.py collectstatic --noinput --settings=graphite.settings
+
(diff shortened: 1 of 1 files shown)


deniszh:
Hello @jvanasco,
> django-admin.py is located in $GRAPHITE_ROOT/bin/

Sorry, but that's not true. `django-admin.py` is part of Django and it's location depends on your Django installation path - it could be '/usr/bin', '/usr/local/bin' etc. :
```
$ which django-admin.py
/usr/local/bin/django-admin.py
```
Usually this path is belong to system path, that's why documentation just ask to run `django-admin.py` - without full path.

jvanasco:
Sorry, I just realized where the issue happened as I got deeper into the (re)install.

That command needs `scandir` to run, which doesn't get installed via dependencies somehow.

installing that, then running it, works -- and runs a handful of tasks that my method didn't pick up .

PR #1908, opened 2017-04-20
Duplicated results when using remote prefetch with multiple targets

DanCech:
This patch fixes an issue when using multiple targets in a single query with remote prefetch enabled.

Without this patch, a query with 2 targets will return the results twice, a query with 3 targets will return them 3 times, etc.

To test, try a query like `ttp://localhost:8000/render?format=json&target=collectd.*.cpu-0.cpu-idle&target=collectd.*.cpu-1.cpu-idle&from=-1min`


deniszh:
Good catch!

PR #1906, opened 2017-04-20
Only fetch user profile when it's needed

deejay1:
This should avoid calling the database for every /metrics/find
request in clustered environments.

PR #1910, opened 2017-04-21
Only fetch user profile when it's needed

deejay1:
This should avoid calling the database for every /metrics/find
request in clustered environments.


deejay1:
This is a "backport" of https://github.com/graphite-project/graphite-web/pull/1906

deniszh:
Thanks!

Issue #1915, opened 2017-04-26
Improve render performance


wy92:
We have encountered some performance issues when many data points are involved in functions computation, such as summarize, sumSeries, etc. 

For example, our metric's resolution is mainly `10s:1d`,  `summarize(span.*.client.*.app1.*.request_time.percentile.95, "7d")`, this query may involve many metrics and cost more than 10 seconds, we discover that the cpu utility is almost 100%, and mainly is user cpu, so we assume the bottleneck is in `summarize`.

We are working on it, and have some ideas, we would like to know what is the official plan about this and see if we can work together.

This is our initial idea:
* we first parse target, generate a plan and merge plan, and then execute the plan.
* assume the query's from=`from1`, until=`until1`，we separate the interval into N partitions, and process each partition with a new graphite-web request.

deniszh:
The official plan - made Graphite better, without breaking it. If you're proposing some radical changes - we can merge it to master and release later, in some 1.1.x release. If your changes fixing some problem and are not bringing troubles - we can even include that in a current stable release.

deniszh:
Ah, one more thing - all breaking changes should be configurable by config file and disabled by default.

DanCech:
Looking at summarize there are definitely some gains to be made internally since right now it uses a multi-step process and generates an intermediate list of buckets.  I have a rough patch to reduce that so it only walks the data once, you can try that out in https://github.com/graphite-project/graphite-web/tree/summarize if you're interested.  In my testing I'm seeing about a 40% reduction in the time spent in summarize()

deniszh:
@wy92 - sorry for bugging you, but did you have a chance to test @DanCech's patch?

reyjrar:
Was looking for a place to converse, but I'm seeing another major performance degradation on queries with large numbers of metrics.  In older 0.9.15 versions of graphite, I could get a dataset of 10,000 metrics to return in under 20 seconds.. Now in 1.0.2, that same query is taking upwards of 90 seconds.  Group, filtering, and combining functions also fail because the fetch for the data takes too long.

I'm stuck on pre-1.x.x because most of queries select data from thousands of metrics and display the most interesting ones.  I think this is related to the OP's bug.

PR #1916, opened 2017-04-26
improve efficiency of summarize and smartSummarize functions

DanCech:
This improves the internal efficiency of the summarize and smartSummarize functions to help improve the performance issue discussed in #1915 

As far as I can tell it should be logically identical to the existing code, but I'd appreciate any feedback.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -3765,45 +3765,50 @@ def smartSummarize(requestContext, seriesList, intervalString, func='sum', align
   interval = delta.seconds + (delta.days * 86400)
   results = []
   for series in seriesList:
-    buckets = {} # { timestamp: [values] }
-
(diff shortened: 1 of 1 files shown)


deniszh:
LGTM, but is it really gives much gain in performance?

DanCech:
@deniszh it depends on how many data points you have, with a large number of points I was seeing about a 40% decrease in runtime compared to master which could easily be several seconds difference.

deniszh:
@DanCech - should we merge it to master?

DanCech:
If it looks good to you and @iksaif I don't see why not.

deniszh:
For me it's fine 👍

pipamc:
@DanCech Can I have your test if you feel Okey? In my compute, I can't get so high improvement.

DanCech:
It's been a while, IIRC I was just running requests using summarize over large amounts of data from collectd and collecting timing information.

PR #1931, opened 2017-05-13
Update local_settings cluster config documentation

b13n1u:
Looks like the docu is still showing the old config (pre 1.0) arguments for the cluster settings which is a little bit misleading. I have modified and added the arguments based on local_settings.conf.example from 1.0.1.


deniszh:
Thanks @b13n1u !

Issue #1936, opened 2017-05-18
[Feature request] Query in alias


mageo:
Hello,

I would like to alias metrics based on the value of (an)other metric(s).

For example, if a have the following metrics:

- channel.power.1
- channel.frequency.1 (last value = 565)

I would like to alias the first one using the value of the second, so that when I'm graphing channel.power.*, the label for the metric looks like "Channel 565 MHz".

This would require:

- To extract the meaningful part from the first metric, with a regex for example: /channel\.power\.([0-9]+)/
- To build the metric to query using the result of the previous regex: "channel.frequency.$1"
- To build the final alias using the value returned from the second metric: "Channel $1 MHz"

It could be written as a single function: aliasQuery('/channel\.power\.([0-9]+)/', 'channel.frequency.$1', 'Channel $1 MHz')

Do you think this would be feasible ?

PR #1928, opened 2017-05-11
divideSeriesLists function

leoleovich:
Iterates over a two lists and divides list1[0] by list2[0], list1[1] by list2[1] and so on.

PR #1937, opened 2017-05-19
Issue #1936 - aliasQuery function

mageo:
Hello,

Following my feature request #1936, I had a look to the code and it appeared not so difficult to implement it by myself.

Here is the result.

I hope someone else may find this useful.

Code has been tested in production.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -1734,6 +1734,27 @@ def aliasSub(requestContext, seriesList, search, replace):
       series.name = re.sub(search, replace, series.name)
   return seriesList
 
+def aliasQuery(requestContext, seriesList, search, query, newName):
+  """
+  Performs a query to alias the metrics in seriesList.
+
+  .. code-block:: none
+
+    &target=aliasQuery(channel.power.*,"channel\.power\.([0-9]+)","channel.frequency.\\1", "Channel \\1 MHz")
+  """
+  for serie in seriesList:
+    newQuery = re.sub(search, query, serie.name)
+    newSeriesList = evaluateTarget(requestContext, newQuery)
+    if newSeriesList is None or len(newSeriesList) == 0:
+      raise Exception('No serie found with query: ' + newQuery)
+    current = safeLast(newSeriesList[0])
+    if current is None:
+      raise Exception('Cannot get last value of serie: ' + newSeriesList[0])
+    current = safeLast(newSeriesList[0])
+    newSerieName = re.sub(u'(.*)', newName, current)
+    serie.name = newSerieName
+  return seriesList
+
 def alias(requestContext, seriesList, newName):
   """
   Takes one metric or a wildcard seriesList and a string in quotes.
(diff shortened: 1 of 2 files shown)


deniszh:
Hello @mageo,
Thanks for your contribution, much appreciated!
But we have strong policy about testing functions - would you mind to add tests to your function?
You can check examples in webapp/tests/test_functions.py or check other PRs, e.g. https://github.com/graphite-project/graphite-web/pull/1928

mageo:
Hello @deniszh,
I will have a look at the examples and write a test function.

deniszh:
Looks good. Codacy is too picky for lambdas, can be ignored.
Let us review it.

webapp/graphite/render/functions.py
```diff
@@ -1734,6 +1734,25 @@ def aliasSub(requestContext, seriesList, search, replace):
...
+
+  .. code-block:: none
+
+    &target=aliasQuery(channel.power.*,"channel\.power\.([0-9]+)","channel.frequency.\\1", "Channel \\1 MHz")
+  """
+  for serie in seriesList:
+    newQuery = re.sub(search, query, serie.name)
+    newSeriesList = evaluateTarget(requestContext, newQuery)
+    if newSeriesList is None or len(newSeriesList) == 0:
+      raise Exception('No serie found with query: ' + newQuery)
```
cbowman0: Singular of series is still series.  At least for the error message, can you change it to be series?

cbowman0: Also, I don't think this failure condition is tested in the tests provided.  Can you add another test or two to test the failure conditions?

The rest looks good!

mageo: Sorry, series is similar in french and singular is serie. I will change that. I will also update the test function to test the failure conditions.

cbowman0: Ah.  I thought it would be something like that.  English isn't always logical...

deniszh:
Thanks, @mageo, merged!

DanCech:
I was going to say that I like this a lot but feel that it'd be more powerful if it used sprintf syntax for building the alias vs regex replacement

DanCech:
@deniszh as a more general comment, we need to be cognizant that once a new function is added to the API we're making a commitment to supporting it going forward and we should allow time for discussion to happen before we merge something like this.

deniszh:
@DanCech :
Yes, you're right, I was too fast with merging that. Usually, I wait a couple of days before merging, but these was approved so fast and looks OK...
We can fix it, though - or even revert - it's still in master, which is subject to change.

Could you please elaborate what you mean "be more powerful if it used sprintf syntax for building the alias" btw?

DanCech:
@deniszh cool, I'll throw together a PR to demo what I mean.

mageo:
@DanCech yes, your approach would give way more flexibility !

Do you think something like this would be enough ?

`series.name = newName.format(current)`

DanCech:
More or less, I have a patch that uses `series.name = newName % current` since that syntax is much closer to the standard c-style printf syntax.

https://docs.python.org/2/library/stdtypes.html#string-formatting-operations
https://en.wikipedia.org/wiki/Printf_format_string

I'm just expanding the docs a bit and I'll open a PR for that.

DanCech:
#1938

PR #1938, opened 2017-05-19
use sprintf-style format strings in aliasQuery

DanCech:
This expands on #1936 to allow more control over how the resulting aliases are formatted.


deniszh:
Ah, got your point. Much better now. 👍

PR #1942, opened 2017-05-28
finders/readers: split and clean

iksaif:
This is a first step towards cleaning the finder/reader code. The
end goal is to also clean storage.py and render/views.py to remove
implementation details and make 'remote' a real finder/reader plugin.

- split the various finders/readers but keep importing symbols to
  preserve compatibility.
- run autopep8 on every modified files.


iksaif:
See #1943

deniszh:
I'm very 👍 for idea of splitting finders/readers, much appreciated!
Not sure why lint has so much complains about importing, though...

iksaif:
The format changes are "caused" by running `autopep8` on the code that I modified. Some bits can be manually tweaked later. I'll take care of the lint errors during the day.

PR #1956, opened 2017-06-07
render: more robust maxDataPoints

iksaif:
I'm not sure why we take the widest range possible to get the
number of points to remove, but this clearly needs to be done per
timeseries as they don't have any guarantee to all have the same
time range.

### DIFF
--- a/webapp/graphite/app_settings.py
+++ b/webapp/graphite/app_settings.py
@@ -20,11 +20,6 @@ from os.path import dirname, join, abspath
 #Django settings below, do not touch!
 APPEND_SLASH = False
 TEMPLATE_DEBUG = False
-CACHES = {
-    'default': {
-        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
-    },
-}
 
 TEMPLATES = [
     {
--- a/webapp/graphite/settings.py
+++ b/webapp/graphite/settings.py
@@ -247,6 +247,13 @@ if MEMCACHE_HOSTS:
         'OPTIONS': MEMCACHE_OPTIONS,
     }
 
+if not CACHES:
+  CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.dummy.DummyCache',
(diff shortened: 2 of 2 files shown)


DanCech:
I think the idea is that the maxDataPoints value is usually set to the width of the graph in pixels (Grafana does that unless you manually override), so you want the total number of intervals in the result (across all series) to equal that number to end up with an interval per pixel.  If you do it per-series then you'll end up with more points than you have pixels to plot them on if the time ranges don't all align.

deniszh:
Yep, what @DanCech says.  What's a problem with maxDataPoints? it's not mandatory, just not use it if not needed.

iksaif:
I don't with it's for the width of the graph. Isn't this controlled by from/until ? For me it was just to avoid drowing / tranmitting points that would not be useful.

The current code doesn't seem to work if there is an empty timeseries. Maybe `valuesToLose = min(int(nudge/series.step), len(series)` is better ?

deniszh:
>I don't with it's for the width of the graph. Isn't this controlled by from/until ? For me it was just to avoid drowing / tranmitting points that would not be useful.

No, from/until will limit your timespan, but not a number of data points. If you have from=-1y and second resolution (without aggregation ofc) you'll get json with 31536000 points, which definitely kill your browser, which will try to fit it into 1000 pixels on screen width anyway. maxDataPoints was introduced in graphite-api for work with Grafana because of that issue.

> The current code doesn't seem to work if there is an empty timeseries.
Why? Could you elaborate a bit more? The current code/logic is working in Graphite starting from 0.9.14, graphite-api and caronapi, nobody complaint about much about it before.
We can introduce another smoothing algorithm for maxDataPoints ofc, but IMO it should work as intended - return not more than maxDataPoints. :)

iksaif:
Closing for now as I lost my patch by mistake. I also fix the finder that was returning an empty timeseries so I guess we will see next time it happens.

PR #1960, opened 2017-06-14
Django 1.10 and 1.11 support

brutasse:
Various changes related to supporting more Django version: 1.9 is no longer receiving security updates. Django 1.11 is the last in the 1.x series and is a LTS. Upcoming Django 2.0 will be python-3 only.

I haven't dropped Django 1.9 here, but I'd be in favor of removing support either now or immediately after next release. Thoughts?


deniszh:
Hi @brutasse,
Great stuff, as usual! 👍 
I'm very 👍 for supporting Django LTS and dropping 1.9 after next release.

deniszh:
Merging, if no objections.

setup.py
```diff
@@ -96,7 +100,7 @@
         ['templates/*', 'local_settings.py.example']},
       scripts=glob('bin/*'),
       data_files=webapp_content.items() + storage_dirs + conf_files + examples,
-      install_requires=['Django>=1.9,<1.9.99', 'django-tagging==0.4.3', 'pytz', 'pyparsing', 'cairocffi', 'urllib3', 'scandir'],
+      install_requires=['Django>=1.9', 'django-tagging==0.4.3', 'pytz', 'pyparsing', 'cairocffi', 'urllib3', 'scandir'],
```
DanCech: Should we have `,<1.11.99` here and in `requirements.txt` in case of breaking changes in future releases?

tox.ini
```diff
@@ -25,7 +25,9 @@ deps =
 	git+git://github.com/graphite-project/whisper.git#egg=whisper
 	git+git://github.com/graphite-project/ceres.git#egg=ceres
 	pyparsing2: pyparsing
-	django19: Django>=1.9,<1.9.99
+	django19: Django>=1.9,<1.10
+	django110: Django>=1.10,<1.11
+	django111: Django>=1.11,<1.12
```
DanCech: These will allow the `django19` env to pull in pre-release versions of `1.10` etc, we want to keep the same requirement pattern as we had before.

deniszh:
Agreed, will make a fix now.

deniszh:
@DanCech, https://github.com/graphite-project/graphite-web/pull/1964 - please review

Issue #1957, opened 2017-06-09
holtWintersAberration no longer works correctly in 0.9.16


bemeyert:
Hi all,

We just upgraded from 0.9.15 to 0.9.16 and our graphs using holtWintersAberration no longer work correctly. When we set "from" to less than 9 days we get the following error:
```
Traceback (most recent call last):
  File "/usr/lib/python2.7/site-packages/django/core/handlers/base.py", line 112, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/lib/python2.7/site-packages/graphite/render/views.py", line 123, in renderView
    seriesList = evaluateTarget(requestContext, target)
  File "/usr/lib/python2.7/site-packages/graphite/render/evaluator.py", line 10, in evaluateTarget
    result = evaluateTokens(requestContext, tokens)
  File "/usr/lib/python2.7/site-packages/graphite/render/evaluator.py", line 21, in evaluateTokens
    return evaluateTokens(requestContext, tokens.expression)
  File "/usr/lib/python2.7/site-packages/graphite/render/evaluator.py", line 31, in evaluateTokens
    return func(requestContext, *args)
  File "/usr/lib/python2.7/site-packages/graphite/render/functions.py", line 2078, in holtWintersAberration
    elif upperBand[i] is not None and series[i] > upperBand[i]:
IndexError: list index out of range
```
The request looks like this:
```
http://metrics.server/render?&rawData=true&format=json&from=-2d&until=now&target=holtWintersAberration(groupByNode(scaleToSeconds(serieslist,%201),%207,%20%27sum%27),%2010)
```
This worked fine in 0.9.15. Since we use CentOS 7 the latest available version is 0.9.16. 

Thx for any help

deniszh:
Looks like it's caused by https://github.com/graphite-project/graphite-web/commit/2e940a34d9aae0a27d1629dd1db8255763aeef51
which was backport of fix for issue #220 from a master.
So, I think version 1.0.x also contain this. 
Not sure how that patch could cause that, though. `series[i]` is defined, otherwise it should fail on line 1975 instead, so it's caused by bad `upperBand[i]` - but it was there before and reverting this patch doesn't help.
And I do not see any other changes in holtWintersAberration.
So, I'm suspect that you just getting NULLs in your data when getting older data, maybe because of wrong aggregation policy. Could you please check that? just remove holtWintersAberration and check rest of your data.

bemeyert:
[...]

PR #1961, opened 2017-06-15
Introducing configurable previewDays for Holt-Winters functions

deniszh:
previewDays (7 days by default) is original interval for bootstrapping HW coefficients.
Probably fixing issue #1957

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2812,12 +2812,12 @@ def holtWintersAnalysis(series):
         }
   return results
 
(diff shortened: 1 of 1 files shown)


deniszh:
will make another change, with settings instead of days.

DanCech:
@deniszh I'd suggest the same syntax as movingAverage and summarize use, there's already a dedicated parser for it.

deniszh:
@DanCech: Ah, very good suggestion, thanks!

deniszh:
How does it looks now - https://github.com/graphite-project/graphite-web/pull/1965 ?

PR #1962, opened 2017-06-16
LDAP broken since Graphite 1.0

GLundh:
LDAP_BASE_USER/LDAP_BASE_PASS setting has been broken since
the LDAP_USER_DN_TEMPLATE setting was introduced in commit
871a6cb427cbb1d8a1c33905d6f25e4b39d8a5b7.

By not giving LDAP_USER_DN_TEMPLATE a default value in
settings.py, the expected attribute could not be fetched in
ldapBackend.py and logins fails with:

AttributeError: 'Settings' object has no attribute
	'LDAP_USER_DN_TEMPLATE'


deniszh:
Thanks!

Should be backported to 1.0.x too...

PR #1965, opened 2017-06-16
Introducing bootstrapInterval for HoltWinters functions

deniszh:
Default is '7d' - one week, as before. Format - same as for all others time intevals in Graphite.
Fixing #1957


deniszh:
Yes, agreed. Now all logic in HW bootstrap is based on seconds, so, decided to not put more complexity there, at least now.

deniszh:
It's fixing the issue, so, merging.

Issue #1967, opened 2017-06-16
What are the reasons we don't support Django 1.8?


cbowman0:
I can't find any reasons in prior issues or PRs.

1.8 is one of the LTS releases.  It would make installation on Ubuntu far easier for me personally, since my environment prefers to use the OS packages.

deniszh:
It slipped out between releases - too old for 1.0.x but too new for 0.9.x.
I'm totally 👍 for supporting it but have no idea TBH what need to be done.
Maybe @brutasse knows?

piotr1212:
I'm the Fedora/EPEL Graphite package maintainer. If not too much trouble I would be quite happy with 1.8 support in Graphite. Django packaging is a bit of a pain in EPEL. Supporting 1.8 would make it easier to get Graphite 1.x in EPEL 7.

cbowman0:
Support for Django 1.8 was added to 1.0.x and master branches.  Next release(s) will contain support unless some major issue is found.

PR #1968, opened 2017-06-17
Django 1.8

cbowman0:
Add Django 1.8 to the tests and other relevant sections.

As discussion in issue #1967.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,7 +37,7 @@
 #   deactivate
 #
 
-Django>=1.9
+Django>=1.8
 python-memcached==1.47
 txAMQP==0.4
 simplejson==2.1.6
--- a/requirements.txt
+++ b/requirements.txt
@@ -37,7 +37,7 @@
 #   deactivate
 #
 
-Django>=1.8
+Django>=1.8,<1.11.99
 python-memcached==1.47
 txAMQP==0.4
 simplejson==2.1.6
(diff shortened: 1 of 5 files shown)


deniszh:
But does it work with Django 1.8 at all btw?

This should be ported to 1.0.x branch then.

cbowman0:
I have found no issues in my testing​ so far.  I want try try a few more items before merging.

requirements.txt
```diff
@@ -37,7 +37,7 @@
 #   deactivate
 #
 
-Django>=1.9
+Django>=1.8
```
DanCech: Can we add `,<1.11.99` here to match setup.py?

cbowman0: Yes.  Done.

cbowman0:
I've found no issues.  Does anyone have anything specific to check?

deniszh:
I think we can merge it.
Not sure do we need to backport to 1.0.x and release 1.0.2 or release 1.1.0 instead?
Maybe 1.0.2 is better, master is still very fresh.

DanCech:
Yeah, we still need to do more testing on the refactored clustering before I'd consider it stable enough to cut 1.1.0

cbowman0:
I submitted PR #1975 for the 1.0.x branch.  OK to merge here?

DanCech:
Please do

PR #1970, opened 2017-06-21
Remote storage fixes

iksaif:
This set fixes two bugs I discovered with the latest changes:
* When *not* using REMOTE_PREFTECH_DATA the code checking in flight requests suffered from a performance regression (re-fetching urls each time)
* When using REMOTE_PREFTECH_DATA and `timeShift()` the wrong data was behind used since we didn't check the start and end time of the query.

### DIFF
--- a/webapp/graphite/finders/remote.py
+++ b/webapp/graphite/finders/remote.py
@@ -75,12 +75,12 @@ class RemoteFinder(BaseFinder):
                 else:
                     continue
 
-            log.debug("Got a find result after %fs" % (time.time() - start))
+            log.debug("Got a remote find result after %fs" % (time.time() - start))
             done += 1
             for node in nodes or []:
                 yield node
 
-        log.debug("Got all find results in %fs" % (time.time() - start))
+        log.debug("Got all remote find results in %fs" % (time.time() - start))
 
     def fetch(self, nodes_or_patterns, start_time, end_time, now=None, requestContext=None):
         # Go through all of the remote nodes, and launch a fetch for each one.
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -12,13 +12,19 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License."""
 
-from graphite.logger import log
-from graphite.storage import STORE
-from graphite.readers.utils import wait_for_result
-from django.conf import settings
-from graphite.util import timebounds, logtime
+import collections
 
 from traceback import format_exc
+
+from django.conf import settings
+
+from graphite.future import Future
+from graphite.logger import log
+from graphite.readers.utils import wait_for_result
+from graphite.storage import STORE
+from graphite.util import timebounds, logtime
+from graphite.render.utils import extractPathExpressions
+
 
 class TimeSeries(list):
   def __init__(self, name, start, end, step, values, consolidate='average'):
@@ -108,7 +114,18 @@ class TimeSeries(list):
 
 @logtime()
 def _fetchData(pathExpr, startTime, endTime, now, requestContext, seriesList):
-  local = settings.REMOTE_PREFETCH_DATA or requestContext['localOnly']
+  result_queue = []
+  remote_done = False
+
+  if settings.REMOTE_PREFETCH_DATA:
+    prefetched = requestContext['prefetched'].get((startTime, endTime, now), None)
+    if prefetched is not None:
+      for result in prefetched[pathExpr]:
+        result_queue.append(result)
+      # Since we pre-fetched remote data only, now we can get local data only.
+      remote_done = True
+
+  local = remote_done or requestContext['localOnly']
   matching_nodes = STORE.find(
     pathExpr, startTime, endTime,
     local=local,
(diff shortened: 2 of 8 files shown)


deniszh:
Generally looks good, but tests are still broken

iksaif:
ok, there is only one more failing test, but it really doesn't seem to be related. What do you think ?

deniszh:
Yes, that's flaky one, just need to be restarted.

iksaif:
I'll give a chance to @DanCech  to look at this :)

webapp/graphite/render/utils.py
```diff
@@ -0,0 +1,22 @@
...
+      return extractPathExpression(tokens.expression)
+    elif tokens.pathExpression:
+      pathExpressions.add(tokens.pathExpression)
+    elif tokens.call:
+      for a in tokens.call.args:
+        extractPathExpression(a)
+
+  for target in targets:
+    tokens = grammar.parseString(target)
+    extractPathExpression(tokens)
```
DanCech: Since we're refactoring this, let's promote `extractPathExpression` to a standalone function `extractPathExpression(target)` and have it parse the target into tokens internally rather than doing that in `extractPathExpressions`

webapp/graphite/render/datalib.py
```diff
@@ -167,7 +185,7 @@ def _fetchData(pathExpr, startTime, endTime, now, requestContext, seriesList):
       else:
         known_nones = len([val for val in known if val is None])
 
-      if known_nones > candidate_nones:
+      if known_nones > candidate_nones and len(series):
```
DanCech: isn't this change going to mean that an empty result will always overwrite the `seriesList` entry down in the `else` clause?

webapp/graphite/render/datalib.py
```diff
@@ -108,22 +114,34 @@ def getInfo(self):
...
 
-  result_queue = [
+  result_queue.extend([
     (node.path, node.fetch(startTime, endTime, now, requestContext))
     for node in matching_nodes
-  ]
-  prefetched = requestContext['prefetched']
-  for result in prefetched[pathExpr]:
-    result_queue.append(result)
+  ])
```
DanCech: Might be worth profiling this against:
```
for node in matching_nodes:
  result_queue.append((node.path, node.fetch(startTime, endTime, now, requestContext)))
```

webapp/graphite/render/datalib.py
```diff
@@ -243,3 +261,51 @@ def nonempty(series):
...
+  """
+  log.rendering("Prefetching remote data")
+  pathExpressions = extractPathExpressions(targets)
+
+  time_info = timebounds(requestContext)
+  (startTime, endTime, now) = time_info
+
+  results = STORE.fetch_remote(pathExpressions, startTime, endTime, now, requestContext)
+
+  requestContext['prefetched'][time_info] = PrefetchedData(results)
```
DanCech: We should be careful here, since when we pull the data back out we use the explicit `prefetched = requestContext['prefetched'].get((startTime, endTime, now), None)`, if `timebounds()` is updated to return any additional items then the lookup will fail.  This should likely be `requestContext['prefetched'][(startTime, endTime, now)] = PrefetchedData(results)`

webapp/graphite/finders/remote.py
```diff
@@ -95,6 +95,7 @@ def fetch(self, nodes_or_patterns, start_time, end_time, now=None, requestContex
             else:
                 patterns.append(v.path)
 
+        # TODO: Use the thread pool here.
```
DanCech: This does use the thread pool, in `RemoteReader._fetch_list_locked()`

webapp/graphite/render/datalib.py
```diff
@@ -108,22 +114,34 @@ def getInfo(self):
...
+
+  if settings.REMOTE_PREFETCH_DATA:
+    prefetched = requestContext['prefetched'].get((startTime, endTime, now), None)
+    if prefetched is not None:
+      for result in prefetched[pathExpr]:
+        result_queue.append(result)
+        # Since we pre-fetched remote data only, now we can get local data only.
+        remote_done = True
+
+  local = remote_done or requestContext['localOnly']
```
DanCech: I'm not 100% clear on what this is doing, it seems like if we're using prefetch but the remote nodes didn't return any results then it's going to try to query them again.

webapp/graphite/render/utils.py
iksaif: done

webapp/graphite/render/datalib.py
iksaif: wrong indentation, fixed. We only set remote_done to False if we don't query with the same start/end/now.

webapp/graphite/render/datalib.py
iksaif: changed, more readable

webapp/graphite/render/datalib.py
iksaif: Are you use you're looking at the right `else` ?

webapp/graphite/render/datalib.py
iksaif: right, this was going to break

webapp/graphite/finders/remote.py
iksaif: Ah right...

I feel like the mix of FetchInProgress + Threads + generators will need to be cleaned to use a single way to doing async stuff.

webapp/graphite/render/datalib.py
DanCech: Nope, I was looking at line 222, you're right.

webapp/graphite/render/utils.py
DanCech: Fair enough, it's definitely more involved than I realized with the way it handles `pathExpressions`

DanCech:
Looking good now

DanCech:
@deniszh this fixes an issue in 1.0.2 where timeShift doesn't work properly with remote prefetch (it returns non-shifted data), what do you think about merging it to 1.0.x and cutting 1.0.3?

DanCech:
@deniszh After thinking about this a little more I think the time may have come to release 1.1.x, since there are a bunch of other changes in master that it would be good to get into a stable release.  What do you think?

deniszh:
Indeed - I would prefer 1.1.0 release, but I want to merge tagging support
first. Sorry for being quiet lately, I'm still on vacation.

Issue #1979, opened 2017-06-26
Saving a graph with an accent in title causes hiding all saved graphs, shows "no saved graphs" in tree "My Graphs".


fidergo-stephane-gourichon:
Hi

# Summary

Logged in user cannot see any saved graphs in "My Graphs" or "User Graphs".  Short URLs do work, suggesting that something is still registered.

# Steps to reproduce

* On Ubuntu 16.04 AMD64,
* installed packages `graphite-carbon` and `graphite-web`. Version is 0.9.15-1 .
* `graphite-manage syncdb` to create database and user.
* run test instance:

```
su -s /bin/bash -c '/usr/bin/django-admin runserver --settings \
 graphite.settings 0.0.0.0:8080' _graphite
```

* insert data point in database `{ echo "foo 1 1" | nc -q 0 127.0.0.1 2003 ; echo "foo 10 10" | nc -q 0 127.0.0.1 2003 ; }`
* go to http://0.0.0.0:8080/
* log in as newly created user.
* expand tree to reveal data 'foo'
* give a title to the graph with an accent, like "Écritures"
* click on disk icon to save graph, enter name "foo", click "ok".
* see window "Save to My Graphs - Result" "Graph saved successfully"

## Expected behavior

* tree display on left shows newly created graph

## Observed behavior

* tree display on left flashes with a new entry but reverts immediately to "No saved graphs"
* refreshing the page still shows "No saved graphs"

## Additional information

* Using the "Short URL" feature does provide a URL like this: http://0.0.0.0:8080/S/B that renders the graph, showing that something was stored on the server.
* I see URLs logged in terminal where `django-admin runserver` runs, no error

fidergo-stephane-gourichon:
Turns out giving a title to the graph with an accent appeared to cause the problem.
Manually deleting the entry in carbon graph database allowed to see all previously saved graphs.

deniszh:
Ah, good to know, thanks!

fidergo-stephane-gourichon:
To see where is the problematic entry:

`sqlite3 /var/lib/graphite/graphite.db`

`select * from account_mygraph ;`

To delete it:

`delete from account_mygraph where ID=1`

fidergo-stephane-gourichon:
We now know the actual cause and an efficient workaround, yet the issue is still fully present.  Also, when given a title with an accent, another consequence is that the graph rendering feature is immediately broken, the graph image is replaced with a small broken icon in top left.

[...]

PR #1980, opened 2017-06-27
Warn against putting utf-8 characters that will break app.

fidergo-stephane-gourichon:
Very poor man's workaround for issue #1979.

Not sure it's worth to merge into master, I let you ponder.

Con: scares average us-ascii charset user.
Pro: saves users all over the world that would have broken their whole setup with only one accent.
Pro: hints savvy users/devs around the world to work on a real fix.

Cheers!

### DIFF
--- a/webapp/content/js/composer_widgets.js
+++ b/webapp/content/js/composer_widgets.js
@@ -1256,7 +1256,7 @@ function createOptionsMenu() {
 
   var yAxisLeftMenu = new Ext.menu.Menu({
     items: [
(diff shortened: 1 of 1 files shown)


brutasse:
I might be biased but I'd favor the fix in #1981 :)

fidergo-stephane-gourichon:
Bias? ;o) My PR has cleaner codecov-report ;-). (Just kidding, I guess PR #1980 has already reached the best it could do.) Thanks a lot for your effort!

PR #1981, opened 2017-06-27
Fix #1979 -- properly handle unicode dashboards in browser/usergraph

### LINKED ISSUE #1979: Saving a graph with an accent in title causes hiding all saved graphs, shows "no saved graphs" in tree "My Graphs".

brutasse:


DanCech:
What character set do we use for the database? That's typically the thing that trips utf-8 up.

brutasse:
@DanCech it depends on the backend being used, but for mysql/postgres it's not graphite-web's responsibility, and Django does its best to handle encoding properly: https://docs.djangoproject.com/en/1.11/ref/unicode/#creating-the-database.

Here the thing that trips utf-8 up is all the casting to `str()`, which are bytes in Python 2. `str(<some-unicode-string>)` uses ASCII encoding, so it works most of the time but breaks whenever the unicode string has characters out of the ASCII range.

Instead, if bytes are really needed, one should call `.encode('utf-8')` on unicode strings.

Python 2's bytes type is `str`, and its actual string type is `unicode`. Python 3 clears this confusion by having a `str` type which is proper unicode, and a `bytes` type. This means however that `str` has a different meaning across Python versions, and should only be used as last resort. Encoding strings and decoding bytes should be the norm.

PR #1982, opened 2017-06-27
Change dashboard find to only query for the name

cbowman0:
This improves speed of loading the dashboard finder window because it does not query and sort the state column where it doesn't need it.

### DIFF
--- a/webapp/graphite/dashboard/views.py
+++ b/webapp/graphite/dashboard/views.py
@@ -310,8 +310,8 @@ def find(request):
   results = []
 
   # Find all dashboard names that contain each of our query terms as a substring
-  for dashboard in Dashboard.objects.order_by('name'):
-    name = dashboard.name.lower()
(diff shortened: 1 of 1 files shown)


webapp/graphite/dashboard/views.py
```diff
@@ -310,8 +310,8 @@ def find(request):
   results = []
 
   # Find all dashboard names that contain each of our query terms as a substring
-  for dashboard in Dashboard.objects.order_by('name'):
-    name = dashboard.name.lower()
+  for dashboard in Dashboard.objects.values('name').order_by('name'):
```
brutasse: You could even do 

```python
for name in Dashboard.objects.order_by('name').values_list('name', flat=True):
    name = …
```

cbowman0: Ah.  Yes.  Still need to have a local variable with the lower() value for comparison.  I think it's cleaner to read, though, so I changed it to that.

PR #1987, opened 2017-07-05
is_local_interface and IPv6

DanCech:
This patch makes a couple of changes to the way that `is_local_interface` works, which allow it to work consistently across hosts with and without IPv6 interfaces.

- detect IPv6 based on the number of colons, an IPv4 address will always have 0 or 1
- test the given hostname for local-ness by attempting to bind to it as a source address rather than connecting to it and seeing whether the local (source) ip matched
- change behavior for `::1:8080` which is treated as a valid IP on hosts with IPv6 connectivity
- add tests for checking local-ness of dns names as well as ip addresses

### DIFF
--- a/webapp/graphite/util.py
+++ b/webapp/graphite/util.py
@@ -75,18 +75,19 @@ def timebounds(requestContext):
 
 def is_local_interface(host):
   is_ipv6 = False
-  if ':' in host:
-    try:
-      if host.find('[', 0, 2) != -1:
-        last_bracket_position  = host.rfind(']')
-        last_colon_position = host.rfind(':')
-        if last_colon_position > last_bracket_position:
-          host = host.rsplit(':', 1)[0]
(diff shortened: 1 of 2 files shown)


webapp/graphite/util.py
```diff
@@ -75,35 +75,32 @@ def timebounds(requestContext):
...
 
   try:
     if is_ipv6:
       sock = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
     else:
       sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
-    sock.connect( (host, 4242) )
-    local_ip = sock.getsockname()[0]
-    sock.close()
+    sock.bind( (host, 0) )
```
deniszh: Looks much better then `sock.connect( (host, 4242) )`
👍

Issue #1983, opened 2017-06-27
Errors with graphType=pie


cbowman0:
Can someone else confirm that setting `graphType=pie` on a render URL results in the following error?

```
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/graphite/webapp/graphite/render/views.py", line 377, in renderView
    image = doImageRender(requestOptions['graphClass'], graphOptions)
  File "/opt/graphite/webapp/graphite/render/views.py", line 630, in doImageRender
    img = graphClass(**graphOptions)
  File "/opt/graphite/webapp/graphite/render/glyph.py", line 577, in __init__
    self.drawGraph(**params)
  File "/opt/graphite/webapp/graphite/render/glyph.py", line 1803, in drawGraph
    self.drawLegend(elements)
  File "/opt/graphite/webapp/graphite/render/glyph.py", line 712, in drawLegend
    longestName = sorted([e[0] for e in elements],key=len)[-1]
IndexError: list index out of range
```

I'm testing out the current 1.0.x branch and noticed this error when attempting to load a dashboard with pie charts on it.  I don't think anything in my local changes or deployment *should* cause this.

cbowman0:
Graphs work if I set `REMOTE_PREFETCH_DATA = False`.

DanCech:
Sounds like an issue with the code that checks whether the remote data was prefetched or not

cbowman0:
It looks like the prefetch code is buried under the `  elif requestOptions['graphType'] == 'line':` section in webapp/graphite/render/views.py

DanCech:
Yes, but the code that does the actual fetches should see that it wasn't prefetched and fetch it normally, regardless of the prefetch setting

cbowman0:
I don't see where that would happen.  Should it be in datalib.py's `fetchData` or `_fetchData`?

cbowman0:
Making this change gets pie charts back to working: https://github.com/cbowman0/graphite-web/commit/1510d0afb587d5746876f032b5fcf4004fb64c78

PR #1997, opened 2017-07-11
Pie charts with remote prefetch

cbowman0:
Fix for #1983

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -78,7 +78,7 @@ def renderView(request):
 
   # Now we prepare the requested data
   if requestOptions['graphType'] == 'pie':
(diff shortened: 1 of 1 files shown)


webapp/graphite/render/views.py
```diff
@@ -78,6 +78,12 @@ def renderView(request):
 
   # Now we prepare the requested data
   if requestOptions['graphType'] == 'pie':
+    targets = [target for target in requestOptions['targets'] if target.find(':') < 0]
```
DanCech: Looking good, let's just move this inside the `if` to avoid doing the extra work if we don't need it.

cbowman0: Ok.  That makes sense.  Done.

PR #2007, opened 2017-07-28
fix divideByZero when FIND_CACHE_DURATION=0

### LINKED ISSUE #2006: RemoteStore.find fails when from or until is set and FIND_CACHE_DURATION=0
No exception is logged, but the issue is due to a "ZeroDivisionError: integer division or modulo by zero" in graphite.finders.remoate.FindRequest.__init__()

The exception is not bubbled up to the caller as the error occurs in a worker_pool.

https://github.com/graphite-project/graphite-web/blob/fec874a3e933419ee96ea18f5507a03c9fcccdec/webapp/graphite/finders/remote.py#L120-L128

woodsaj:
fixes #2006

### DIFF
--- a/webapp/graphite/finders/remote.py
+++ b/webapp/graphite/finders/remote.py
@@ -142,15 +142,15 @@ class FindRequest(object):
         self.store = store
         self.query = query
 
+        # prevent divide by 0
+        cacheTTL = settings.FIND_CACHE_DURATION or 1
         if query.startTime:
-            start = query.startTime - \
-                (query.startTime % settings.FIND_CACHE_DURATION)
+            start = query.startTime - (query.startTime % cacheTTL)
         else:
             start = ""
 
         if query.endTime:
-            end = query.endTime - (query.endTime %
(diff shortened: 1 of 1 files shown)


deniszh:
Cool, good catch!
Not sure about the fix, though. Now setting FIND_CACHE_DURATION=0 will move start and end time to -1 second. It's propably harmless, but not needed side effect IMO. @DanCech @iksaif ?

DanCech:
Looks good go me, anything mod 1 is going to be zero so it'll work well.On Jul 28, 2017 6:42 PM, Denis Zhdanov <notifications@github.com> wrote:Cool, good catch!
Not sure about the fix, though. Now setting FIND_CACHE_DURATION=0 will move start and end time to -1 second. It's propably harmless, but not needed side effect IMO. @DanCech @iksaif ?

deniszh:
Oh, stupid me,  mod 1 is 0, not 1 -  indeed! *puts on sackcloth and ashes*

deniszh:
Merging then, to hide my shame! :)

PR #2010, opened 2017-08-05
update metrics/find to support the same from/until specifications as render

DanCech:
This patch updates metrics/find to use the same parsing library for from/until dates as render does, to keep things consistent across the endpoints.

### DIFF
--- a/webapp/graphite/metrics/views.py
+++ b/webapp/graphite/metrics/views.py
@@ -13,8 +13,10 @@ See the License for the specific language governing permissions and
 limitations under the License."""
 import fnmatch
 import os
+import pytz
 import urllib
 
+from datetime import datetime
 from django.conf import settings
 from graphite.compat import HttpResponse, HttpResponseBadRequest
 from graphite.user_util import getProfile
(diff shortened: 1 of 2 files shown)


webapp/graphite/metrics/views.py
```diff
@@ -100,8 +104,29 @@ def find_view(request):
...
+    except pytz.UnknownTimeZoneError:
+      pass
+
+  if 'now' in queryParams:
+    now = parseATTime(queryParams['now'], tzinfo)
+  else:
+    now = datetime.now(tzinfo)
+
+  if 'from' in queryParams and str(queryParams['from']) != '-1':
+    fromTime = int(epoch(parseATTime(queryParams['from'], tzinfo, now)))
```
deniszh: How will it work if `from` is timestamp? Maybe we should make some shortcut for this ?

webapp/graphite/metrics/views.py
```diff
@@ -100,8 +104,29 @@ def find_view(request):
...
+  else:
+    fromTime = -1
+
+  if 'until' in queryParams and str(queryParams['from']) != '-1':
+    untilTime = int(epoch(parseATTime(queryParams['until'], tzinfo, now)))
```
deniszh: same for `until`

webapp/graphite/metrics/views.py
deniszh: Ah, nvm. Probably it's fine.

DanCech: Yep, `parseATTime` supports timestamps

PR #2011, opened 2017-08-07
Min-Max Normalization Function

msk610:
When processing data, various normalization and scaling techniques are utilized, and a very popular technique is Min-Max normalization. I have noticed that many people use this technique, especially when working with various machine learning algorithms, to process raw data, which is why I thought this function should be added as a graphite function. The formula is very simple, for each point do the following: (point - min) / (max - min). You can read more about it here: http://sebastianraschka.com/Articles/2014_about_feature_scaling.html#about-min-max-scaling

I have tested this new function on my own simple graphite server. If you want to see how it normalizes raw data live then look at: http://34.207.148.148/render/?target=minMax(collectd.ip-172-31-22-200_ec2_internal.entropy.entropy) and then look at http://34.207.148.148/render/?target=collectd.ip-172-31-22-200_ec2_internal.entropy.entropy to see how the raw data looks. This is a simple `Ubuntu 16.04.2` server I setup, which is running `Python 2.7.12`. It is also running `Django 1.8.7` and the Graphite version is `0.9.15`. Please let me know if I need to provide anything else for this function.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -4168,6 +4168,7 @@ SeriesFunctions = {
   'hitcount': hitcount,
   'absolute': absolute,
   'interpolate': interpolate,
+  'minMax': minMax,
 
   # Calculate functions
   'movingAverage': movingAverage,
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -4053,14 +4053,10 @@ def events(requestContext, *tags):
   """
   Returns the number of events at this point in time. Usable with
   drawAsInfinite.
-
   Example:
-
   .. code-block:: none
-
     &target=events("tag-one", "tag-two")
     &target=events("*")
-
   Returns all events tagged as "tag-one" and "tag-two" and the second one
   returns all events.
   """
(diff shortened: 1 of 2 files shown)


webapp/graphite/render/functions.py
```diff
@@ -4079,19 +4079,36 @@ def events(requestContext, *tags):
                                     epoch_to_dt(end_timestamp),
                                     tags=tags)
 
-  values = [None] * points
```
deniszh: Ahem.... Why do you kill half of `events()` function?

webapp/tests/test_functions.py
```diff
@@ -4650,6 +4650,29 @@ def mock_evaluateTokens(reqCtx, tokens, replacements=None):
 
         self.assertEqual(result, expectedResults)
 
+    # test_minMax
+    def test_minMax(self):
```
deniszh: Not sure if we want to test function like this. Please check other tests. Usually, you need to pick 2-3-5 etc predefined series, apply your function and compare results with right one.
Function below is just comparing your function with some reference implementation - but if we know that reference implementation is fine - why do we need to check that at all?

deniszh:
Thanks, @msk610, please check my comments. Thanks for your contribution!

msk610:
@deniszh thank you for the review, I am not sure why I removed part of the events function, I will add that back and fix my test function later today.

deniszh:
Now it still complaining about events function: 
```graphite-web/webapp/graphite/render/functions.py:docstring of graphite.render.functions.events:5: ERROR: Unexpected indentation.```
Please restore `events()` function as it was before - wil all spaces and empty lines - they're important.

deniszh:
Going to merge that too. More reviews, please?

iksaif:
LGTM

Issue #2023, opened 2017-09-05
An exception occurs when the metric path contains Unicode characters


zasca:
graphite-web 1.0.2 (0.9.15 not affected)

```
$ find /var/lib/graphite/whisper/app
/var/lib/graphite/whisper/app/
/var/lib/graphite/whisper/app/пример_рф
/var/lib/graphite/whisper/app/пример_рф/m1_rate.wsp
/var/lib/graphite/whisper/app/пример_рф/m15_rate.wsp

$ curl '127.0.0.1:8080/render/?from=-2hours&target=sumSeriesWithWildcards(app.*.m15_rate,%201)&format=json'
  ...
  File &quot;/opt/graphite/webapp/graphite/finders/standard.py&quot;, line 117, in _find_paths
    for match in self._find_paths(absolute_path, patterns):
  File &quot;/opt/graphite/webapp/graphite/finders/standard.py&quot;, line 83, in _find_paths
    entries = [x.name for x in scandir(current_dir)]
UnicodeDecodeError: &#39;ascii&#39; codec can&#39;t decode byte 0xd0 in position 30: ordinal not in range(128)
```

zasca:
patch
```
--- standard.py_orig    2017-08-31 15:09:46.000000000 +0300
+++ standard.py 2017-09-05 11:28:14.000000000 +0300
@@ -90,7 +90,7 @@
     if using_globstar:
         matching_subdirs = map(operator.itemgetter(0), walk(current_dir))
     else:
-        subdirs = [entry for entry in entries if isdir(join(current_dir, entry))]
+        subdirs = [entry for entry in entries if isdir(join(current_dir, str(entry)))]
         matching_subdirs = match_entries(subdirs, pattern)
 
     # if this is a terminal globstar, add a pattern for all files in subdirs
@@ -113,15 +113,15 @@
     if patterns: #we've still got more directories to traverse
       for subdir in matching_subdirs:
 
-        absolute_path = join(current_dir, subdir)
+        absolute_path = join(current_dir, str(subdir))
         for match in self._find_paths(absolute_path, patterns):
           yield match
 
     else: #we've got the last pattern
       if not has_wildcard:
         entries = [ pattern + '.wsp', pattern + '.wsp.gz', pattern + '.rrd' ]
-      files = [entry for entry in entries if isfile(join(current_dir, entry))]
+      files = [entry for entry in entries if isfile(join(current_dir, str(entry)))]
[...]

PR #2025, opened 2017-09-05
Handle metrics which contains Unicode in the path

zasca:
The issue #2023

### DIFF
--- a/webapp/graphite/finders/standard.py
+++ b/webapp/graphite/finders/standard.py
@@ -104,7 +104,7 @@ class StandardFinder(BaseFinder):
                 entry for entry in entries if isdir(
                     join(
                         current_dir,
-                        entry))]
+                        str(entry)))]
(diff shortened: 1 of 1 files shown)


deniszh:
@piotr1212 - could you please review it?

piotr1212:
Will do, I think I solved this specific issue in a different way.

deniszh:
Yes, that's what I remembered. Thanks!

DanCech:
Handling this properly is going to be a non-trivial exercise, especially while maintaining support for both python 2.x and 3.x.  Under 2.x this seems like it would still result in `UnicodeEncodeError` getting raised when it tries to convert unicode strings coming back from the filesystem into ascii.

brutasse:
Yes, `str()` calls will add more breakage in Python 2 and do nothing in Python 3…

Issue #2028, opened 2017-09-07
perSecond() returns OverflowError:  long int too large to convert to float


woodsaj:
When there are large sections of nulls in the data, perSecond() raises an "OverflowError:  long int too large to convert to float" exception.

the problem appears to be at
https://github.com/graphite-project/graphite-web/blob/1.0.2/webapp/graphite/render/functions.py#L1500

The step is being doubled on every consecutive null, when it should just be getting incremented by the original step.

PR #2029, opened 2017-09-07
fix perSecond null handling, use explicit integer division

DanCech:
As reported in #2028 there is an issue with how null values are handled in `perSecond`, it doubles the step value for every null encountered instead of incrementing it by the series step.  This has been corrected and the test suite has been updated to test the scenario where multiple null values are encountered.

While debugging this issue I saw that there are many places in `functions.py` including in `perSecond` where the code is reliant on the non-intuitive Python 2.3-2.7 behavior where the division operator `/` performs integer/floor division if both arguments are integers and regular division if one or more are floats.

See https://www.python.org/dev/peps/pep-0238/ for a discussion of division in Python and the changes introduced in 2.3 and in 3.0.

To both make it clearer in the code which type of division is being used and to bring us closer to 3.x compatibility, I added `from __future__ import division` and converted all instances of integer division in `functions.py` to use `//`.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -12,6 +12,10 @@
 #See the License for the specific language governing permissions and
 #limitations under the License.
 
+# make / work consistently between python 2.x and 3.x
+# https://www.python.org/dev/peps/pep-0238/
+from __future__ import division
+
 import math
 import random
 import re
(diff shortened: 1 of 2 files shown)


deniszh:
👍 LGTM
Thanks!

PR #2042, opened 2017-10-01
support "pipe" function chaining syntax

DanCech:
This PR adds support for chaining functions together by "piping" the output of each function into the next (as the first parameter), much like the way they're presented in the Grafana UI.

When dealing with deeply-nested queries this keeps the arguments for each function together, and can help increase readability vs the current "nested" syntax.

With this update the following query:
```
aliasByNode(movingAverage(sortByName(test.*),"5min"),1)
```
can be written as:
```
test.*|sortByName()|movingAverage("5min")|aliasByNode(1)
```
or:
```
sortByName(test.*)|movingAverage("5min")|aliasByNode(1)
```
Piping is supported within function arguments, so the same expression could also be written as
```
movingAverage(test.*|sortByName(),"5min")|aliasByNode(1)
```
A more useful example would be selecting series for the `total` parameter of `asPercent`:
```
test.*|sortByName()|asPercent(test.total|movingAverage("5min"))
```

The syntax is unwrapped in the parser and evaluator, so there is no difference in how the actual render functions are called, the evaluator walks through the piped calls from right to left, passing the chain up to that point into the function as the first parameter.

The one change I had to make that would affect existing syntax is that the pipe character `|` is added to the list of symbols that aren't allowed in pathExpressions, but I doubt that's widely used since it would make working with the whisper files painful.  If a user does have series with a pipe in the name they can still be selected but the pipe would need to be escaped with a backslash.

### DIFF
--- a/webapp/graphite/render/evaluator.py
+++ b/webapp/graphite/render/evaluator.py
@@ -14,7 +14,7 @@ def evaluateTarget(requestContext, target):
     return result
 
 
-def evaluateTokens(requestContext, tokens, replacements=None):
+def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
   if tokens.template:
     arglist = dict()
     if tokens.template.kwargs:
@@ -26,6 +26,10 @@ def evaluateTokens(requestContext, tokens, replacements=None):
     return evaluateTokens(requestContext, tokens.template, arglist)
 
   elif tokens.expression:
+    if tokens.expression.pipedCalls:
+      rightMost = tokens.expression.pipedCalls.pop()
+      return evaluateTokens(requestContext, rightMost, replacements, tokens)
+
     return evaluateTokens(requestContext, tokens.expression, replacements)
 
   elif tokens.pathExpression:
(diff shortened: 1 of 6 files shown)


DanCech:
Obviously this needs docs before it would be merge-able, I'd also like to write some tests for both the parser and evaluator since coverage is pretty weak there right now.  Feedback would be appreciated before I go ahead with that though.

iksaif:
That would certainly make it more readable.

webapp/graphite/render/evaluator.py
```diff
@@ -14,7 +14,7 @@ def evaluateTarget(requestContext, target):
     return result
 
 
-def evaluateTokens(requestContext, tokens, replacements=None):
+def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
```
deniszh: Could you please elaborate how pipedArg is passed here?

deniszh:
But in general, I'm very 👍 👍 👍 about that change, very nice addition to function syntax.
Let's properly test and document this first, though.

webapp/graphite/render/evaluator.py
DanCech: `pipedArg` is passed when evaluating the rightmost piped function, the value passed in will be the parse tree for the remaining part of the expression.  You can see on line 30 & 31 that the `rightMost` piped call is `pop()`ed off the stack and `tokens` is then passed as `pipedArg` when it's evaluated.  This mechanism is how the piped calls are "converted" back to a regular nested set of function calls.

deniszh: Ah, missed that, sorry. Thanks!

DanCech: No worries, it's definitely the most non-obvious part of the change.  I'll add some comments to explain what's going on there.

DanCech: Comments added

PR #2044, opened 2017-10-02
add aggregateWithWildcards function, minor refactor for movingWindow

DanCech:
In much the same way that `movingWindow` provides a centralized implementation for the various `moving*` functions and added support for additional aggregation methods, this PR defines a new `aggregateWithWildcards` function that provides a single implementation for all `*SeriesWithWildcards` functions and adds support for using all defined series aggregator function with wildcards (full list is in the docs).

It also includes a minor refactor for `movingWindow` to take the block that defines the aggregation func outside the loop.


DanCech:
@deniszh I tweaked things a little to make the function name less verbose and to sync the implementation with groupByNodes, which is very similar but accepts the nodes that match for each group rather than the nodes that are different between each group.  I'll update the docs so that each function refers to the other to make the relationship clearer.

PR #2047, opened 2017-10-04
TagDB refactor

DanCech:
This PR refactors the redis TagDB implementation to increase performance by evaluating a single "selector" expression and then treating any remaining expressions as filters on that set, rather than evaluating each expression separately and intersecting the results.  This makes a big difference in cases where previously a large number of values would have had to be read from redis and tested in Python (expressions using `!=`, `=~` or `!=~`).

It also refactors the local db implementation to use database joins only for expressions that don't match empty tags, and to use filtering to evaluate any remaining expressions.

It also adds an optimization in the prefetch code so that `extractPathExpressions` isn't called if there are no remote stores enabled.  In a future PR we'll want to add some caching to the TagDB lookups to avoid the duplicated TagDB lookups when using prefetch.

### DIFF
--- a/webapp/graphite/finders/remote.py
+++ b/webapp/graphite/finders/remote.py
@@ -24,7 +24,6 @@ from graphite.worker_pool.pool import get_pool, pool_apply
 
 class RemoteFinder(BaseFinder):
     local = False
-    disabled = False
 
     def __init__(self, hosts=None):
         if hosts is None:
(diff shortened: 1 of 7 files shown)


webapp/graphite/finders/remote.py
```diff
@@ -24,6 +24,7 @@
 
 class RemoteFinder(BaseFinder):
     local = False
+    disabled = False
```
iksaif: Maybe add disabled = False in BaseFinder ?

PR #2050, opened 2017-10-05
support matching series by node in asPercent

DanCech:
This PR adds support for specifying nodes to be used to match value and total series in calls to `asPercent` using the same logic as in `groupByNodes`.  This allows it to work much more reliably with glob expressions, and doesn't rely on the lists being the same length.  It also makes it much easier to calculate things like disk percentage free/used from raw values.

Usage description and examples are in the updated docblock.  I opted to add the extra params to `asPercent` rather than creating a separate `asPercentByNodes` function both to avoid further bloating the function list, and because I suspect that it will become the preferred way to use `asPercent` when dealing with multiple series.

In order to support the usage where nodes are specified without a `total` series, support has been added for passing `None` in function calls, which previously would have been passed through to the finders as a path expression.

### DIFF
--- a/webapp/graphite/render/evaluator.py
+++ b/webapp/graphite/render/evaluator.py
@@ -83,6 +83,9 @@ def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
   elif tokens.boolean:
     return tokens.boolean[0] == 'true'
 
+  elif tokens.none:
+    return None
+
   else:
     raise ValueError("unknown token in target evaluator")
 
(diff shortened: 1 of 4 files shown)


DanCech:
The latest commit changes the behavior when there are entries in `seriesList` that don't have a corresponding series in `total` and vice-versa.  Previously they were ignored, now they produce output series that have `MISSING` in the corresponding part of their name and values that are all `None`.  They can be filtered out if desired by piping through `|exclude('MISSING')`

deniszh:
@DanCech 
>The latest commit changes the behavior when there are entries in seriesList that don't have a corresponding series in total and vice-versa. Previously they were ignored, now they produce output series that have MISSING in the corresponding part of their name and values that are all None. They can be filtered out if desired by piping through |exclude('MISSING')

Nice feature! But maybe we should mention that in docs section?

DanCech:
Good call, I'll add it to the docs

PR #2051, opened 2017-10-06
Fix time series consolidation

mbethke:
A time series having a lot of None values would previously lead to O(n²) time complexity and thus very slow rendering when consolidating values.

This fix is mostly the work of @powo with two new consolidation functions, an added fix for the broken "average" consolidation function and deprecated exception syntax by myself.


DanCech:
Thanks @mbethke I have a version I'm going to publish shortly that includes updates for all the tests to include `consolidate(3)`

PR #2040, opened 2017-09-25
fix for issue 2012

leochen4891:
to fix https://github.com/graphite-project/graphite-web/issues/2012
remove file extension when getting `base_fs_path `
use `os.sep` to replace `'/'`
rename 
  `real_fs_path `to `real_aboslute_path`
  `relative_real_fs_path  `to `real_relative_fs_path `

PR #2052, opened 2017-10-06
Fix consolidation

DanCech:
This PR updates the code in #771 and incorporates updates from #2051 

In addition, it fixes an issue where consolidating a series by an interval that divides evenly into the number of points in the series results in a None value being appended to the result.

The new tests do highlight the need for support for `xFilesFactor` here, since the final value for some aggregations can be quite nonsensical when there aren't enough points in the interval.

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -107,7 +107,7 @@ class TimeSeries(list):
           yield None
     if buf:
       yield cf(buf)
-    else:
+    elif valcnt > 0:
       yield None
     raise StopIteration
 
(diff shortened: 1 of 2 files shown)


deniszh:
Change is good, but I think you meant #771 and not #2040 ?

DanCech:
@iksaif @cbowman0 @obfuscurity any input on the change to not always append `None` when the consolidation interval divides evenly into the number of raw points? I looked through the history and it seems that behavior has been around since the initial import by @cdavis but it does seem incorrect to me.

iksaif:
I'm honestly not sure what would be the implications of that. But it's worth trying, it can be reverted in a minor version if it causes too much trouble.

cbowman0:
I agree with @iksaif

DanCech:
Cool, I'll go ahead and merge this 👍 

I have what's turned into a decent-sized patch coming shortly that wires `xFilesFactor` support through all the remaining functions that aggregate across series or intervals, just needs a little more polish and some work on test coverage.

PR #2054, opened 2017-10-09
Render view refactor & maxDataPoints=1

DanCech:
This PR refactors `render/views.py` to break up the monolithic `renderView()` function into a set of helpers for each output format, and to unify handling of response caching and render logging.

It also expands the render test suite to increase coverage for the new helper functions to 100% and cover more of the available query parameters.

Finally, the second commit introduces a change to the consolidation behavior when `maxDataPoints=1` is specified.  Previously this path used the same "nudging" logic as other `maxDataPoints` values, but that would often produce nonsensical results because it would "nudge" out a large number of data points.  With this change there is no longer any "nudging" with `maxDataPoints=1` and all available points are simply consolidated.

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -63,23 +63,25 @@ def renderView(request):
     'localOnly' : requestOptions['localOnly'],
     'template' : requestOptions['template'],
     'tzinfo' : requestOptions['tzinfo'],
-    'forwardHeaders': extractForwardHeaders(request),
+    'forwardHeaders': requestOptions['forwardHeaders'],
     'data' : [],
     'prefetched' : {},
     'xFilesFactor' : requestOptions['xFilesFactor'],
   }
   data = requestContext['data']
 
+  response = None
+
   # First we check the request cache
   if useCache:
     requestKey = hashRequest(request)
-    cachedResponse = cache.get(requestKey)
-    if cachedResponse:
+    response = cache.get(requestKey)
+    if response:
       log.cache('Request-Cache hit [%s]' % requestKey)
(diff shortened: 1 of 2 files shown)


deniszh:
Could you please elaborate what is "maxDataPoints=1", and why it could be processed that way?
I think this should be included also in documentation - https://graphite.readthedocs.io/en/latest/render_api.html#maxdatapoints

DanCech:
@deniszh the most common use for setting `maxDataPoints=1` is the singlestat panel in Grafana, where doing that means that Graphite only needs to return a single data point to Grafana (instead of returning a list and having Grafana do the aggregation client-side, or hacking it in Graphite with `|summarize("10y")` or similar).

The problem with current master is that if you do that then typically you're going to throw out quite a lot of the points due to the "nudging", which makes the result more or less useless (and quite confusing to the end user).  If you look at the diff for c1752ff you can see that the tests show what happens to the values, the sum over the interval under the old scheme is quite a bit lower than you'd expect because most of the values are being thrown away.

I'm not really sure what we should put in the docs, to really explain it you have to first get into a discussion about how the "nudging" works to align the intervals when using `maxDataPoints`.  The way I look at it, with this patch the user is going to get back results that don't need to be explained, where current master you definitely need to understand what's going on "under the hood" to see why your value isn't what you'd expect.  Any suggestions?

deniszh:
OK, agreed. Does it really happens only for maxDataPoints=1 or just for low maxDataPoints?

DanCech:
The effect is there for other low `maxDataPoints`, but it's less pronounced and I can't think of any real-world scenarios besides 1 where you'd use less than 10 or so.

PR #2057, opened 2017-10-10
HTTP(S) TagDB

DanCech:
This PR adds a TagDB that forwards all tag-related operations to an HTTP(S) server that implements the graphite-web tagging API.

It's designed to be used as a standard way to integrate with outboard tag databases, or with clustered graphite configurations.

I'm not sure what the best way is to go about writing tests for this, since you really need to have 2 graphite installations running for it to work.  Any suggestions appreciated!

I also fixed a minor bug in the remote reader, which wasn't properly formatting the error log entry if the remote returned a non-200 status code.

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -94,6 +94,33 @@ As Whisper and other storage backends are designed to hold simple time-series da
 
   Tag support requires Graphite webapp & carbon version 1.1.0 or newer.
 
+Local Database TagDB
+^^^^^^^^^^^^^^^^^^^^
+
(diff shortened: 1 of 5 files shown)


deniszh:
👍 ,  but I'm wondering should we include some documentation for Redis tagdb or HTTP tagdb in local_settings.py.example or [documentation](https://graphite.readthedocs.io/en/latest/tags.html) maybe?

deniszh:
For testing - yes, we can mock HTTP tagdb separately, but it will be not very easy and not sure if have any sense...

DanCech:
Good call, I'll see about adding some detail to the docs.

DanCech:
I figured out how to test it, will check out docs tomorrow

PR #2064, opened 2017-10-12
Polish the valid datetime

hzy001:
The length of the valid datetime may not be 13, therefore we polish it.

Signed-off-by: Hao Ziyu <haoziyu@qiyi.com>

### DIFF
--- a/webapp/graphite/render/attime.py
+++ b/webapp/graphite/render/attime.py
@@ -37,7 +37,7 @@ def parseATTime(s, tzinfo=None, now=None):
       pass #Fall back because its not a timestamp, its YYYYMMDD form
     else:
       return datetime.fromtimestamp(int(s),tzinfo)
-  elif ':' in s and len(s) == 13:
+  elif ':' in s and s.replace(':','').isdigit():
     return tzinfo.localize(datetime.strptime(s,'%H:%M%Y%m%d'), daylight)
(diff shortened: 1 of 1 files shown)


deniszh:
Kicked stuck test.
Agreed, the previous check was quite naive.

DanCech:
I need to check, but I think this will break parsing for specifications like `8:50` when `now` is specified since that would now go through line 41 instead of line 82, and so with this patch would be relative to the current wall clock time rather than the `now` timestamp.

DanCech:
@hzy001 do you have any examples of time specifications you're seeing that aren't being parsed properly?

hzy001:
@DanCech 

First, I don't think it will break parsing for specifications like `8:50` when `now` is specified because the `s.replace(':','').isdigit()` will be False.

Second, I jusk try a set of examples and the parsing fails because the length of the string isn't 13, for example `8:5020171013` etc.

I just loose the restrict and I do think we need more advanced checks.

DanCech:
@hzy001 I'm talking about the url parameter `&now=` which is fed into `parseATTime` as the parameter `now`.  I'll take a look at adding tests to make that clear, and whether we can relax the restriction without breaking things.

DanCech:
Closing in favor of #2070

PR #2065, opened 2017-10-12
Add timeshift test coverage

cbowman0:


DanCech:
Any reason not to use mock?

cbowman0:
For `functions.evaluateTarget`?

DanCech:
Yeah, I see that there are a few places in `test_functions.py` with the manual patching, and other places that use `with patch('graphite.render.evaluator.fetchData', mock_data_fetcher):` for scenarios like this.

cbowman0:
Right.  I used the manual patching to keep with the other methods.  Mock makes more sense.  I'll rework this and the other cases that were similar to mock `fetchData`.

PR #2066, opened 2017-10-12
support sending logs to stderr

DanCech:
This PR adds 4 new config settings for the names of the 4 files that graphite-web logs to, and adds code to send these logs to `stderr` instead if those filenames are specified as `-`.

When sending logs to `stderr` the format is modified to make them easier to distinguish, and to support that it removes the (bogus) custom level names that were breaking formatting of warning messages, and updates the rendering and cache loggers to use `INFO` level.

### DIFF
--- a/webapp/graphite/settings.py
+++ b/webapp/graphite/settings.py
@@ -95,6 +95,11 @@ LOG_CACHE_PERFORMANCE = False
 LOG_ROTATION = True
 LOG_ROTATION_COUNT = 1
 
+LOG_FILE_INFO = 'info.log'
+LOG_FILE_EXCEPTION = 'exception.log'
+LOG_FILE_CACHE = 'cache.log'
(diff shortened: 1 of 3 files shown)


webapp/graphite/settings.py
```diff
@@ -95,6 +95,11 @@
 LOG_ROTATION = True
 LOG_ROTATION_COUNT = 1
 
+LOG_FILE_INFO = 'info.log'
```
iksaif: Maybe add a link to https://docs.djangoproject.com/en/1.11/topics/logging/#examples ?

deniszh: Does it applicable here?

iksaif: Yes, it's a logging/django thing, it always work :)

DanCech: It is linked on line 92

PR #2068, opened 2017-10-13
Add more test coverage for functions

cbowman0:
Tests for timeStack, useSeriesAbove, fallbackSeries, mostDeviant and stdev.

### DIFF
--- a/webapp/tests/test_functions.py
+++ b/webapp/tests/test_functions.py
@@ -2184,6 +2184,54 @@ class FunctionsTest(TestCase):
 
         self.assertEqual(results, [inputSeries[1], inputSeries[0]])
 
+    def test_stdev(self):
+        inputSeries = self._gen_series_list_with_data(
+            key=['test.foo'],
+            start=0,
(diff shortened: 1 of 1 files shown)


webapp/tests/test_functions.py
```diff
@@ -3457,7 +3686,6 @@ def mock_data_fetcher(reqCtx, path_expression):
 
 
         with patch('graphite.render.evaluator.fetchData', mock_data_fetcher):
-            # input values will be ignored and replaced by regression function
```
deniszh: Why removing this and comments below? Just curious

DanCech: Same question, looks great otherwise

cbowman0: They were erroneously copied from other checks when I created them this week.  I'll see if I can put comments there that do make sense.

cbowman0:
Some comments added back.

Also, unless I missed something, added tests for the final 2 functions with no coverage.

PR #2076, opened 2017-10-16
Use prefetch logic for sub-requests

DanCech:
When functions need to perform sub-requests to get data for a time range that's different from the main query (eg `moving*`, `timeShift`, etc) they currently always go through the regular find/read pipeline and can't use the faster prefetch mechanism to optimize those requests (instead potentially making a large number of small requests to the backend servers).

This PR moves the triggering of the prefetch mechanism from `views.py` into `evaluateTarget` and updates it to be able to accept pre-tokenized input, a raw string, or a list of tokens/strings.  All functions now use `evaluateTarget` instead of `evaluateTokens` so that they can take advantage of prefetching.

The special case here is `seriesByTag` which explicitly disables prefetching because the series will already have been prefetched by code in `extractPathExpression` that handles `seriesByTag` to be able to grab the matching series in the initial prefetch and not require a second prefetch cycle.

### DIFF
--- a/webapp/graphite/render/evaluator.py
+++ b/webapp/graphite/render/evaluator.py
@@ -1,18 +1,32 @@
 import re
+from time import time
+
+from django.conf import settings
+from graphite.logger import log
 from graphite.render.grammar import grammar
-from graphite.render.datalib import fetchData, TimeSeries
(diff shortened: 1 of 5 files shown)


deniszh:
Change are great, but codacy and lint are complaining for unused imports...

PR #2077, opened 2017-10-17
Fix example vhost filename

redNixon:
The filename for the example vhost configuration is "example-graphite-vhost.conf", it is documented correctly once and incorrectly referred to as  "example-graphite-vhosts.conf" once.  Not a huge issue, but it can cause confusion when users copy/paste this name when searching for the file on their filesystem(ie me).


deniszh:
Thanks!

PR #2078, opened 2017-10-17
Auto complete for tags

DanCech:
This PR implements a pair of endpoints to be used for providing autocomplete for tags and values based on series matching a given set of tag expressions.

```
# autocomplete list of tags starting with "tag" for series that match "tag1=value1"
/tags/autoComplete/tags?expr=tag1=value1&tagPrefix=tag

# autocomplete list of values starting with "value" for tag "tag2"
# for series that match "tag1=value1"
/tags/autoComplete/values?expr=tag1=value1&tag=tag2&valuePrefix=value
```

Both urls accept 0 or more `expr` or `expr[]` items and will return at most 100 matches.  `tagPrefix` and `valuePrefix` are both optional.

### DIFF
--- a/docs/tags.rst
+++ b/docs/tags.rst
@@ -148,10 +148,10 @@ To get a list of defined tags:
 
     [
       {
-        "tag": "name"
+        "tag": "datacenter"
       },
       {
-        "tag": "datacenter"
+        "tag": "name"
       },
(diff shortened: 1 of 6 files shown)


DanCech:
Getting some docs together now

PR #2079, opened 2017-10-17
Test class PrefetchedData

cbowman0:
Add tests for the render/datalib's PrefetchData class.


DanCech:
Looks good, just need to lose the unneeded import.  Any thoughts on #2076 ?

cbowman0:
Github was slow with processing the change removing that import.  https://status.github.com/messages shows queue delays.  I'll check out #2076 now.

PR #2081, opened 2017-10-18
whisper reader strange results for counters after merge_with_carbonlink

olevchyk:
fixes https://github.com/graphite-project/graphite-web/issues/2080

PR #2082, opened 2017-10-18
Aggregate points from carbon cache like whisper

DanCech:
This follows on from #2081 and simplifies the logic used to pre-aggregate the points from carbon cache before using the specified consolidation function.  It also includes a small optimization to use an array comprehension instead of `append` in a loop to build `consolidated`.

@olevchyk does this make sense to you?


olevchyk:
Yeah this looks much better.

PR #2088, opened 2017-10-23
use timezone-aware dates in hitcount, update epoch to handle naive dates

DanCech:
The `hitcount` function creates a `requestContext['startTime']` date that isn't timezone-aware, which can cause problems. This patch updates it to pass the `tzinfo` parameter like `smartSummarize` does.

The `epoch` function fails when passed a naive datetime (without timezone info), this patch updates it to interpret it as a date in the timezone specified in settings.

This patch also adds test for `epoch` and `epoch_to_dt`.

### DIFF
--- a/webapp/graphite/util.py
+++ b/webapp/graphite/util.py
@@ -57,6 +57,8 @@ def epoch(dt):
   """
   Returns the epoch timestamp of a timezone-aware datetime object.
   """
+  if not dt.tzinfo:
+    return calendar.timegm(make_aware(dt, pytz.timezone(settings.TIME_ZONE)).astimezone(pytz.utc).timetuple())
   return calendar.timegm(dt.astimezone(pytz.utc).timetuple())
 
 
--- a/webapp/graphite/util.py
+++ b/webapp/graphite/util.py
@@ -19,6 +19,7 @@ import time
 import sys
 import calendar
 import pytz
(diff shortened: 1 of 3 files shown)


webapp/graphite/util.py
```diff
@@ -57,6 +57,8 @@ def epoch(dt):
   """
   Returns the epoch timestamp of a timezone-aware datetime object.
   """
+  if not dt.tzinfo:
+    return calendar.timegm(make_aware(dt, pytz.timezone(settings.TIME_ZONE)).astimezone(pytz.utc).timetuple())
```
brutasse: IIRC this was done on purpose: an epoch is a fixed point in time, whereas a naive datetime can have multiple interpretations. I'm not sure using settings.TIME_ZONE as default fallback is always correct.

The absence of fallback is a way to raise issues when naive datetime are passed around. The fix is to make these datetimes aware (as you did 👍), but I'd keep epoch() fail rather than being at risk of being off by X hours.

DanCech: I understand what you're saying, but at the same time a hard failure is a pretty nasty thing.  I think that using the configured timezone is a pretty reasonable approach, and if we're worried about it then I can easily add a "warning non-timezone aware datetime passed to epoch" log entry there.

PR #2089, opened 2017-10-25
Added check for empty series list on percentileOfSeries

melnikk:
This closes issue with empty series which leads to unwanted exceptions when using this function.


deniszh:
That's much better, thank you!

PR #2093, opened 2017-10-27
Find/fetch/prefetch/threading refactor

DanCech:
This PR started out as an effort to be able to support surfacing errors from remote backends, but has evolved into a refactor of how the find/fetch cycle operates, removal of the remaining special handling for the remote finder and reader, and support for all finders to be able to find and fetch multiple patterns at once.  This is a pretty hefty change, and there are a few things that could probably be broken out into separate PRs if that is a concern.  I did take pains to improve test coverage as much as possible as I went.

The core change that opened up this line of thinking was the idea that if finders are made responsible for defining their own `factory()` class-method then it becomes possible for the remote finder to create a separate finder for each cluster host rather than trying to handle all cluster hosts together within a single finder.  That in turn makes it possible to drastically simplify the logic within both `RemoteFinder` and `RemoteReader` as now they only need to handle a single backend host and all thread handling can be moved into `STORE.find` and `STORE.fetch` (nee `STORE.remote_fetch`, see below).

The next change was to move the handling of the `find_nodes` / `node.fetch` logic down into `BaseFinder` in the `fetch` method, which in turn uses `find_multi` to get the list of matching nodes.  This means that `STORE.fetch` is able to dispatch a call to `fetch` on each finder (legacy finders that don't extend `BaseFinder` are monkey-patched to ensure backward compatibility) and let it handle either going through the standard `find_nodes`/`node.fetch` process or to redefine `fetch` as `RemoteFinder` does.

The next big change is that all fetches now use the prefetch path, so `evaluateTarget` dispatches `datalib.prefetchData` that in turn calls `STORE.fetch` and stores the returned data from all finders in `requestContext`.  Remote and local finders are treated just the same, `datalib` has no concept of finders at all, and `STORE` only uses the `local` flag on a finder is to honor the `local` request parameter.  `datalib.fetchData` just pulls the prefetched data for the given path expression from `requestContext` and merges the results from the various finders.  The `REMOTE_PREFETCH_DATA` setting was removed as it is no longer needed.  It is also worth noting that the `REMOTE_FIND_TIMEOUT` and `REMOTE_FETCH_TIMEOUT` settings now apply to all finds & fetches, we may want to consider adding `FIND_TIME_LIMIT` and `FETCH_TIME_LIMIT` or similar settings to control the total time that the find/fetch threads can run, and leave the `REMOTE_` settings to control only the timeouts on the individual HTTP calls.

Once all that was done and each finder was now operating in its own thread with no need to handle results from multiple backends etc or re-dispatching requests for data that was already being fetched there was no longer any need for the `FetchInProgress`, `wait_for_result` etc stuff, so all that was removed.

There are a number of other supporting refactors included to cut down on duplicated code and to fix bugs that I came across as I worked through things, I will go through and add comments in the patch to explain them in more detail.

As always, any and all feedback is appreciated!


tox.ini
```diff
@@ -17,9 +17,9 @@ passenv = TEST_MYSQL_* TEST_POSTGRESQL_*
 basepython = python2.7
 changedir = webapp
 commands =
-	coverage run --branch manage.py test
-	coverage xml --include=graphite/*
-	coverage report --include=graphite/*
+	coverage run --branch --include=graphite/* manage.py test
+	coverage xml
+	coverage report
 deps =
```
DanCech: This is just an improvement to the efficiency of the tests, because coverage isn't going to collect info on files outside of graphite only to discard it when generating the reports.

webapp/graphite/finders/remote.py
```diff
@@ -240,12 +114,45 @@ def send(self, headers=None, msg_setter=None):
...
                 node = BranchNode(path)
 
             node.local = False
             yield node
+
+    def fetch(self, patterns, start_time, end_time, now=None, requestContext=None):
+        reader = RemoteReader(self, {}, bulk_query=patterns)
+        return reader.fetch_multi(start_time, end_time, now, requestContext)
+
+    def request(self, url, fields=None, headers=None, timeout=None):
```
DanCech: This method centralizes the code used to make requests to remote hosts, and is also used by `RemoteReader`

webapp/graphite/readers/remote.py
```diff
@@ -1,222 +1,94 @@
...
-            data = []
-
-        log.debug("RemoteReader:: Completed _fetch %s" % url_full)
-        return data
-
-
-class _Results(object):
+              log.exception("Got an exception when fetching data! Try: %i of %i. Root cause:\n%s" %
+                           (retries, settings.MAX_FETCH_RETRIES, format_exc()))
+              retries += 1
```
DanCech: Retry logic for remote fetches was moved here, which allows each remote host to be retried individually.

webapp/graphite/render/evaluator.py
```diff
@@ -29,18 +34,19 @@ def evaluateTarget(requestContext, targets, noPrefetch=False):
...
+
 def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
   if tokens.template:
     arglist = dict()
     if tokens.template.kwargs:
-      arglist.update(dict([(kwarg.argname, evaluateTokens(requestContext, kwarg.args[0])) for kwarg in tokens.template.kwargs]))
+      arglist.update(dict([(kwarg.argname, evaluateScalarTokens(kwarg.args[0])) for kwarg in tokens.template.kwargs]))
     if tokens.template.args:
-      arglist.update(dict([(str(i+1), evaluateTokens(requestContext, arg)) for i, arg in enumerate(tokens.template.args)]))
+      arglist.update(dict([(str(i+1), evaluateScalarTokens(arg)) for i, arg in enumerate(tokens.template.args)]))
```
DanCech: This change ensures that template arg values are scalars, to be consistent with documentation and match a bugfix in `extractPathExpressions` which didn't handle templates properly.

webapp/graphite/render/evaluator.py
```diff
@@ -84,25 +90,80 @@ def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
...
-  elif tokens.none:
+  if tokens.none:
     return None
 
-  else:
-    raise ValueError("unknown token in target evaluator")
+  raise ValueError("unknown token in target evaluator")
+
+
+def extractPathExpressions(requestContext, targets):
```
DanCech: This function was moved here because it is so closely related to `evaluateTokens`

webapp/graphite/storage.py
```diff
@@ -14,58 +13,96 @@
...
+
+    if getattr(cls, 'factory', None):
+        return cls.factory()
+
+    # monkey patch so legacy finders will work
+    finder = cls()
+    finder.fetch = BaseFinder.fetch
+    finder.find_multi = BaseFinder.find_multi
+
+    return [finder]
```
DanCech: This change is what allows a finder module to create multiple finders, and is responsible for adding `fetch` and `find_multi` to legacy finders that don't extend `BaseFinder`.

webapp/graphite/util.py
```diff
@@ -220,48 +221,49 @@ def write_index(whisper_dir=None, ceres_dir=None, index=None):
...
+      timer.msg = 'failed in'
+      raise
+    finally:
+      timer.stop()
+
+  return wrapped_f
+
+
+@logtime
+def build_index(base_path, extension, fd, timer=None):
```
DanCech: This refactor simplifies use of `@logtime`, and makes it more extensible in future by adding functionality to the `Timer` class.

webapp/graphite/worker_pool/pool.py
```diff
@@ -1,58 +1,117 @@
...
+import time
 
 from threading import Lock
 from django.conf import settings
 from multiprocessing.pool import ThreadPool
 
-__init_lock = Lock()
-__pools = {}
+_init_lock = Lock()
+_pools = {}
```
DanCech: These variables are renamed to make them accessible for testing.

webapp/graphite/worker_pool/pool.py
```diff
@@ -1,58 +1,117 @@
...
+    self.args = args
+    self.kwargs = kwargs
+    self.result = None
+    self.exception = None
+
+  def run(self):
+    try:
+      self.result = self.func(*self.args, **self.kwargs)
+    except Exception as err:
+      self.exception = err
```
DanCech: The pool now accepts and returns `Job` objects, which makes it easier to keep track of what's going on, and to deal with functions that raise exceptions.

webapp/graphite/worker_pool/pool.py
```diff
@@ -1,58 +1,117 @@
...
   else:
     for job in jobs:
-      queue.put(job[0](*job[1:]))
-  return queue
+      job.run()
+
+      if time.time() > deadline:
+        raise PoolTimeoutError("Timed out after %fs" % (time.time() - start))
+
+      yield job
```
DanCech: `pool_exec` takes care of a few things that were previously duplicated in different places that used `pool_apply`.  It takes care of tracking timeouts (and adds timeout checking between jobs if the worker pool isn't used), it also allows a caller to just iterate over the result instead of having to deal with the queue directly.

webapp/tests/test_functions.py
```diff
@@ -2053,6 +2053,7 @@ def test_sortByMinima(self):
         result = functions.sortByMinima({}, seriesList)
         self.assertEqual(result, expectedResult)
 
+    @patch('graphite.render.evaluator.prefetchData', lambda *_: None)
```
DanCech: This patch is needed to avoid the real `prefetchData` being called when patching `fetchData` in the test.

webapp/tests/test_render_datalib.py
```diff
@@ -287,22 +255,9 @@ def _build_requestContext(self, startTime=datetime(1970, 1, 1, 0, 0, 0, 0, pytz.
...
-      self.assertFalse(nonempty(series))
-
-    def test_nonempty_false_nones(self):
-      series = TimeSeries("collectd.test-db.load.value", 0, 4, 1, [None, None, None, None])
-      self.assertFalse(nonempty(series))
-
     def test__merge_results(self):
       pathExpr = 'collectd.test-db.load.value'
-      startTime=datetime(1970, 1, 1, 0, 10, 0, 0, pytz.timezone(settings.TIME_ZONE)),
+      startTime=datetime(1970, 1, 1, 0, 10, 0, 0, pytz.timezone(settings.TIME_ZONE))
```
DanCech: This is a bugfix, previously startTime was getting defined as a tuple because of the trailing comma.  It didn't directly affect these test but became obvious in `test_prefetchData`

webapp/tests/test_tags.py
```diff
@@ -170,6 +170,23 @@ def test_local_tagdb(self):
...
+    self.assertEqual(cache_get.call_count, 1)
+    self.assertEqual(cache_get.call_args[0][0], 'TagDB.find_series:tag1=value1:tag2=value2')
+    self.assertEqual(result, [])
+
+    self.assertEqual(log_info.call_count, 1)
+    self.assertRegexpMatches(
+      log_info.call_args[0][0],
+      'graphite\.tags\.localdatabase\.LocalDatabaseTagDB\.find_series :: completed \(cached\) in [-.e0-9]+s'
+    )
+
```
DanCech: This test just verifies the TagDB behavior when there is a cached result for a query.

DanCech:
@iksaif I'd love to get your feedback on this

iksaif:
I'm out until next week. Will try to look on Monday.

iksaif:
This looks good. I'm mostly afraid about plugins compatibility here as it will be quite hard to support 1.0 and >1.1 with all these changes in the finder/reader API.

Do you think there would be a way to keep some of the symbols just for backward compatibility ? (Thinking about FetchInProgress and alike).

DanCech:
Hmm, that is an interesting question.  I'd really like to avoid maintaining that stuff moving forward, especially since it isn't the easiest thing in the world to follow.  Let's think about what that would look like, since we'd also need to re-add code in the store to run wait_for_results or similar.

deniszh:
Maybe we should release 1.1 before merging this and then merge it to master and give users some time to change 3rd party plugins?

DanCech:
Well, I doubt there are many (any besides biggraphite?) plugins using the 1.x "plugin api", so I'd rather get it released ASAP than have things drag on.  At this point anyone running 1.0.x should really upgrade to 1.1.x anyway, unless we want to backport some stuff and cut a 1.0.3

iksaif:
Ok, makes sense.

deniszh:
Well, all existing plugins are used "1.0.x" plugin api AKA "master" plugin api, because there's no other api exist, right? :)
I know at least 3 more-less actively used in a wild - https://github.com/InfluxGraph/influxgraph, https://github.com/brutasse/graphite-cyanite and https://github.com/yandex/graphouse/tree/master/src/main/pySources and a bunch of more-less outdated. But probably not a big deal, though.
I mostly worried about documentation here. This - https://graphite.readthedocs.io/en/latest/storage-backends.html - is mostly outdated now, right? Can we fix that before releasing this? 
And or we need some working custom finder there, probably `big-graphite` is enough.

iksaif:
I'll make biggraphite compatible during the week

DanCech:
@deniszh all that documentation is still correct, I didn't change compatibility with any 0.9.x-style plugins.  Looking at the code `influxgraph`, `graphite-cyanite` & `graphouse` will all work.  Right now there is no support for the `graphite-api`-style `fetch_multi` call so it won't take advantage of that for influxgraph or cyanite, but I think that it should be possible with a little refactoring in `BaseFinder`.  I will check that out and update the plugin API docs with the 1.1+ spec later this week.

DanCech:
Looking deeper the issue with trying to support `fetch_multi` is that it doesn't return which path expression each result is related to, which is required for the prefetch mechanism to work properly.  At this point I think the best way to move forward is to update the docs so people have a guide to implementing the more efficient `fetch` path, but use the code in this branch as-is.

deniszh:
>At this point I think the best way to move forward is to update the docs so people have a guide to implementing the more efficient fetch path, but use the code in this branch as-is.

If old plugins will work - then that's great, let's proceed! 👍 
We never had `fetch_multi` support anyway.
Thanks for digging that, @DanCech !

DanCech:
Cool, I'm going to go ahead with this merge so that I can open a PR for the index.json update, will open an issue for updating the docs to include the new capabilities that are currently only documented in the code of `BaseFinder` and `BaseReader`.

PR #2094, opened 2017-10-31
Fix name of variable from xFilesFactor to xfilesfactor

RudW0lf:
Proof of change.
I tried to find why the carbon don`t set xFilesFactor into my configuration:

[carbon]
pattern = ^carbon\.
xFilesFactor = 0.1
retentions = 1m:30d,15m:91d,1h:3y
aggregation-method = max

I use ceres as storage backed of graphite. I expect in .ceres_node something like that:
{"timeStep": 60, "retentions": [[60, 129600]], "xFilesFactor": 0.1, "aggregationMethod": "average"}
but xFilesFactor still have value 0.5 (value by-default)
I looked for the sources and found this expression:

# grep -R "xFilesFactor" lib64/
lib64/carbon/database.py:      options['xFilesFactor'] = options.pop('xfilesfactor')
lib64/carbon/database.py:      options['xFilesFactor'] = options.pop('xfilesfactor')

### DIFF
--- a/docs/config-carbon.rst
+++ b/docs/config-carbon.rst
@@ -99,7 +99,7 @@ Important notes before continuing:
 
 * This file is optional.  If it is not present, defaults will be used.
 * There is no ``retentions`` line.  Instead, there are ``xFilesFactor`` and/or ``aggregationMethod`` lines.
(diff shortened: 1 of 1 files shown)


DanCech:
`ConfigParser` converts all option names to lowercase, it looks like you may have an older version of the carbon library installed since in `master` `options.get('xfilesfactor')` only occurs once in `carbon/lib/storage.py`.

PR #2103, opened 2017-11-06
Move /metrics/index.json handling into finders & use worker pool

DanCech:
This PR moves all code used for generating `/metrics/index.json` into a new `get_index` function in `STORE` which uses the same worker pool approach as `find` and `fetch`.  It also provides a default implementation of `get_index` in `BaseFinder` so that all existing plugins will now expose their metrics automatically (the default implementation maps `get_index` to a `find_nodes` call with pattern `**`).  Finders are able to override `get_index` to provide their own more efficient implementation, which is what the `standard`, `ceres` & `remote` finders now do.

The only non-BC change introduced is to remove the old `cluster` parameter and behavior from `/metrics/index.json` and replace it with the same `local=1` behavior as `/metrics/find` and `/render`.  This means the following changes:

```
Old behavior:
/metrics/index.json           Returns an index of only local metrics
/metrics/index.json?cluster=1 Returns an index of only remote metrics

New behavior:
/metrics/index.json           Returns an index of all known (local & remote) metrics
/metrics/index.json?local=1   Returns an index of only local metrics
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ env:
   - TOXENV=py27-django18-pyparsing2
   - TOXENV=py27-django19-pyparsing2
   - TOXENV=py27-django110-pyparsing2
-  - TOXENV=py27-django111-pyparsing2
+  - TOXENV=py27-django111-pyparsing2-rrdtool
   - TOXENV=py27-django111-pyparsing2-mysql TEST_MYSQL_PASSWORD=graphite
   - TOXENV=py27-django111-pyparsing2-postgresql TEST_POSTGRESQL_PASSWORD=graphite
   - TOXENV=docs
@@ -16,6 +16,7 @@ addons:
   apt:
     packages:
       - libcairo2-dev
+      - librrd-dev
   postgresql: "9.5"
 
 services:
--- a/tox.ini
+++ b/tox.ini
@@ -1,7 +1,6 @@
 [tox]
 envlist =
-	py27-django1{8,9,10,11}-pyparsing2,
-	py27-django111-pyparsing2-{mysql,postgresql,rrdtool},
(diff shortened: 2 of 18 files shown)


DanCech:
I need to work on test coverage still, but this is working.

DanCech:
Ok, I think this is ready for review now 😃

DanCech:
I realized that with the changes it only makes sense to have a worker thread per finder because each python process will handle only a single request at a time, so we'll only ever dispatch a job for each finder at any given time.  Knowing that, I removed the `POOL_WORKERS` and `POOL_WORKERS_PER_BACKEND` settings, since they are no longer used.  I also removed the dependency of worker_pool on settings, the check for `USE_WORKER_POOL` is now handled in `STORE`.

The other big change is that when not using the worker pool we check the timeout value before scheduling a job, that way if the timeout is exceeded during execution of the last job it will still be returned rather than dropping all the completed work on the floor and returning an error.

iksaif:
Shouldn't we keep `POOL_WORKERS` to cap the number of threads ?

DanCech:
You mean to use it as a maximum if someone has a really large number of finders?

iksaif:
Yes (as one remote host = one finder)

Le 11 nov. 2017 10:19, "Dan Cech" <notifications@github.com> a écrit :

You mean to use it as a maximum if someone has a really large number of
finders?

—
You are receiving this because your review was requested.
Reply to this email directly, view it on GitHub
<https://github.com/graphite-project/graphite-web/pull/2103#issuecomment-343651762>,
or mute the thread

DanCech:
@iksaif that what you were thinking of?

iksaif:
👍

PR #2111, opened 2017-11-09
Correcting Calculate functions menu.

acl-oss:


deniszh:
Ah, good catch. Thanks!

PR #2112, opened 2017-11-10
tags: avoid dependencies in base.TagDB

iksaif:
This makes it easilly importable anywhere.
- auto-completion code moved to autocomplete.py
- caching code more to storage.py

Another option could be to keep a light interface
and have a less abstract class with a few pre-filled methods.

### DIFF
--- a/webapp/graphite/render/evaluator.py
+++ b/webapp/graphite/render/evaluator.py
@@ -147,7 +147,7 @@ def extractPathExpressions(requestContext, targets):
       # if we're prefetching seriesByTag, look up the matching series and prefetch those
       if tokens.call.funcname == 'seriesByTag':
         if STORE.tagdb:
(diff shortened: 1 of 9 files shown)


DanCech:
From IRC:
```
(9:25:55 AM) aussiedan: I'm not a huge fan of forcing people to import autocomplete which then imports settings anyway and forcing people to put that boilerplate into each tagdb
(9:26:39 AM) aussiedan: thinking it might make more sense to have the store provide settings and cache to the tagdb via __init__ when it sets it up
(9:27:25 AM) aussiedan: then we don't need to have the hoops for autocomplete and it having to go back out and import STORE to get the caching, etc
(9:28:05 AM) aussiedan: the tagdb just gets passed settings and cache and uses them without needing any imports
(9:28:57 AM) aussiedan: basically https://en.wikipedia.org/wiki/Dependency_injection
```

deniszh:
Should be closed in favor of #2121, right?

Issue #2113, opened 2017-11-10
Timestamps count differ from datapoints count.


xtus:
Hi,

Using version 1.1.0-dev I encounter an issue in a call to the `aggregate` function, and a successive call to `summarize`(in this case is then `_summarizeValues`).

The test case involve two TimeSeries, for instance:
(a) 'start': 1510324560, 'end': 1510328280, 'valuesPerPoint': 1, 'step': 120
(b) 'start': 1510324620, 'end': 1510328220, 'valuesPerPoint': 1, 'step': 30

In `aggregate`, before the call to `normalize`, the number of datapoints(i.e. `len(list(series)))` ) for the series is:
(a) 31
(b) 120

After the call to `normalize`, the datapoint count for the series is now:
(a) 31 (valuesPerPoint=1)
(b) 30 (valuesPerPoint=4)
and the (start, end, step) output from `normalize` is: (1510324560, 1510328280, 120)

At this point, the` len(values)` generated by the `aggregate` function is 30.

Considering the way that is used in `_summarizeValues` to calculate `timestamps` and `datapoints`, i.e.:

```
  timestamps = range( int(series.start), int(series.end), int(series.step) )
  datapoints = list(series)
```

The TimeSeries output of `aggregate` has:
```
len(timestamps) = 31
len(datapoints) = 30
```

And eventually, when trying to access  in `_summarizeValues` the 31st element in `datapoints` is leading to:
```
  File "/opt/graphite/webapp/graphite/render/functions.py", line 3998, in _summarizeValues
    if timestamps[i] >= timestamp_ and datapoints[i] is not None:
IndexError: list index out of range
```

The exception has the effect to have the WSGI thread to spin and allocate increasing amount of memory, until is killed.

I run the same test case on version 0.9.12, and what I get is that the datapoints count after the `normalize` is for both series as 31, so no exception triggered in this case.

Can you pinpoint the culprit? My guess is `normalize`.

Thanks

DanCech:
That is interesting, I'll have to see if anything has changed in `normalize` between the versions.  Not sure what would be causing the memory allocation, possibly the backtrace.

xtus:
Just to clarify, my guess it that `normalize` should output a (start, end, step) range that guarantee same amount of timestamps and datapoints between the series involved.

And I guess instead that in version 0.9.12 there was a bug with the `__consolidatingGenerator` in `TimeSeries`, that was returning an extra `None`.

[...]

PR #2114, opened 2017-11-10
Fix issues when combining series with different intervals

DanCech:
When combining series with different intervals using `normalize`, the number of resulting values may not always be equal.  In the past the consolidating generator worked around this problem by yielding an extra `None` value, but that wasn't a great way to solve the problem.

Now that we require Python 2.7, we can switch from using `izip` to `izip_longest` when combining values from multiple `TimeSeries`, which will pad any short series with `None` to avoid this problem when aggregating series with a different number of values.

In `_summarizeValues` I also made a change to avoid trying to walk past the end of the `datapoints` array if for some reason it still ended up shorter than expected.

Finally, I added some test code to validate functionality with series having differing steps, as the current tests only use series with `step=1`.

This should solve the issue in #2113

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -22,7 +22,7 @@ import re
 import time
 
 from datetime import datetime, timedelta
-from itertools import izip
+from itertools import izip, izip_longest
 from os import environ
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


deniszh:
Wow, cool stuff!

PR #2120, opened 2017-11-16
Fix 'Moving Max' composer widget

wridgers:
Erroneously uses `movingMin` function instead of `movingMax`.


deniszh:
Already fixed in https://github.com/graphite-project/graphite-web/pull/2111/

wridgers:
> Thanks, but already fixed in #2111

Ah. Nothing showed up when I search for it.

Oh well, glad it's fixed!

Issue #1406, opened 2015-11-24
Problem pulling metrics from graphite clusters (CLUSTER_SERVERS)


nicholaskuechler:
grahpite-web version: 0.9.12

I have graphite clusters in multiple datacenters (aka "regions"), and I'd like to present a unified view of all metrics in a "global" level graphite-web (aka "global").

Architecturally, it looks like this: ![http://i.imgur.com/OCbwnKC.png](http://i.imgur.com/OCbwnKC.png)

Config wise the regional graphite-web nodes look like this: 

``` python
# datacenter 1 (aka a "regional" cluster)
# there is a load balancer in front of the graphite-web nodes,
# with vip: 10.1.1.1
# regional graphite-web node local_settings.py:
CLUSTER_SERVERS = ["10.1.1.10:80", "10.1.1.20:80"]
```

And the "global" graphite-web node:

``` python
# "global" graphite-web local_settings.py, using the IP address of the
# load balancer vip for the "regional" clusters
CLUSTER_SERVERS = ['10.1.1.1:80', '10.2.2.1:80']
```

Note: I'm not running carbon on any of the graphite-web nodes so the CARBONLINK_HOSTS is commented out.

The regional cluster's graphite-web pulls metrics from carbon-cache fine, but the global graphite-web is unable to pull metrics. I see global graphite-web making connections to the cluster's graphite-web with this query: "GET /metrics/find/?local=1&format=pickle&query=infrastructure.%2A" but that query returns 0 results.

When I run that query against one of the carbon-cache nodes in the regional cluster, the query works fine.

It seems like the regional graphite-web nodes are not honoring CLUSTER_SERVERS for the query above.

I would expect the cluster-level graphite-web nodes to query the CLUSTER_SERVERS on query "GET /metrics/find/?local=1&format=pickle&query=whatever" but it appears to only be looking locally, and since there are no local metrics on the grahpite-web nodes, it's seeing 0 results.

deniszh:
That's true, when server makes queries to CLUSTER_SERVERS it will always add "local=1" to query, to make local requests only, for avoiding loops. AFAIK it's mandatory and can't be changed. 
But I'm still not really understand your setup and why you need "recursive" cluster requests...

bmhatfield:
[...]

PR #2123, opened 2017-11-22
Msgpack support

DanCech:
This PR adds support for msgpack output for render and find endpoints.  The data returned is almost identical to the pickled data, except that intervals are returned as arrays of `[start, end]` so that we don't require a custom msgpack extension.  msgpack output is enabled using `format=msgpack`.

The PR also add support for using msgpack encoding for intracluster render and find calls, and adds support for specifying per-host options when defining cluster hosts by appending them as query parameters, as well as for specifying the protocol to use (http or https).

Currently supported options are `format` (`pickle` (still the default) or `msgpack`) and `local` (`1` (default) or `0`).  For example, specifying a cluster host of `https://some.host?format=msgpack&local=0` will cause graphite-web to make requests to that host via https (overriding the `INTRACLUSTER_HTTPS`, which is still used in no protocol is specified), to use `msgpack` format, and to pass `local=0` in the requests to allow the hosts to forward the request to any remote finders it has defined as described in #1406 .

To avoid adding another hard dependency I have bundled the MIT-licensed https://github.com/vsergeev/u-msgpack-python library as `graphite.umsgpack`, which is used unless the user installs https://github.com/msgpack/msgpack-python  I also tidied up the imports for json, pickle & StringIO to just import them from `graphite.util` to make it easier to manage which ones are used (since we now require 2.7+ I think we can actually remove the try/except stuff around cPickle and cStringIO as they should always be available).

Finally, the intracluster requests now no longer buffer the entire response into a string before decoding it, instead streaming the response to the decoder to save some time and memory.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
   - TOXENV=py27-django19-pyparsing2
   - TOXENV=py27-django110-pyparsing2
   - TOXENV=py27-django111-pyparsing2-rrdtool
+  - TOXENV=py27-django111-pyparsing2-msgpack
   - TOXENV=py27-django111-pyparsing2-mysql TEST_MYSQL_PASSWORD=graphite
   - TOXENV=py27-django111-pyparsing2-postgresql TEST_POSTGRESQL_PASSWORD=graphite
   - TOXENV=docs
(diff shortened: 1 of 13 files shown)


DanCech:
14577ba also includes docs updates for the changes in #2093

PR #2124, opened 2017-11-22
set pathExpression on seriesByTag results

DanCech:
Currently the series returned by `seriesByTag` have `pathExpression` set to `group(<list of matching series>`, which can be very long and unwieldy.  This PR changes the behavior so that `pathExpression` on the returned series are set to match the `seriesByTag` call, as suggested by @shanson7.

It also includes an update to `getNodeOrTag` to save some work when using tags, and to properly return the last node in the name of a tagged series (currently it would include all the tags).

Finally, I fixed a typo where the `absolute()` function adds the wrong tag.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -158,9 +158,15 @@ def xff(nonNull, total, xFilesFactor=None):
 
 def getNodeOrTag(series, n, pathExpression=None):
   try:
-    return (pathExpression or series.name).split('.')[n]
-  except (IndexError, TypeError):
-    return series.tags.get(n, '')
+    # if n is an integer use it as a node, otherwise catch ValueError and treat it as a tag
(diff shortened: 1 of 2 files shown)


shanson7:
For something like `seriesByTag('name=~node.*.node3')` would `_getFirstPathExpression` return `None`? If so, this looks good to me.

DanCech:
Yes, it would.

webapp/graphite/render/functions.py
```diff
@@ -1281,7 +1287,7 @@ def absolute(requestContext, seriesList):
     &target=absolute(Server.instance*.threads.busy)
   """
   for series in seriesList:
-    series.tags['squareRoot'] = 1
+    series.tags['absolute'] = 1
```
deniszh: nice catch

PR #2128, opened 2017-11-28
Tag federation & py3 fixes

DanCech:
This PR implements support for distributed tag databases in a clustered graphite scenario, forwarding `seriesByTag` expressions to cluster hosts that support tags.  This will allow tagged graphite to scale out across a cluster horizontally the same way that non-tagged graphite does.

It is currently functional, but needs work on test coverage and cleaning up any rough edges.

The basic concept is that `seriesByTag` handling is moved into the `STORE.fetch()` method, and it can look up per-finder whether the finder supports tags and if so send the `seriesByTag()` expression through to the finder.  For finders that don't support tags the TagDB is used to look up the matching series and the resulting series are fetched.

### DIFF
--- a/webapp/graphite/storage.py
+++ b/webapp/graphite/storage.py
@@ -405,7 +405,7 @@ class Store(object):
 
         results = set()
 
-        # if we're using the local tagdb then execute it
+        # if we're using the local tagdb then execute it (in the main thread so that LocalDatabaseTagDB will work)
         if use_tagdb:
           results.update(self.tagdb.auto_complete_tags(exprs, tagPrefix=tagPrefix, limit=limit, requestContext=requestContext))
 
@@ -420,16 +420,17 @@ class Store(object):
 
             if job.exception:
               errors += 1
-              raise Exception('%s %s %s %s' % (job.func, job.args, job.kwargs, job.exception))
               log.info("Autocomplete tags for %s %s failed after %fs: %s" % (str(exprs), tagPrefix or '', time.time() - start, str(job.exception)))
               continue
 
             log.debug("Got an autocomplete result for %s %s after %fs" % (str(exprs), tagPrefix or '', time.time() - start))
             results.update(job.result)
         except PoolTimeoutError:
-          log.info("Timed out in autocomplete tags after %fs" % (time.time() - start))
+          raise Exception("Timed out in autocomplete tags for %s %s after %fs" % (str(exprs), tagPrefix or '', time.time() - start))
 
         if errors == done:
+          if errors == 1:
+            raise Exception("Autocomplete tags for %s %s failed: %s" % (str(exprs), tagPrefix or '', str(job.exception)))
           raise Exception('All autocomplete tag requests failed for %s %s' % (str(exprs), tagPrefix or ''))
 
         # sort & limit results
@@ -466,7 +467,7 @@ class Store(object):
 
         results = set()
 
-        # if we're using the local tagdb then execute it
+        # if we're using the local tagdb then execute it (in the main thread so that LocalDatabaseTagDB will work)
         if use_tagdb:
           results.update(self.tagdb.auto_complete_values(exprs, tag, valuePrefix=valuePrefix, limit=limit, requestContext=requestContext))
 
@@ -487,9 +488,11 @@ class Store(object):
             log.debug("Got an autocomplete result for %s %s %s after %fs" % (str(exprs), tag, valuePrefix or '', time.time() - start))
             results.update(job.result)
         except PoolTimeoutError:
-          log.info("Timed out in autocomplete values after %fs" % (time.time() - start))
+          raise Exception("Timed out in autocomplete values for %s %s %s after %fs" % (str(exprs), tag, valuePrefix or '', time.time() - start))
 
         if errors == done:
+          if errors == 1:
+            raise Exception("Autocomplete values for %s %s %s failed: %s" % (str(exprs), tag, valuePrefix or '', str(job.exception)))
           raise Exception('All autocomplete value requests failed for %s %s %s' % (str(exprs), tag, valuePrefix or ''))
 
         # sort & limit results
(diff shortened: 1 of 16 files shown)


webapp/graphite/storage.py
```diff
@@ -331,6 +371,127 @@ def distance_to_requested_interval(node):
...
+        # Start fetches
+        start = time.time()
+        try:
+          for job in jobs:
+            done += 1
+
+            if job.exception:
+              errors += 1
+              raise Exception('%s %s %s %s' % (job.func, job.args, job.kwargs, job.exception))
+              log.info("Autocomplete tags for %s %s failed after %fs: %s" % (str(exprs), tagPrefix or '', time.time() - start, str(job.exception)))
```
deniszh: should it be before raise Exception?

DanCech: That is some testing code that got through, will clean it up when I get the auto-complete tests fleshed out.

webapp/graphite/storage.py
```diff
@@ -331,6 +371,127 @@ def distance_to_requested_interval(node):
...
+              continue
+
+            log.debug("Got an autocomplete result for %s %s after %fs" % (str(exprs), tagPrefix or '', time.time() - start))
+            results.update(job.result)
+        except PoolTimeoutError:
+          log.info("Timed out in autocomplete tags after %fs" % (time.time() - start))
+
+        if errors == done:
+          raise Exception('All autocomplete tag requests failed for %s %s' % (str(exprs), tagPrefix or ''))
```
deniszh: should we raise exception here? It will not be caught, right?

webapp/graphite/storage.py
```diff
@@ -331,6 +371,127 @@ def distance_to_requested_interval(node):
...
+              log.info("Autocomplete values for %s %s %s failed after %fs: %s" % (str(exprs), tag, valuePrefix or '', time.time() - start, str(job.exception)))
+              continue
+
+            log.debug("Got an autocomplete result for %s %s %s after %fs" % (str(exprs), tag, valuePrefix or '', time.time() - start))
+            results.update(job.result)
+        except PoolTimeoutError:
+          log.info("Timed out in autocomplete values after %fs" % (time.time() - start))
+
+        if errors == done:
+          raise Exception('All autocomplete value requests failed for %s %s %s' % (str(exprs), tag, valuePrefix or ''))
```
deniszh: Same here

DanCech: It would get caught in tags/views.py and output as a json error

DanCech: I have a few changes for the auto-complete stuff as I'm working on the tests for it, right now I'd really appreciate feedback on the render pipeline changes.

deniszh: Pipeline changes look good. 👍

webapp/graphite/storage.py
```diff
@@ -77,18 +79,42 @@ def pool_exec(self, jobs, timeout):
...
-        jobs = [
-            Job(finder.fetch, patterns, startTime, endTime, now=now, requestContext=requestContext)
-            for finder in self.get_finders(requestContext.get('localOnly'))
-        ]
+        jobs = []
+        tag_patterns = None
+        pattern_aliases = defaultdict(list)
+        for finder in self.get_finders(requestContext.get('localOnly')):
+          if getattr(finder, 'tags', False):
+            job_patterns = patterns
```
iksaif: maybe this could be a continue here to avoid the else going to much to the right ?

webapp/graphite/storage.py
```diff
@@ -77,18 +79,42 @@ def pool_exec(self, jobs, timeout):
...
+          else:
+            if tag_patterns is None:
+              tag_patterns = []
+              for pattern in patterns:
+                if not pattern.startswith('seriesByTag('):
```
iksaif: many levels of indentation here, any luck this could go in a function ?

webapp/graphite/storage.py
```diff
@@ -331,6 +371,127 @@ def distance_to_requested_interval(node):
...
 
+    def tagdb_auto_complete_tags(self, exprs, tagPrefix=None, limit=None, requestContext=None):
+        log.debug(
+            'graphite.storage.Store.auto_complete_tags :: Starting lookup on all backends')
+
+        if requestContext is None:
+          requestContext = {}
+
+        jobs = []
+        use_tagdb = False
```
iksaif: part of this code seems duplicated between tags and value

DanCech: They're broadly similar, but there are a lot of small differences (log messages, job parameters).  I could try to abstract it away into a helper but I'm not sure how complex it would get trying to handle the differences.

webapp/graphite/util.py
```diff
@@ -36,7 +36,18 @@
...
+
+  class StringIO(BytesIO):
+    def __init__(self, buffer=None):
+      if buffer is None:
+        super().__init__()
+      elif isinstance(buffer, six.string_types):
+        super().__init__(buffer.encode('utf-8'))
+      else:
+        super().__init__(buffer)
+
```
DanCech: @takluyver I came up with this as an alternative to having to worry about passing strings to StringIO, does it make sense to you?

takluyver: Yes-ish. I think it will work, but:

- It may be surprising that constructing the object accepts either str or bytes, but its `.write()` method only accepts bytes.
- The naming is not ideal from a Py3 perspective - I expect a StringIO object to be accepting and returning str data, not bytes. This is my fault from the porting.
- In the longer term, I'd see it as a stopgap solution. It's best for application code to be clear about where it's handling unicode and where it's handling bytes, and minimise the code paths that can be either.

DanCech:
Ok, I think I've hit all the outstanding issues here.  Much happier now that we're using BytesIO consistently.

DanCech:
@deniszh @iksaif care to give this a final once-over?

iksaif:
:+1:

deniszh:
LGTM!

PR #2133, opened 2017-11-30
use a buffered reader for http responses in remote finder/reader

woodsaj:
- pickle.load() makes lots of very small reads.  When operating on
  a http response object this is really slow.  To improve performance
  we need to read from the response objects in chunks.

With internal testing a query loading 1.8Mb of pickle data was taking 45seconds.  It now takes 300ms.


deniszh:
Good idea, but it expects [urllib3.connectionpool.HTTPConnectionPool()](https://urllib3.readthedocs.io/en/1.5/pools.html#urllib3.connectionpool.HTTPConnectionPool) in `result`.

DanCech:
I'm working on an updated patch, will open a PR shortly

Issue #750, opened 2014-06-24
Add Python 3 support


chhantyal:

mscuthbert:
My guess is that's what the OP is suggesting (well, probably a hybrid, .six-based version).  After `supervisor`, carbon/graphite is now the second most used Python package not to support Py3 [and two of the next three either have a perfect Py2/3 drop-in replacement (PyMYSQL for MySQL-Python) or is a backport of a Py3 library].  It would be a wonderful thing to be able to use in Py3 if any of the developers are interested.

stanjer:
Are there any problems with porting it to python3 (except work needs to be done by someone)?

deniszh:
@stanjer : "except the work" - no, I don't think so.
But please note that we should stay to be compatible with 2.7 too,  so, it should be dual compatible.

takluyver:
If anyone is looking to work on this, [modernize](https://pypi.python.org/pypi/modernize) and [future](https://pypi.python.org/pypi/future) are two tools that can automate a lot of the porting work - like 2to3, but they aim to keep compatibility with Python 2 as well.

Modernize is generally the more conservative of the two, and it will insert use of the well-known [six](http://six.readthedocs.io/) compatibility library. Future's `futurize` tool makes the code more like idiomatic Python 3, using its companion `future` library to make it work on Python 2.

piotr1212:
@takluyver While I was working on porting carbon https://github.com/graphite-project/carbon/pull/656 it turned out a lot more complicated than to just run some automated tools. Issues which you'll run into: str<->unicode mess, functions returning different types, missing testcases, imported libraries which are not py3 compatible yet.

takluyver:
Yup, the tools can help with the straightforward syntax changes, like changing `print x` to `print(x)`, but figuring out the changes to the data model still needs human work. I'm all too familiar with those issues ;-). But the job is much more tractable with the tools to help than doing it all manually.

takluyver:
Attempted in #2139 .

cclauss:
Is this issue closed?

deniszh:
I think yes.

codeasashu:
What the heck? Why is the issue closed?

deniszh:
Because Graphite in general currently Python 3 compatible. If you have some
specific issue - please open specific issue.

PR #2139, opened 2017-12-02
Python 3 support

takluyver:
See #750. This adds Python 3 support while preserving Python 2.7 support. It adds the popular `six` compatibility library as a dependency.

Testing locally, I see one remaining test failure that's specific to Python 3. I think it's because the code under test is using `os.walk()`, and on Python 3 that uses `scandir()` rather than `listdir()`. But I'm not 100% confident, and I don't understand why the test is counting calls to `listdir()`, so I haven't changed it yet.

```
FAIL: test_ceres_finder (tests.test_finders.CeresFinderTest)
  File "/home/takluyver/Code/graphite-web/webapp/tests/test_finders.py", line 359, in test_ceres_finder
    self.assertEqual(self._listdir_counter, 8)
AssertionError: 2 != 8
```

There are also a couple of other failures that I see under Python 2 and 3, but I assume they are issues with my system. They were there before I started changing anything.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,8 @@ try:
           'Programming Language :: Python',
           'Programming Language :: Python :: 2',
           'Programming Language :: Python :: 2.7',
-          'Programming Language :: Python :: 2 :: Only',
+          'Programming Language :: Python :: 3',
+          'Programming Language :: Python :: 3.6',
           ],
       **setup_kwargs
     )
--- a/setup.py
+++ b/setup.py
@@ -116,6 +116,8 @@ try:
           'Programming Language :: Python :: 2',
           'Programming Language :: Python :: 2.7',
           'Programming Language :: Python :: 3',
+          'Programming Language :: Python :: 3.4',
+          'Programming Language :: Python :: 3.5',
           'Programming Language :: Python :: 3.6',
           ],
       **setup_kwargs
--- a/setup.py
+++ b/setup.py
@@ -11,6 +11,8 @@ except ImportError:
 from glob import glob
 from collections import defaultdict
 
+# io.StringIO is strictly unicode only. Python 2 StringIO.StringIO accepts
+# bytes, so we'll conveniently ignore decoding and reencoding the file there.
 try:
     from StringIO import StringIO  # Python 2
 except ImportError:
--- a/setup.py
+++ b/setup.py
@@ -3,15 +3,18 @@
 from __future__ import with_statement
 
 import os
-import ConfigParser
+try:
+    from ConfigParser import ConfigParser, DuplicateSectionError  # Python 2
+except ImportError:
+    from configparser import ConfigParser, DuplicateSectionError  # Python 3
 
 from glob import glob
 from collections import defaultdict
 
 try:
-    from io import BytesIO
+    from StringIO import StringIO  # Python 2
 except ImportError:
-    from StringIO import StringIO as BytesIO
+    from io import StringIO  # Python 3
 
 # Graphite historically has an install prefix set in setup.cfg. Being in a
 # configuration file, it's not easy to override it or unset it (for installing
@@ -24,22 +27,22 @@ except ImportError:
 # ``setup.cfg``.
 with open('setup.cfg', 'r') as f:
     orig_setup_cfg = f.read()
-cf = ConfigParser.ConfigParser()
-cf.readfp(BytesIO(orig_setup_cfg), 'setup.cfg')
+cf = ConfigParser()
+cf.readfp(StringIO(orig_setup_cfg), 'setup.cfg')
 
 if os.environ.get('GRAPHITE_NO_PREFIX') or os.environ.get('READTHEDOCS'):
     cf.remove_section('install')
 else:
     try:
         cf.add_section('install')
-    except ConfigParser.DuplicateSectionError:
+    except DuplicateSectionError:
         pass
     if not cf.has_option('install', 'prefix'):
         cf.set('install', 'prefix', '/opt/graphite')
     if not cf.has_option('install', 'install-lib'):
         cf.set('install', 'install-lib', '%(prefix)s/webapp')
 
-with open('setup.cfg', 'wb') as f:
+with open('setup.cfg', 'w') as f:
     cf.write(f)
 
 if os.environ.get('USE_SETUPTOOLS'):
@@ -101,7 +104,7 @@ try:
       package_data={'graphite' :
         ['templates/*', 'local_settings.py.example']},
       scripts=glob('bin/*'),
-      data_files=webapp_content.items() + storage_dirs + conf_files + examples,
+      data_files=list(webapp_content.items()) + storage_dirs + conf_files + examples,
       install_requires=['Django>=1.8,<1.11.99', 'django-tagging==0.4.3', 'pytz', 'pyparsing', 'cairocffi', 'urllib3', 'scandir', 'six'],
       classifiers=[
           'Intended Audience :: Developers',
--- a/webapp/graphite/dashboard/views.py
+++ b/webapp/graphite/dashboard/views.py
@@ -1,4 +1,3 @@
-from __future__ import absolute_import
 import json
 import re
 import errno
@@ -15,7 +14,7 @@ from django.utils.safestring import mark_safe
 from graphite.compat import HttpResponse
 from graphite.dashboard.models import Dashboard, Template
 from graphite.render.views import renderView
-from .send_graph import send_graph_email
+from graphite.dashboard.send_graph import send_graph_email
 
 
 fieldRegex = re.compile(r'<([^>]+)>')
--- a/webapp/graphite/dashboard/views.py
+++ b/webapp/graphite/dashboard/views.py
@@ -1,4 +1,3 @@
-import json
 import re
 import errno
 
@@ -13,9 +12,9 @@ from django.contrib.staticfiles import finders
 from django.utils.safestring import mark_safe
 from graphite.compat import HttpResponse
 from graphite.dashboard.models import Dashboard, Template
-from graphite.render.views import renderView
 from graphite.dashboard.send_graph import send_graph_email
-
+from graphite.render.views import renderView
+from graphite.util import json
 
 fieldRegex = re.compile(r'<([^>]+)>')
 defaultScheme = {
--- a/webapp/graphite/node.py
+++ b/webapp/graphite/node.py
@@ -18,7 +18,7 @@ class BranchNode(Node):
 
 
 class LeafNode(Node):
-  __slots__ = ('reader', 'is_leaf')
+  __slots__ = ('reader', )
 
   def __init__(self, path, reader):
     Node.__init__(self, path)
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -13,8 +13,16 @@ See the License for the specific language governing permissions and
 limitations under the License."""
 
 import math, itertools, re, sys
-from six.moves import range
-from six.moves import zip
+from datetime import datetime, timedelta
+from six.moves import range, zip
+from six.moves.urllib.parse import unquote_plus
+from six.moves.configparser import SafeConfigParser
+from django.conf import settings
+import pytz
+
+from graphite.render.datalib import TimeSeries
+from graphite.util import json
+
 try:
     import cairocffi as cairo
 except ImportError:
@@ -29,14 +37,7 @@ else:
     from cStringIO import StringIO
   except ImportError:
     from StringIO import StringIO
-from datetime import datetime, timedelta
-from six.moves.urllib.parse import unquote_plus
-from six.moves.configparser import SafeConfigParser
-from django.conf import settings
-from graphite.render.datalib import TimeSeries
-from graphite.util import json
 
-import pytz
 
 INFINITY = float('inf')
 
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -1049,7 +1049,7 @@ class LineGraph(Graph):
 
     for series in self.data:
       if not hasattr(series, 'color'):
-        series.color = self.colors.next()
+        series.color = next(self.colors)
 
     titleSize = self.defaultFontParams['size'] + math.floor( math.log(self.defaultFontParams['size']) )
     self.setFont( size=titleSize )
@@ -1802,7 +1802,7 @@ class PieGraph(Graph):
         'name' : name,
         'value' : value,
         'percent' : value / self.total,
-        'color' : self.colors.next(),
+        'color' : next(self.colors),
         'alpha' : self.alpha,
       })
 
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -13,6 +13,7 @@ See the License for the specific language governing permissions and
 limitations under the License."""
 
 import math, itertools, re, sys
+from six.moves import range
 try:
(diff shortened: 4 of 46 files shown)


DanCech:
Very cool! Will check this out

takluyver:
With a few tweaks to the redis interface, Travis now agrees with me that the only remaining failure is the `_listdir_counter` one I posted above.

deniszh:
I made a quick hack to test python3 in docker -  https://github.com/graphite-project/docker-graphite-statsd/tree/python3
It mostly works, after fixing some dependencies:
```root@9aca3511819c:/# ps awx | grep python3
    1 ?        Ss     0:00 /usr/bin/python3 -u /sbin/my_init
   26 ?        S      0:00 /usr/bin/python3 /usr/local/bin/gunicorn wsgi --workers=4 --bind=127.0.0.1:8080 --log-file=/var/log/gunicorn.log --preload --pythonpath=/opt/graphite/webapp/graphite
   31 ?        S      0:00 /usr/bin/python3 /opt/graphite/bin/carbon-aggregator.py start --debug
   34 ?        Sl     0:00 /usr/bin/python3 /opt/graphite/bin/carbon-cache.py start --debug
   61 ?        S      0:00 /usr/bin/python3 /usr/local/bin/gunicorn wsgi --workers=4 --bind=127.0.0.1:8080 --log-file=/var/log/gunicorn.log --preload --pythonpath=/opt/graphite/webapp/graphite
   62 ?        S      0:00 /usr/bin/python3 /usr/local/bin/gunicorn wsgi --workers=4 --bind=127.0.0.1:8080 --log-file=/var/log/gunicorn.log --preload --pythonpath=/opt/graphite/webapp/graphite
   63 ?        S      0:00 /usr/bin/python3 /usr/local/bin/gunicorn wsgi --workers=4 --bind=127.0.0.1:8080 --log-file=/var/log/gunicorn.log --preload --pythonpath=/opt/graphite/webapp/graphite
   64 ?        S      0:00 /usr/bin/python3 /usr/local/bin/gunicorn wsgi --workers=4 --bind=127.0.0.1:8080 --log-file=/var/log/gunicorn.log --preload --pythonpath=/opt/graphite/webapp/graphite```
So, will try to find out why ceres test failing. I'm afraid ceres is not python3 ready, probably we should exclude it from tests then.

**Comment by takluyver** [comment]
I think [ceres is compatible](https://github.com/graphite-project/ceres/blob/0804b6e15857aec461aec76b365bd94c40e30fae/setup.py#L19-L27). The test is making an assumption about the inner workings of ceres - specifically, how it looks at directories - that doesn't appear to hold on Py3. It should be easy to fix the test, but I don't understand why it's checking that, so I've left it for now.

**Comment by deniszh** [comment]
FYI here: Current `requirements.txt` is not compatible with Python 3. Now I just include it with docker image explicitly - https://github.com/graphite-project/docker-graphite-statsd/blob/python3/Dockerfile#L35-L48, but I'm wondering - how do we should work on that?
Having something like `requirements3.txt`?
Ideas? @DanCech @iksaif @piotr1212 @brutasse ?

**Comment by cbowman0** [comment]
Maybe using environment markers in requirements.txt? https://www.python.org/dev/peps/pep-0508/#environment-markers

**Comment by iksaif** [comment]
Why is our `requirements.txt` not compatible with python 3 ?

**Comment by deniszh** [comment]
@iksaif 
>Why is our requirements.txt not compatible with python 3 ?

Mostly because pinned versions of modules are not `python3` compatible.
```
python-memcached==1.47
txAMQP==0.4
simplejson==2.1.6
```

pip3-freeze giving me:
```
python-memcached==1.58
txAMQP==0.7.0
simplejson==3.13.2
```

**Comment by iksaif** [comment]
Do they really need to be pinned ? Isn't pip smart enough to see that some versions don't support Python2 ?

**Comment by deniszh** [comment]
@iksaif 
>Do they really need to be pinned ? 

Good question. Tried to run on latest modules - simplejson looks good, but I didn't test AMQP and memcached though :)

> Isn't pip smart enough to see that some versions don't support Python2 ?
No, it's quite dumb.

**Comment by takluyver** [comment]
Pip has recently gained the ability to look at the `Requires-Python` metadata to check compatibility with the Python version it's running on. E.g. if you `pip install ipython` on Python 2, you should get a compatible 5.x version: it will skip over the 6.x releases that require Python 3.

This was quite a recent shift, and only a few packages are taking advantage of it so far. But most packages still support Python 2, so the latest versions will probably work.

**Comment by deniszh** [comment]
Well. I think we can unpin `txamqp`, `memcached` and `simplejson` and fix any (possible) issues later.

**Comment by DanCech** [comment]
I wouldn't unpin, but I'd definitely be in favor of updating the pin to the latest versions if they support both 2 and 3.

**Comment by takluyver** [comment]
I've updated the pinned requirements to the latest available versions. It looks like they support Python 3 except for txamqp - the [work has been done](https://git.launchpad.net/txamqp/log/) for that one, but it's not yet in a release.

**Comment by takluyver** [comment]
Here's a better link for the txamqp work: https://github.com/txamqp/txamqp/pull/1

**Comment by DanCech** [comment]
We should be able to handle txamqp the same way as we do in carbon, try the import and ignore it if it fails.

**Comment by takluyver** [comment]
The only code that I can see importing txamqp is the `check-dependencies` script, so I assume it's only there as an indirect dependency via carbon.

**Comment by DanCech** [comment]
Ok, then we can either remove it or have the output say that AMQP support won't be available

**Comment by brutasse** [comment]
I think `simplejson` can go: it probably dates from when JSON support wasn't so great in the standard library. Nowadays `import json` behaves consistently across python versions `>=2.6` so [this](https://github.com/graphite-project/graphite-web/blob/5a58e4fd5ce13b8d4b22a068e6034888af6dff4e/webapp/graphite/util.py#L51-L62) is dead code.

**Comment by takluyver** [comment]
Yup, I agree. Do you want that removed as part of this PR?

**Comment by DanCech** [comment]
Good call @brutasse , there are quite a few other pieces of code that are leftovers from pre-2.7 support and can go also, like handling lack of cPickle and/or cStringIO

**Comment by deniszh** [comment]
>Yup, I agree. Do you want that removed as part of this PR?

Yes, it's fine, I think.

File: webapp/graphite/render/glyph.py [hunk 1]
```diff
@@ -12,21 +12,32 @@
...
-from graphite.util import json
+# BytesIO is needed on py3 as StringIO does not operate on byte input anymore
+# We could use BytesIO on py2 as well but it is slower than StringIO
+if sys.version_info >= (3, 0):
+  from io import BytesIO as StringIO
+else:
+  try:
+    from cStringIO import StringIO
+  except ImportError:
+    from StringIO import StringIO
```
DanCech: We can eliminate the try/except here since we don't need to support <2.7

webapp/graphite/render/views.py
```diff
@@ -14,13 +14,13 @@
...
-import httplib
+import six.moves.http_client
 
 from datetime import datetime
 from time import time
 from random import shuffle
-from urllib import urlencode
-from urlparse import urlsplit, urlunsplit
+from six.moves.urllib.parse import urlencode
+from six.moves.urllib.parse import urlsplit, urlunsplit
```
DanCech: consolidate to a single line

webapp/graphite/util.py
```diff
@@ -24,17 +24,26 @@
...
-  from StringIO import StringIO
+# BytesIO is needed on py3 as StringIO does not operate on byte input anymore
+# We could use BytesIO on py2 as well but it is slower than StringIO
+if sys.version_info >= (3, 0):
+  from io import BytesIO as StringIO
+else:
+  try:
+    from cStringIO import StringIO
+  except ImportError:
+    from StringIO import StringIO
```
DanCech: eliminate unneeded try/except

webapp/graphite/util.py
```diff
@@ -191,10 +196,15 @@ def load(cls, file):
   unpickle = SafeUnpickler
 
 else:
+  if sys.version_info[0] >= 3:
```
DanCech: This whole branch can be eliminated, every supported python has cPickle

webapp/tests/settings.py
```diff
@@ -57,7 +58,7 @@ def atexit_tmpremover(dirname):
     """ Utility to remove a temporary directory during program exit. """
     try:
         shutil.rmtree(dirname)
-        print("Removed temporary directory: %s" % dirname)
+        print(("Removed temporary directory: %s" % dirname))
```
DanCech: What is the second set of brackets for?

webapp/tests/test_events.py
```diff
@@ -79,7 +80,13 @@ def test_event_tags(self):
...
         response = self.client.get(creation_url)
         self.assertEqual(response.status_code, 200)
-        self.assertRegexpMatches(response.content, re.compile('^<html>.+<title>Events</title>.+Something happened.+<td>\\[u&#39;foo&#39;, u&#39;bar&#39;\\]</td>.+</html>$', re.DOTALL))
+        if sys.version_info[0] >= 3:
+            # Python 3: 'strings'
+            expected_re = b'^<html>.+<title>Events</title>.+Something happened.+<td>\\[&#39;foo&#39;, &#39;bar&#39;\\]</td>.+</html>$'
+        else:
+            # Python 2: u'strings'
+            expected_re = b'^<html>.+<title>Events</title>.+Something happened.+<td>\\[u&#39;foo&#39;, u&#39;bar&#39;\\]</td>.+</html>$'
+        self.assertRegexpMatches(response.content, re.compile(expected_re, re.DOTALL))
```
DanCech: We should make the output consistent between the 2 versions instead, I'd be in favor of dropping the `u` prefix in py2 output

webapp/tests/test_tags.py
```diff
@@ -376,7 +377,11 @@ def test_tag_views(self):
     response = self.client.post(url + '/tagSeries', {'path': 'test.a;hello=tiger;blah=blah'})
     self.assertEqual(response.status_code, 200)
     self.assertEqual(response['Content-Type'], 'application/json')
-    self.assertEqual(response.content, json.dumps(expected, indent=2, sort_keys=True))
+    if sys.version_info[0] >= 3:
+      self.assertEqual(response.content,
+                 json.dumps(expected, indent=2, sort_keys=True).encode('utf-8'))
+    else:
+      self.assertEqual(response.content, json.dumps(expected, indent=2, sort_keys=True))
```
DanCech: Instead of having these through the whole file, can we add a helper function to use throughout the file?
```
def json_dumps(obj, *args, **kwargs):
  if sys.version_info[0] >= 3:
    return json.dumps(obj, *args, **kwargs).encode('utf-8')
  return json.dumps(obj, *args, **kwargs)
```

takluyver:
I've cleaned up simplejson and StringIO as suggested. I'm leaving the pickle fallback, because the import works a bit differently on Python 3: you always import `pickle`, and it automatically uses the compiled version if it's available, or the pure Python fallback if not.

webapp/tests/test_tags.py
DanCech: See takluyver/graphite-web#1 for more on this, we should be able to handle this all in `graphite.util.json`

deniszh: 👍

DanCech:
As far as the ceres test, after mucking with it a bit I'm of the opinion that we shouldn't really be reaching into the ceres innards from within graphite-web at all.  The number of times it calls `os.listdir` is really an implementation detail that should be handled in the ceres test suite.  I'll update takluyver/graphite-web#1 shortly removing the references to os.listdir

setup.py
```diff
@@ -3,15 +3,18 @@
...
 
 from glob import glob
 from collections import defaultdict
 
 try:
-    from io import BytesIO
+    from StringIO import StringIO  # Python 2
 except ImportError:
-    from StringIO import StringIO as BytesIO
+    from io import StringIO  # Python 3
```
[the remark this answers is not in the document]
takluyver: It does, but `io.StringIO` is strictly unicode only, whereas `StringIO.StringIO` attempts to handle bytes and unicode. I'm not sure if Python 2's ConfigParser handles unicode properly, so this is a convenient fudge.

webapp/graphite/dashboard/views.py
```diff
@@ -14,7 +15,7 @@
 from graphite.compat import HttpResponse
 from graphite.dashboard.models import Dashboard, Template
 from graphite.render.views import renderView
-from send_graph import send_graph_email
+from .send_graph import send_graph_email
```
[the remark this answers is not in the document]
takluyver: The future import isn't needed to make this work - explicit relative imports work in Py2 as well.

The future import makes it a bit harder to add back implicit relative imports (i.e. what I changed this from). I can remove it if you prefer.

setup.py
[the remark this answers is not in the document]
DanCech: Interesting, definitely a case where I'd like to see a comment explaining what's going on

webapp/graphite/node.py
```diff
@@ -18,7 +18,7 @@ class BranchNode(Node):
 
 
 class LeafNode(Node):
-  __slots__ = ('reader', 'intervals')
+  __slots__ = ('reader', 'is_leaf')
```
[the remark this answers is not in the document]
takluyver: Ah, I misunderstood how slots work. Fixing.

webapp/graphite/util.py
takluyver: As far as I can work out, overriding the Python 3 pickle interface works this way even when it's using the compiled implementation, so we do need both branches. I'll make it clearer what's going on.

webapp/tests/settings.py
takluyver: Automatic tool being overenthusiastic. I'll fix it.

webapp/graphite/dashboard/views.py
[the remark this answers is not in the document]
DanCech: I think it makes sense to just use the absolute import and not bother with the __future__ import since the tests will all fail for py3.x if someone tries to use an implicit relative import

webapp/tests/test_tags.py
```diff
@@ -550,15 +559,15 @@ def test_tag_views(self):
...
+    self.assertEqual(response.content, json_bytes(expected, indent=2, sort_keys=True))
 
     # find nonexistent series
     expected = []
 
     response = self.client.get(url + '/findSeries?expr=name=test.a&expr=hello=tiger&expr=blah=blah')
     self.assertEqual(response.status_code, 200)
     self.assertEqual(response['Content-Type'], 'application/json')
-    self.assertEqual(response.content, json.dumps(expected, indent=2, sort_keys=True))
+    self.assertEqual(response.content, json_bytes(expected, indent=2, sort_keys=True))
```
DanCech: Technically since the last 4 tests don't pass `pretty=1` we should be expecting non-indented json (right now they pass because the indented json is only a single line).

takluyver:
I think I've responded to all the comments - let me know if you think I've missed some.

DanCech:
@deniszh how does this look to you?  I'm thinking that it makes sense to squash merge this one

iksaif:
👍  great !

Time to make biggraphite Py3 compatible now :P

deniszh:
Great! Let's merge this!
Will cut 1.1.0-pre5 (now with Python 3 support!)

takluyver:
Thanks everyone :-)

deniszh:
Thanks a lot for your contribution, @takluyver , it was impressive!

takluyver:
txAMQP 0.8.0 was just released with Python 3 support, if someone wants to bump the pinned version again.

Issue #2144, opened 2017-12-05
Javascript linting


cbowman0:
Few questions:
1. Who's the admin on the Codacy integration?  We should have it ignore the 3rd party javascript libraries, example code and the test scripts.

2. I'd like to set up an ESLint config file for the javascript files in [webapp/content/js](https://github.com/graphite-project/graphite-web/tree/master/webapp/content/js).  Anyone have a problem with that?

3. Anyone have a preference on ESLint settings?  If not, I'll attempt to match the most prevalent choices in the 5 javascript files that exist.

4. I think we can manipulate travis-ci to run the checks via another entry in the matrix options.  Anyone else have experience with multiple languages in the .travis.yml

deniszh:
1. We all three - @iksaif @DanCech and me. I can add you too - give me your (github account) email please. Or give me list of files which we should exclude.
2. Very 👍 . But how you going to run it?
3. Fine for me.
4. My Travis experience is quite limited, but I think it's currently not supported - https://github.com/travis-ci/travis-ci/issues/4090

DanCech:
Once we create an eslint config file codacy will use it

cbowman0:
My email is my github username at gmail.com.

https://stackoverflow.com/questions/27644586/how-to-set-up-travis-ci-with-multiple-languages suggests a way to make it work.  At least we can attempt it, but an ESLint config file can be used by Codacy, so it's not useless.

DanCech:
The easiest way is likely just to make another matrix entry that overrides all the defaults, installs node & eslint and runs it

deniszh:
>My email is my github username at gmail.com.

Please check now - https://www.codacy.com/

PR #2145, opened 2017-12-05
Adding new ESLint config file.

cbowman0:
Codacy will be the only thing that potentially uses this right now.

This is the first round of adding this in.  Nothing conforms completely to this, but that's to be expected.  It matches mostly what Codacy checks right now anyways.

I chose single quotes over double quotes since that seems to be the most predominant, but we can change that if consensus is another way (double or ignore)

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,34 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es6": true,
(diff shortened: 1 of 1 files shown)


cbowman0:
Start of work described in #2144

DanCech:
single quotes all the way 👍

PR #2146, opened 2017-12-06
Function plugin support

DanCech:
This PR implements support for function plugins, and provides mechanisms for loading custom plugins from files in `/opt/graphite/webapp/graphite/functions/custom` (for one-off functions) as well as from regular python packages (for 3rd-party function library packages).

It also provides a `/functions` API endpoint that exposes a list of available functions and some details about each one including the function signature and the docstring.  It also provides a mechanism for defining a `group` attribute, which could be used to provide grouping functionality similar to that seen in Grafana (I didn't go through and set that attribute for all the internal functions, but we'd want to do that before merging this).

I found myself using the `jsonResponse` decorator from the tags view, so I moved that into `graphite.util` and expanded its capabilities a little.  I also switched the events view to use it and got rid of the old `JsonResponse` class from `graphite.compat`.

To avoid circular import issues the new `graphite.functions` module exposes functions rather than raw `dict`s to delay the import of `graphite.render.functions`.  That also means that we're able to handle the user specifying a nonexistent function a little more gracefully.

As part of this work I have added several new functions and converted 16 existing functions into aliases:

- `filterSeries`: supports all the functionality of `(average|current|min|max)(Above|Below)` as well as additional comparisons and all available aggregations (see list below).  Those 10 functions are now aliases for `filterSeries` with the corresponding parameters.
- `highest` & `lowest`: support all functionality of `highest(Average|Current|Max)` and `lowest(Average|Current)` respectively, and supports all available aggregations.  Those 5 functions are now aliases.  All `highest*` functions now consistently return results in descending order.
- `sortBy`: supports sorting by any supported aggregation and sorting in reverse order, `sortByTotal` is now an alias.  `sortByMinima` and `sortByMaxima` are not aliases, because `sortByMinima` excludes empty series.

The only other BC-impacting change is that the median calculation correction in #2132 is now applied consistently, previously `movingWindow` and `movingMedian` still had the old non-adjusted behavior but now they are consistent with `aggregate` and will stay that way because they use the same underlying aggregation function.

The new documentation publishes a consistent set of names for the aggregation functions (`average, median, sum, min, max, diff, stddev, count, range, multiply, last`), but there is also a set of aliases (`rangeOf, avg, total, current`) that are supported to ensure backward compatibility.

As always I'd appreciate any and all comments/suggestions.


deniszh:
It's 👍 , but could you please elaborate what is `group` attribute and how grouping functionality works ?

DanCech:
@deniszh In Grafana the functions are organized hierarchically into groups, so the group functionality here is intended to help organize the functions and to support Grafana being able to pull a list of supported functions to build the UI so that we can get to a point where the Grafana UI can automatically show all supported functions (including any defined in plugins).

I also added a `params` block and fleshed out some examples of what that might look like.  I still need to update the docs to describe the structure for that but you can see the examples in the diff.

@torkelo would love to get your input on this!

DanCech:
Side note: Once we have the built-in functions categorized into groups we could think about breaking up `graphite.render.functions` into multiple `graphite.functions.<group>` files to help keep things manageable.

deniszh:
Ah, now I got it, thanks! 
Not sure if current mapping method is very convenient, though. OTOH we should do it only once....

torkelo:
Functions are grouped in the default graphite Web UI as well, which is where the grafana groups comes from :)

cbowman0:
Anything allowing this to be dynamic would be nice: https://github.com/graphite-project/graphite-web/blob/master/webapp/content/js/composer_widgets.js#L1072-L1216

DanCech:
@chris If you want to take a crack at the frontend that would be awesome ;) I'm also interested in any feedback/suggestions to the api output or the way the params are structured that would make that easier.

DanCech:
All functions are now documented with a group and parameter info, I'm open to suggestions for improvements to the current groups.  I also have Grafana able to load and use the definitions 😁 but still have to update the docs showing how the params definition works and documenting the available options.

Additional updates:
* The `_*_key` functions added in #2139 have been replaced with a single function `keyFunc` that works with all `safe*` functions, e.g. `seriesList.sort(key=keyFunc(safeLast))`
* All aggregation functions are centralized in `aggFuncs` (& `aggFuncAliases`), and `aggregate`, `movingWindow`, `legendValue`, `filterSeries` (see below), `aggregateLine`, `groupByNodes`, `smartSummarize`, `summarize` now all support the complete set of aggregations.
* New `filterSeries` function added, it consolidates handling of `maximumAbove`, `minimumAbove`, `maximumBelow`, `minimumBelow`, `currentAbove`, `currentBelow`, `averageAbove` & `averageBelow` into a single function that supports all defined aggregation functions and a complete list of comparisons (`=`, `!=`, `>`, `>=`, `<`, `<=`).  As part of this update `averageAbove` and `currentAbove` were changed to be consistent with `maximumAbove` and `minimumAbove` in using the `>` comparison (all the `*Below` functions were already consistent in using `<=`).

deniszh:
LGTM!
I didn't really like function describe format, but now looks better, and @DanCech made all hard work already. :)
(I'm still busy with release notes, pre-Christmas days are always hard...)

DanCech:
Good point, I'll add that

docs/functions.rst
```diff
@@ -11,8 +11,81 @@ Most functions are applied to one :term:`series list`. Functions with the parame
...
+Each function can have a docstring, ``.group``, and ``.params`` attributes defined, these are used in the function API output as hints for query builders.
+
+The ``.group`` attribute is the group name as a string, the ``.params`` attribute is a list of parameter definitions.
+
+Each parameter definition is a dict with the following attributes (note that `requestContext` is not included):
+
+- **name**: The name of the parameter
+- **type**: The parameter type, one of:
+
+  - **aggFunc**: An aggregation function name
```
iksaif: Can you add an enum for that ? (to avoid typos)

docs/functions.rst
```diff
@@ -11,8 +11,81 @@ Most functions are applied to one :term:`series list`. Functions with the parame
...
+    }
+
+Each function can have a docstring, ``.group``, and ``.params`` attributes defined, these are used in the function API output as hints for query builders.
+
+The ``.group`` attribute is the group name as a string, the ``.params`` attribute is a list of parameter definitions.
+
+Each parameter definition is a dict with the following attributes (note that `requestContext` is not included):
+
+- **name**: The name of the parameter
+- **type**: The parameter type, one of:
```
iksaif: Can you add an enum for that ? (to avoid typos)

webapp/graphite/functions/__init__.py
```diff
@@ -0,0 +1,70 @@
...
+  custom_modules = []
+  for filename in listdir(customDir):
+    module_name, extension = splitext(filename)
+    if extension != '.py' or module_name == '__init__':
+      continue
+    custom_modules.append(customModPrefix + module_name)
+
+  for module_name in custom_modules + settings.FUNCTION_PLUGINS:
+    module = import_module(module_name)
+    _SeriesFunctions.update(getattr(module, 'SeriesFunctions', {}))
```
iksaif: maybe add some function validations here ? (types are known, etc..)

torkelo:
The function model looks solid to me, should be enough to do what we want in the query builder UI in Grafana

DanCech:
@iksaif I think it's much clearer now with `Param` & `ParamTypes`, and the definitions are much more compact too.

@deniszh I think this is ready for a final review :)

docs/functions.rst
```diff
@@ -11,8 +11,142 @@ Most functions are applied to one :term:`series list`. Functions with the parame
...
+    def toUpperCase(requestContext, seriesList):
+      """Custom function that changes series names to UPPERCASE"""
+      for series in seriesList:
+        series.name = series.name.upper()
+      return seriesList
+
+    # optionally set the group attribute
+    toUpperCase.group = 'Custom'
+    toUpperCase.params = [
+      Param('seriesList', ParamTypes.seriesList', required=True),
```
deniszh: missing quote?

deniszh: Looks like extra.

PR #2147, opened 2017-12-06
Convert to single quotes since that is predominant.

cbowman0:
Change made via `eslint --no-eslintrc --rule 'quotes:[error,"single"]' --fix`


cbowman0:
Codacy appears to have a 50 issues per file max.  I removed some known issues by converting quotes, so new ones were "discovered".

PR #2150, opened 2017-12-06
work around timing issue in whisper reader interval tests

DanCech:
This PR aims to decrease the random errors we get from the whisper intervals tests.  As far as I can see it's a timing issue making the tests sometimes return a timestamp 1 less than we were expecting, so this just stops treating that as an error.

I'd like to do some work on whisper to make it less sensitive to wall-clock time, but for now this should stop the random annoying test failures.


deniszh:
Absolutely must! Thanks, @DanCech !

PR #2157, opened 2017-12-14
standardize url handling

DanCech:
This PR addresses some issues with the way that URL parsing is handled.

Django has something of an odd setup by default, where it expects that every URL will have a trailing slash.  This means that we require the user to include the trailing slash, so we end up with situations like `/version/` working but `/version` being a 404.  The simple way around this is to change the route path from `^version/` to `^version/?`, and at first that seems fine, but when we apply it to say `/functions` we can see that it causes problems.

With the current URL scheme we not only match `/functions`, `/functions/` and `/functions/<function>` but also `/functions<function>` because the final regex for function details is `^functions/?(.+)$`.

To solve this problem I'm proposing changing the top level section matching to exclude the slashes altogether and handling them in the 2nd-level rules instead.  This gives us `^functions`, `^/(.+)$` & `^/?$` for final matching rules `^functions/(.+)$` & `^functions/?$`, so now we match the expected URLs `/functions`, `/functions/` and `/functions/<function>` while no longer matching `/functions<function>`.

In this PR I have rolled that approach out across all the routes, and updated the `URL_PREFIX` handling so that we only wrap `graphite_urls` when `URL_PREFIX` is not empty.


deniszh:
Nice work, @DanCech!
Much appreciated!

PR #2159, opened 2017-12-18
Release Notes for 1.1.0 release

deniszh:
Please review. Will proofread that, but I'm not a native English speaker, so, please fix (or notify me about) errors and/or odd phrasing.

### DIFF
--- a/docs/functions.rst
+++ b/docs/functions.rst
@@ -10,12 +10,14 @@ Most functions are applied to one :term:`series list`. Functions with the parame
 ``*seriesLists`` can take an arbitrary number of series lists. To pass multiple series lists
 to a function which only takes one, use the :py:func:`group` function.
 
-.. _list-of-functions :
+.. _list-of-functions:
 
 List of functions
 -----------------
 .. automodule:: graphite.render.functions
    :members:
+
+.. _function-plugins:
 
 Function Plugins
 ----------------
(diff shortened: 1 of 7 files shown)


deniszh:
Thanks!
I have one more question btw.
Should we fork `1.1.x` branch or just tag 1.1.0 release in master and continue development there?
Previously we had incompatible `master` and `0.9.x` branches, so, it was logical to fork `1.0.x`.
But now we're in the different situation.
OTOH stable fork giving us much more freedom for incompatible changes in master, but then we will be forced to merge all master changes back to the stable branch. 
It's also possible to fork 1.1.x e.g. for `graphite-web` and `carbon` (where we have a majority of changes) and keep whisper/carbonate as tagged.
What do you guys think?
/cc @DanCech @iksaif @cbowman0 @obfuscurity

DanCech:
Good question, I haven't done enough release management with git to know the right answer but I'd tend to say that we should fork a 1.1.x branch and use it only for bugfixes, any new features should get released as 1.2.x anyway (as per https://semver.org/)

cbowman0:
I vote for a 1.1.x. branch.

iksaif:
Let's go for a branch, it master doesn't diverge too much it will be just a matter of merging so it won't be a big issue.

deniszh:
OK, submitting that, then forking and releasing. ;)

PR #2160, opened 2017-12-19
Rename release 1.1.0 to 1.1.1

deniszh:
I found out that 1.1.0 release is broken for carbon on pypi. Will rename it to 1.1.1 and will hide 1.1.0


DanCech:
I see 1.1.0 at https://pypi.python.org/pypi/carbon

deniszh:
yes, me too. But `pip install carbon` picks 1.0.2 :(

PR #2165, opened 2017-12-22
Fix hash position computation of fnv1a_ch.

iain-buclaw-sociomantic:
If the fnv1a hash of a given key returns a value less than 0x10000000, then the high and low parts of the hash would be split incorrectly, resulting in a wrongly computed hash position.

This removes the hex formatting and splitting of the fnv1a hash, and just calculates the position using bit manipulation.

Patch taken from graphite-project/carbon#715.

### DIFF
--- a/webapp/graphite/render/hashing.py
+++ b/webapp/graphite/render/hashing.py
@@ -73,11 +73,8 @@ class ConsistentHashRing:
 
   def compute_ring_position(self, key):
     if self.hash_type == 'fnv1a_ch':
-      big_hash = '{:x}'.format(int(fnv32a( str(key) )))
-      if len(big_hash) > 4:
-          small_hash = int(big_hash[:4], 16) ^ int(big_hash[4:], 16)
-      else:
-          small_hash = int(big_hash, 16)
+      big_hash = int(fnv32a(str(key)))
+      small_hash = (big_hash >> 16) ^ (big_hash & 0xffff)
     else:
       big_hash = compactHash(str(key))
       small_hash = int(big_hash[:4], 16)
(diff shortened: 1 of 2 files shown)


deniszh:
Cool, thanks a lot!

webapp/graphite/render/hashing.py
```diff
@@ -73,11 +73,8 @@ def __init__(self, nodes, replica_count=100, hash_type='carbon_ch'):
 
   def compute_ring_position(self, key):
     if self.hash_type == 'fnv1a_ch':
-      big_hash = '{:x}'.format(int(fnv32a( str(key) )))
-      if len(big_hash) > 4:
-          small_hash = int(big_hash[:4], 16) ^ int(big_hash[4:], 16)
-      else:
-          small_hash = int(big_hash, 16)
+      big_hash = int(fnv32a(str(key)))
+      small_hash = (big_hash >> 16) ^ (big_hash & 0xffff)
```
DanCech: I see that here we don't have the same python version check as in carbon, we should make them consistent.  The best solution may be to use `big_hash = int(fnv32a(six.binary_type(key)))` for both 2.x and 3.x.

DanCech: Actually that won't work because on 3.x `bytes` required an encoding parameter.  Likely the simplest solution is to use the same arrangement as in carbon.

DanCech: I realized that the carbon approach doesn't properly handle multibyte characters either, if #2168 makes sense to you @iain-buclaw-sociomantic @deniszh then I'll port it over to carbon also.

deniszh:
For me #2168 looks good, thanks, @DanCech !

iain-buclaw-sociomantic:
OK, lets go for #2168 then.

PR #2166, opened 2017-12-22
Add IRONdb to storage backends

redhotpenguin:
Adds IRONdb (https://irondb.io/) to the list of `Storage Backend Alternates` in the docs. Thanks!

### DIFF
--- a/docs/tools.rst
+++ b/docs/tools.rst
@@ -253,6 +253,9 @@ If you wish to use a backend to graphite other than Whisper, there are some opti
 `InfluxDB`_
   A distributed time series database.
 
+`IRONdb`_
+  A highly performant, highly scalable distributed time-series database compatible with the Graphite storage finder API
+
 `KairosDB`_
   A distributed time-series database written on top of Cassandra.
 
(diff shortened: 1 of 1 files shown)


deniszh:
IronDB is a good software but IMO it's hard to be named as "Storage Backend Alternative" for Graphite, it's compatible commercial alternative, though. It's also closed-sourced and commercial only (InfluxDB is open-sourced and the commercial version is only for clustering). OTOH we're not saying explicitly that we accepting only OSS additions to the list.
What do you guys think? /cc @iksaif @DanCech @obfuscurity ?

DanCech:
I agree, we're not looking to advertise a closed-source proprietary system that promotes itself as a replacement for Graphite.

redhotpenguin:
I saw other commercial alternative storage backends here as was mentioned, so it seemed to fit well to list IRONdb here. IRONdb is not a replacement for Graphite; it only acts as the storage component (hence the Whisper compatible API). IRONdb is closed core like some of the alternatives, but has open source components (like other alternate storage backends listed).

obfuscurity:
I'm a very strong :-1: on this.

deniszh:
Sorry, @redhotpenguin, I agreed with @DanCech and @obfuscurity here. As I said, IronDB is a good software, but it's
 - closed-source
 - commercial (not open-core as it's not open-source).
All other services on the list are open-source, even InfluxDB - which is open-core.
Take our sincere apologies, but we're believing in OSS and will promote it in a first place.

PR #2168, opened 2017-12-22
Fix hash position computation of fnv1a_ch

DanCech:
This PR is an attempt to future-proof the fnv1a_ch hash computation by properly handling multi-byte characters in addition to applying the fix described in graphite-project/carbon#714 and #2165.

The way we use the `fnv32a` hash function is now consistent with other hash functions where any string is encoded into utf-8 before being hashed.  This produces consistent results across python 2.x and 3.x with or without the pyhash library, and should be consistent with any third-party implementations.

I added pyhash to tox and travis, and added a quick test with multibyte characters to verify that the behavior is now consistent (the previous code would just crash and burn with a `UnicodeEncodeError` without pyhash, and produce inconsistent results with pyhash).

### DIFF
--- a/webapp/graphite/render/hashing.py
+++ b/webapp/graphite/render/hashing.py
@@ -16,14 +16,15 @@ limitations under the License."""
 from hashlib import md5
 from itertools import chain
 import bisect
+import sys
 
 try:
   import pyhash
   hasher = pyhash.fnv1a_32()
-  def fnv32a(string, seed=0x811c9dc5):
-    return hasher(string, seed=seed)
+  def fnv32a(data, seed=0x811c9dc5):
+    return hasher(data, seed=seed)
 except ImportError:
-  def fnv32a(string, seed=0x811c9dc5):
+  def fnv32a(data, seed=0x811c9dc5):
     """
     FNV-1a Hash (http://isthe.com/chongo/tech/comp/fnv/) in Python.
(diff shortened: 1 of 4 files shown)


deniszh:
Looks good!
We should apply similar patch to carbon too, right?

DanCech:
@deniszh yes, we definitely should.  Looking at hashing.py in carbon I see that actually there are quite a lot of differences besides the fnv1a_ch hashing that we should really resolve to be sure that it's doing the same thing as graphite-web.

webapp/graphite/render/hashing.py
```diff
@@ -97,6 +100,8 @@ def add_node(self, key):
       else:
         replica_key = "%s:%d" % (key, i)
       position = self.compute_ring_position(replica_key)
+      while position in [r[0] for r in self.ring]:
+        position = position + 1
```
DanCech: This is a port of graphite-project/carbon#196

iain-buclaw-sociomantic:
> Looking at hashing.py in carbon I see that actually there are quite a lot of differences besides the fnv1a_ch hashing that we should really resolve to be sure that it's doing the same thing as graphite-web.

Its almost as if you need a common library between the two projects. ;-)

deniszh:
Will backport this to 1.1.x, and release 1.1.2, but I think in 2018 - let's wait for some other fixes.

PR #2176, opened 2018-01-02
Add a README and examples for invoking test harness by hand

cbowman0:
It's been a little while since I wrote some tests and had to remember how to invoke the test harness and verify the test coverage, so I wrote it down.  Wording and everything open for discussion/adjustment and I may not be describing the optimal ways to do any of this.

* Add README for how the Django-based tests are run.
* Also adjust the coverage invocation so that manual runs can pass in specific test classes or test methods to run during debugging/testing.

### DIFF
--- a/webapp/tests/README.md
+++ b/webapp/tests/README.md
@@ -23,7 +23,7 @@ Passing multiple environments to tox can be done by separating with commas:
 `tox -e py27-django111-pyparsing2,py35-django111-pyparsing2,lint`
 
 To run only a specific test suite run (aka. run the file tests/test_finders.py):
-`tox -e py27-django111-pyparsing2,py35-django111-pyparsing2,lint-- tests.test_finders`
+`tox -e py27-django111-pyparsing2,py35-django111-pyparsing2,lint -- tests.test_finders`
 
 To run only a specific class in a test suite run (aka. run the MetricsTester class inside of test/test_metrics.py):
 
--- a/webapp/tests/README.md
+++ b/webapp/tests/README.md
@@ -41,7 +41,7 @@ Run `coverage html` from the webapp directory.  Afterwards, the webapp/htmlcov/i
 
 ### Setting up an environment to test
 
-There may need to be additional system packages or python dependencies installed in order for tox to successfully install the required packages for the tests.  Additionally, there may be additional system services needed to be running locally for the unit tests to complete successfully.
+There may need to be additional system packages or python dependencies installed in order for tox to successfully install the required packages for the tests.  Also, there may be system services running locally that are needed for the unit tests to complete successfully.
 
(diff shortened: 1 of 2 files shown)


webapp/tests/README.md
```diff
@@ -0,0 +1,42 @@
+# Graphite-Web Test Framework
+
+## Overview
+
+These Graphite-Web unit tests are for the Django portions of the graphite-web code base.  They utilize the Django test framework and run via `manage.py test`.
+
+For pull-requests, the tests run inside of Travic-ci using Python tox (https://pypi.python.org/pypi/tox) to invoke the tests across all the environments defined in tox.ini.
```
deniszh: Travis-ci ?

deniszh:
Very cool, indeed!
Thanks, @cbowman0 !

webapp/tests/README.md
cbowman0: Yep.  Will fix.

cbowman0: Thanks for reading through it.

cbowman0:
Additionally to this quick documentation, I was thinking it would be good to describe/document what dependencies are needed to have an environment where the tests would run.  I believe they currently need Python tox, coverage, a redis server on localhost and a mysql server on localhost.   Maybe an Ubuntu/CentOS VM?  Are there other ways that are simpler/easier?

DanCech:
Actually if you run tox with no additional parameters right now I think it's going to try and run a huge number of different environments (check out the output of `tox -l`) so we should probably be more prescriptive there.  I usually run `py27-django111-pyparsing2` and `py35-django111-pyparsing2` locally along with `lint`.

As far as requirements, those only need a local redis server.  If you want to run `-mysql` or `-postgresql` locally then you'll need to have servers running , database users, etc.  For 99.9% of development that isn't required though, only if you're hacking on the tag db.

cbowman0:
OK.  I changed the words to highlight the environments to test in and added them to the examples.  I'm done changing the words for now.

webapp/tests/README.md
```diff
@@ -0,0 +1,48 @@
...
+A set of minimum environments to run tests against are:
+`py27-django111-pyparsing2`
+`py35-django111-pyparsing2`
+`lint`
+
+Passing multiple environments to tox can be done by separating with commas:
+`tox -e py27-django111-pyparsing2,py35-django111-pyparsing2,lint`
+
+To run only a specific test suite run (aka. run the file tests/test_finders.py):
+`tox -e py27-django111-pyparsing2,py35-django111-pyparsing2,lint-- tests.test_finders`
```
DanCech: missing space?

cbowman0: Yes.

Issue #2175, opened 2018-01-02
build-index is broken


plockaby:
In commit [969cf84311644b97bb97d5431c4b7ad3894ff11f](https://github.com/graphite-project/graphite-web/commit/969cf84311644b97bb97d5431c4b7ad3894ff11f#diff-ab51a9546d45f7c2ee056421d2085881) the function ```write_index``` was removed. However, the ```build-index``` program still uses this function. Was ```write_index``` supposed to be removed or was ```build-index``` supposed to be removed?

deniszh:
I think `build-index` program should be removed, together with documentation. Index are handled by workers now.

DanCech:
It appears that `INDEX_FILE` is still used by `graphite.browser.views.search`, so maybe we should bring back `write_index` but update it to use the code that generates `index.json` instead of trying to walk the directories itself.

PR #2178, opened 2018-01-02
build-index

DanCech:
This PR restores support for the `build-index` command, which now uses `STORE.get_index()` to get the list of defined metrics.

Note that the `--ceres-dir` and `--whisper-dir` args have been removed as they aren't applicable any more.


DanCech:
Fixes #2175

PR #2179, opened 2018-01-04
Added doc for nginx+gunicorn stack (configuration and deployment)

Serphentas:


Serphentas:
Just so you know, the latest CI build fails because it seems Python 3.4 couldn't be downloaded.

DanCech:
It seems to work when you restart the job, not sure what the story is with that.

Thanks for the contribution, will check it out ASAP!

deniszh:
Did you check it, @DanCech ?

deniszh:
Ah, you referring to tests. I also faced this with python 35, looks like temporary glitches on Travis. Merging that now, will try to extend with different installation methods.
Thanks, @Serphentas !

Serphentas:
@deniszh Thanks for merging, and you're welcome.

I will also try to complete other parts of the doc when I have some time.

PR #2180, opened 2018-01-05
graphite: add support for raven/sentry.io (optional)

iksaif:


deniszh:
No additional config needed for Sentry?
Maybe we also should mention it in docs? (maybe not, just thinking out loud)

iksaif:
When you configure a Django app in Sentry they explain you what to put in your local_settings.py file, so the documentation is there already :)

Here I'm just making it so people don't need to import the settings.py file.

PR #2181, opened 2018-01-05
Some cleanups for storage.py

iksaif:


iksaif:
need to fix the tests

iksaif:
tests are happy on my machine, let's see if travis agrees.

DanCech:
Seems reasonable at first glance, I want to go through the logging and error message stuff in more detail

deniszh:
Needs to be rebased now, @iksaif  :(
Did you check it, @DanCech ?

iksaif:
Will rebase on monday, would like to get it in the next release as this really makes the error messages more useful, in particular in case of timeouts.

DanCech:
Haven't had a chance to look in depth yet :( will try to find some time this week.

iksaif:
Rebased

deniszh:
Cool, thanks!
@DanCech, could you please review it when you have time?

deniszh:
The release window is going to close. :)
More reviews, please?

DanCech:
I think this will be ok, one concern is that it uses 'Fetch' which is captialized vs 'find' etc which aren't.  There seems to only be one log entry that places the job/context description at the beginning of the output, so I'd suggest switching that to lowercase and potentially rewording that entry so it's within the line.

This also needs to be rebased, since the changes in config don't make sense after #2196

deniszh:
Not sure if rebase is needed - if Github not complaining then it should find a way how to merge it properly.

DanCech:
They won't make sense because the comment in settings.py refers to settings that have been changed.

deniszh:
I think it's already rebased, so https://github.com/iksaif/graphite-web/blob/06a1fafde6c96fa6c250767f7a69fb8adddf5a14/webapp/graphite/settings.py#L68-L72 is looks exactly as it is in master - https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/settings.py#L68-L71
Or what do you mean, @DanCech ?

webapp/graphite/settings.py
```diff
@@ -66,6 +66,7 @@
 POOL_MAX_WORKERS = 10
 
 # This settings control whether https is used to communicate between cluster members
+# The timeouts also apply to all finder/readers.
```
DanCech: This comment doesn't make sense any more.

iksaif: fixed

iksaif:
I changed the messages a little bit following @DanCech feedback

PR #2182, opened 2018-01-05
readers: re-introduce FetchInProgress

iksaif:
Previous readers would use FetchInProgress as a way
to run queries asynchrously. This leads to serious performance
regressions for plugins such as biggraphite which relied on that.

This re-introduces FetchInProgress and uses it in `fetch_multi()`
only.

### DIFF
--- a/webapp/graphite/finders/utils.py
+++ b/webapp/graphite/finders/utils.py
@@ -1,10 +1,18 @@
 """Utility functions for finders."""
-import time
 import abc
+import time
 
 from graphite.node import BranchNode, LeafNode  # noqa
 from graphite.util import is_pattern
 from graphite.intervals import Interval
+
+
+class FetchInProgress(object):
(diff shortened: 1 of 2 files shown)


deniszh:
Codacy is not happy. :|

iksaif:
better now

DanCech:
The same goal can be achieved by implementing `fetch` in the finder, but if we think it's going to be a common pattern we could include it here so authors don't have to deal with it.

iksaif:
I'll try to implement it in my finder to see how much more code it is :)

iksaif:
Here it is https://github.com/criteo/biggraphite/pull/317

deniszh:
@iksaif : so, this PR can be closed now, right?
Or it's useful to have `FetchInProgress` per se?

iksaif:
Let 's close it for now

PR #2191, opened 2018-01-08
hashing: better performances

iksaif:
See https://github.com/graphite-project/carbon/pull/726

### DIFF
--- a/webapp/graphite/render/hashing.py
+++ b/webapp/graphite/render/hashing.py
@@ -19,6 +19,11 @@ import bisect
 import sys
 
 try:
+  import mmh3
+except ImportError:
+  mmh3 = None
+
+try:
   import pyhash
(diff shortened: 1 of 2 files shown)


iksaif:
Wait for https://github.com/graphite-project/carbon/pull/727#issuecomment-355980527 to be merged, I'll include these changes too

deniszh:
@iksaif @DanCech 
Looks like `get_nodes()` change is not backward compatible, all tests are failed. Then we need to roll it back in carbon too...

iksaif:
Arf .. you're right, the order do matters because the replication factor is checked *after*. Please ignore this one for now, I'll send a fix for carbon.

iksaif:
pushed tentative fix, let's see if travis is happy on both sides

deniszh:
This works, cool! :)

Issue #2192, opened 2018-01-10
"All metrics failed for [metric]" exception since upgrading to graphite-web 1.1.1


wridgers:
Upgraded yesterday and immediately started seeing this:

```
Traceback (most recent call last):
  File "/opt/graphite/webapp/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/graphite/webapp/graphite/render/views.py", line 117, in renderView
    data.extend(evaluateTarget(requestContext, targets))
  File "/opt/graphite/webapp/graphite/render/evaluator.py", line 15, in evaluateTarget
    prefetchData(requestContext, pathExpressions)
  File "/opt/graphite/webapp/graphite/render/datalib.py", line 255, in prefetchData
    for result in STORE.fetch(pathExpressions, startTime, endTime, now, requestContext):
  File "/opt/graphite/webapp/graphite/storage.py", line 138, in fetch
    raise Exception('All fetches failed for %s' % (str(patterns)))
Exception: All fetches failed for [u'some.metric.*.*.*.here']
```

wridgers:
Increasing the `REMOTE_FETCH_TIMEOUT` stopped most of these, but afaiu graphite-web isn't doing any remote fetching here...

deniszh:
You not using clustering and `CLUSTER_SERVERS = []` in config?  
Hmm... Does setting USE_WORKER_POOL=False fixing that?

wridgers:
`CLUSTER_SERVER` and `USE_WORKER_POOL` are commented out, so I'd guess they're whatever the default is?

From local_settings.py:

```
#CLUSTER_SERVERS = ["10.0.2.2:80", "10.0.2.3:80"]
#USE_WORKER_POOL = True
```

Should we try setting `CLUSTER_SERVER = []` and `USE_WORKER_POOL=False`?

wridgers:
We have four caches, four aggregators, and two relays all running on the same machine as graphite-web (nginx + gunicorn). Relays are behind haproxy.

deniszh:
`CLUSTER_SERVERS = []` is on settings.py, so, should be fine. Try to set USE_WORKER_POOL=False 
If it helps then we can fix it like
`if settings.USE_WORKER_POOL & len(CLUSTER_SERVERS): ` in https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/storage.py#L86

[...]

PR #2194, opened 2018-01-10
Only use threads if we have any CLUSTER_SERVERS.

wridgers:
Fixes https://github.com/graphite-project/graphite-web/issues/2192


DanCech:
This isn't a good solution, one thing that might make sense is to only use the pool if we have more than one finder.  I'll look into that.

PR #2195, opened 2018-01-10
[WIP] Prometheus remote read support

DanCech:
This PR adds a new endpoint (currently `/render/prometheus` but could move to `/prometheus/read` or similar) that implements the Prometheus remote read API, allowing Prometheus to query data out of Graphite.

It accepts remote read requests (as snappy-compressed protobuf objects), converts them into native `seriesByTag` requests and executes the queries, then encodes and returns the results to Prometheus.

In order to use this you need to `pip install python-snappy protobuf` (for that to work you'll likely need to `sudo apt install libsnappy-dev` first), I didn't add them to requirements.txt due to the non-trivial installation for `python-snappy`.

This is still a work in progress, I intend to add tests and docs, and I'm working on a matching patch for carbon to add native remote write support.  Right now the simplest way to get Prometheus metrics into Graphite is to use https://github.com/prometheus/prometheus/pull/3533 but you can also use this PR to read regular Graphite data from within Prometheus.


deniszh:
Cool, nice feature!

iksaif:
Note that we plan to add tags support to https://github.com/criteo/graphite-remote-adapter very soon, at least for writes. That would work nicely with this. 

Our graphite-remote-adapter should hopefully be more up to date than the one in prometheus/documentation/example. I'll try to write more doc for it and advertise it a little bit more.

DanCech:
Once I add native support in carbon we won't need an adapter at all for the standard use-case, though looking at graphite-remote-adapter it seems like it would be useful for cases where more control over the final series names etc are needed.

iksaif:
Exactly, also when you have multiple prometheus but a single Graphite cluster you want to be able to control the prefix.

deniszh:
More is better in this case. 👍

PR #2196, opened 2018-01-10
Rename timeout settings

DanCech:
This PR renames the old `REMOTE_FIND_TIMEOUT` and `REMOTE_FETCH_TIMEOUT` settings to `FIND_TIMEOUT` and `FETCH_TIMEOUT` to better reflect how they're used in 1.1.x.  Backward compatibility is retained for anyone using the old names in `local_settings.py`.

It also increases the default setting for FETCH_TIMEOUT from 3 to 6 seconds.


deniszh:
Thanks, Dan, looks good!

PR #2199, opened 2018-01-11
Backport more commits to 1.1.x

deniszh:
[] Do not try merge series with different length
[] rename timeout settings
[] hashing: use a generator instead of looking for all targets
[] hashing: sync with carbon
[] hashing: better performances
[] app_settings: append() doesn't work on tuples
[] remove useless links in webapp config stack list
[] fix webapp config list
[] add doc for setup and deploy with nginx+gunicorn stack

### DIFF
--- a/docs/admin-webapp.rst
+++ b/docs/admin-webapp.rst
@@ -1,2 +1,19 @@
 Administering The Webapp
 ========================
+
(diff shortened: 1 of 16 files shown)


deniszh:
Backport script still doesn't work. Oh well...

PR #2201, opened 2018-01-13
Documentation fixes

deniszh:
- some changes for tools and config-web pages in the documentation.

### DIFF
--- a/docs/config-webapp.rst
+++ b/docs/config-webapp.rst
@@ -3,8 +3,8 @@ Configuring The Webapp
 
 There are multiple ways to expose the Graphite webapp. The following stack configurations exist:
 
-* Apache + mod_wsgi
 * nginx + gunicorn
+* Apache + mod_wsgi
 * nginx + uWSGI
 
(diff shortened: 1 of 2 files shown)


DanCech:
Looks good, not sure how up to date those carbon-api benchmarks are.  AFAIK graphite-remote-adapter doesn't (yet) support tags so I wouldn't call it "fully featured"

deniszh:
>Looks good, not sure how up to date those carbon-api benchmarks are.

Yep, that's why I put "In some benchmarks". Benchmarks are lie, everybody knows that.

>AFAIK graphite-remote-adapter doesn't (yet) support tags so I wouldn't call it "fully featured"

I hope they will finish it soo, right, @iksaif ? :)

deniszh:
We can add some fixes later, of course.

nnuss:
> Looks good, not sure how up to date those carbon-api benchmarks are.

They are almost certainly dated. When we made those comparisons the reference was 0.9.12 as I recall.

PR #2200, opened 2018-01-13
Some changes to the "tools" page

deniszh:
Collectors:
  - `telegraf` was added
 Storage Backend Alternates:
  - removing KairosDB and OpenTSDB. They're nice products but not working with Graphite in compatible way.
  - InfluxDB replaced with `influxgraph` - Graphite storage plugin.
  - Added `carbon/graphite-clickhouse`, `go-carbon`, `Kenshin` and `metrictank`
 Other:
  - Added `graphite-remote-adapter` and `carbonapi`

PR #2202, opened 2018-01-14
[1.1.x] Some changes to the "tools" page | Extending web config page | Adding cassabon

deniszh:
 - Some changes to the "tools" page (#2200)
 - Extending web config page (#2201)
 - Adding cassabon (#2201)


deniszh:
`backport` command is still working, until 10-20 commits or so. Weird.

DanCech:
https://github.com/sqren/backport/blob/master/src/lib/github.js has a hard-coded limit of 20 commits it seems

deniszh:
Ah, I suspected that. Then we need to that earlier.

PR #2204, opened 2018-01-14
Percentile aggregation documentation

wridgers:
Documentation for https://github.com/graphite-project/carbon/pull/696.

### DIFF
--- a/docs/config-carbon.rst
+++ b/docs/config-carbon.rst
@@ -159,13 +159,20 @@ The form of each line in this file should be as follows:
 
   output_template (frequency) = method input_pattern
 
-This will capture any received metrics that match 'input_pattern'
+This will capture any received metrics that match ``input_pattern``
 for calculating an aggregate metric. The calculation will occur
-every 'frequency' seconds and the 'method' can specify 'sum' or
-'avg'. The name of the aggregate metric will be derived from
-'output_template' filling in any captured fields from 'input_pattern'.
-Any metric that will arrive to ``carbon-aggregator`` will proceed to its
-output untouched unless it is overridden by some rule.
+every ``frequency`` seconds using a valid ``method``. The name of the aggregate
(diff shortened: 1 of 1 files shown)


docs/config-carbon.rst
```diff
@@ -104,6 +104,10 @@ Important notes before continuing:
 * ``aggregationMethod`` specifies the function used to aggregate values for the next retention level.  Legal methods are ``average``, ``sum``, ``min``, ``max``, and ``last``. The default is ``average``.
```
deniszh: Should you include percentiles' aggregation on this list too?

wridgers: My PR concerns carbon-aggregator aggregation methods, which I believe to be distinct from whisper storage aggregation methods that are defined [here](https://github.com/wridgers/whisper/blob/master/whisper.py#L552-L574).

Looks like the docs don't include `avg_zero`, `absmax`, or `absmin`.

I don't *think* it would make sense to add percentile aggregation to whisper file storage aggregation, what do you think?

deniszh: >My PR concerns carbon-aggregator aggregation methods, which I believe to be distinct from whisper storage aggregation methods that are defined here.

Yes, I got that - but then maybe the warning should be moved from line 107 to somewhere after line 176, i.e. after aggregator's aggregation methods?

wridgers: Ah yes, you are right. I understand now. Thinking back, I'm not sure why I decided 107 was a good place to put the warning! :)

Fixed and amended the offending commit!

wridgers:
Hey @deniszh, if this was merged then https://github.com/graphite-project/carbon/pull/696 should probably be merged too. :)

deniszh:
Sorry, I thought it was already merged, my bad. Done.

PR #2206, opened 2018-01-16
Make it possible to fully disable tags

iksaif:
This allows one to set TAGDB to None to disable the
tagging code entirely.

Also catch more errors when trying to parse tags. For example
people using the old syntax for time (time(foo) instead of time("foo"))
would have `[]` as a name which would break the match

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -43,12 +43,12 @@ class TimeSeries(list):
     else:
       self.tags = {'name': name}
       # parse for tags if a tagdb is configured and name doesn't look like a function-wrapped name
-      if STORE.tagdb and not re.match('^[a-z]+[(].+[)]$', name, re.IGNORECASE):
-        try:
+      try:
+        if STORE.tagdb and not re.match('^[a-z]+[(].+[)]$', name, re.IGNORECASE):
           self.tags = STORE.tagdb.parse(name).tags
-        except Exception as err:
-          # tags couldn't be parsed, just use "name" tag
-          log.debug("Couldn't parse tags for %s: %s" % (name, err))
+      except Exception as err:
+        # tags couldn't be parsed, just use "name" tag
+        log.debug("Couldn't parse tags for %s: %s" % (name, err))
 
   def __eq__(self, other):
     if not isinstance(other, TimeSeries):
--- a/webapp/graphite/tags/base.py
+++ b/webapp/graphite/tags/base.py
@@ -269,3 +269,27 @@ class BaseTagDB(object):
     spec = m.group(3)
 
     return (tag, operator, spec)
+
+
+class DummyTagDB(BaseTagDB):
+
+  def _find_series(self, tags, requestContext=None):
+    return []
+
+  def get_series(self, path, requestContext=None):
+    return None
+
+  def list_tags(self, tagFilter=None, limit=None, requestContext=None):
+    return []
+
+  def get_tag(self, tag, valueFilter=None, limit=None, requestContext=None):
+    return None
+
+  def list_values(self, tag, valueFilter=None, limit=None, requestContext=None):
+    return []
+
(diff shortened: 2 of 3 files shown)


webapp/graphite/render/datalib.py
```diff
@@ -43,12 +43,12 @@ def __init__(self, name, start, end, step, values, consolidate='average', tags=N
     else:
       self.tags = {'name': name}
       # parse for tags if a tagdb is configured and name doesn't look like a function-wrapped name
-      if STORE.tagdb and not re.match('^[a-z]+[(].+[)]$', name, re.IGNORECASE):
-        try:
+      try:
+        if STORE.tagdb and not re.match('^[a-z]+[(].+[)]$', name, re.IGNORECASE):
```
DanCech: was this throwing an exception?

DanCech:
I'm concerned about the potential errors when setting `STORE.tagdb` to `None`, since there are a number of places in the code that assume it will be a valid TagDB object.  I'd be more comfortable with a subclass of BaseTagDB that ignores input and returns empty results and/or emits log entries saying that the TagDB was called while not configured.

webapp/graphite/render/datalib.py
iksaif: name was a `[]`, which obviously doesn't work well with re.match(). You can reproduce that with `time(time)` (instead of `time("time")`)

iksaif:
I set tagdb to `None` because the few places using tagdb outsife of explicit tag code would check that. I'll thing about something better tomorrow if you are really against this one :)

webapp/graphite/render/datalib.py
DanCech: Aha, fair enough

DanCech:
I just don't want the whole thing to blow up when it tries to call a method on `None` 😉 It shouldn't be too hard to make a dummy TagDB to use for that case.

iksaif:
better :)

webapp/graphite/tags/base.py
```diff
@@ -269,3 +269,27 @@ def parse_tagspec(tagspec):
...
     return (tag, operator, spec)
+
+
+class DummyTagDB(BaseTagDB):
+
+  def _find_series(self, tags, requestContext=None):
+    return []
+
+  def get_series(self, path, requestContext=None):
+    return []
```
DanCech: `return None`

webapp/graphite/tags/base.py
```diff
@@ -269,3 +269,27 @@ def parse_tagspec(tagspec):
...
+
+  def list_tags(self, tagFilter=None, limit=None, requestContext=None):
+    return []
+
+  def get_tag(self, tag, valueFilter=None, limit=None, requestContext=None):
+    return []
```
DanCech: `return None`

webapp/graphite/tags/base.py
```diff
@@ -269,3 +269,27 @@ def parse_tagspec(tagspec):
...
+
+  def list_values(self, tag, valueFilter=None, limit=None, requestContext=None):
+    return []
+
+  def tag_series(self, series, requestContext=None):
+    return []
+
+  def del_series(self, series, requestContext=None):
+    return []
```
DanCech: `return True`

webapp/graphite/tags/base.py
```diff
@@ -269,3 +269,27 @@ def parse_tagspec(tagspec):
...
+
+  def get_tag(self, tag, valueFilter=None, limit=None, requestContext=None):
+    return []
+
+  def list_values(self, tag, valueFilter=None, limit=None, requestContext=None):
+    return []
+
+  def tag_series(self, series, requestContext=None):
+    return []
```
DanCech: `return series` or something like `raise Exception('Tag support disabled')`

iksaif:
I tested it locally plugged to carbon, it errors as expected

PR #2208, opened 2018-01-16
Adding Graphouse to the Tools page

deniszh:
Missed this commit from docs PR somehow


deniszh:
Travis has issues now, but local tests went fine.

PR #2207, opened 2018-01-16
Propagate `now` parameter for whisper reader

deniszh:
Should fix https://github.com/graphite-project/graphite-web/issues/196
Please check https://github.com/graphite-project/graphite-web/issues/196#issuecomment-357964575 for details

PR #2209, opened 2018-01-17
[1.1.x] Propagate `now` parameter for `fetch()` in whisper reader | Revert "Graphouse was added" | removing extra parameter in GzippedWhisperReader | Graphouse was added | proper sorting

deniszh:
 - Propagate `now` parameter for `fetch()` in whisper reader (#2207)
 - Revert "Graphouse was added" (#2207)
 - removing extra parameter in GzippedWhisperReader (#2207)
 - Graphouse was added (#2208)
 - proper sorting (#2208)


deniszh:
Broken

PR #2213, opened 2018-01-25
functions: fix groupByNode() output

iksaif:
`groupByNode(perSecond(foo.bar,1), 1)` would display `bar,1)` instead of `bar`
before these change. I suspect that we might want that for all calls of `aggKey()`

Current workaround is to aliasByNode() before groupping.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -192,6 +192,9 @@ def getNodeOrTag(series, n, pathExpression=None):
   return series.tags.get(str(n), '')
 
 def aggKey(series, nodes, pathExpression=None):
+  # if series.name looks like it includes a function, use the first path expression
+  if pathExpression is None and series.name[-1] == ')':
+    pathExpression = _getFirstPathExpression(series.name)
   return '.'.join([getNodeOrTag(series, n, pathExpression) for n in nodes])
 
 def normalize(seriesLists):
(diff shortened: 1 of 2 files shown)


DanCech:
Good point, it may make sense to update `aggKey` to do this automatically.  Something like this may do the trick:
```
def aggKey(series, nodes, pathExpression=None):
  # if series.name looks like it includes a function, use the first path expression
  if pathExpression is None and series.name[-1] == ')':
    pathExpression = _getFirstPathExpression(series.name)
  return '.'.join([getNodeOrTag(series, n, pathExpression) for n in nodes])
```
With that patch `aliasByNode` would no longer need to pass `pathExpression` to `aggKey` either.

iksaif:
Last time we did something like that I think it broke aliasByNode for some people, do you see anything that would break doing this for all functions ?

DanCech:
I think if we're going to change it then the best thing to do is to change it globally vs having the behavior be different depending on the function that you're using.  I'm not a fan of `aliasByNode` being different from the other functions today and don't want to compound that issue by adding more functions that use `_getFirstPathExpression` without applying it equally across all functions.

With the change to use `getNodeOrTag` we fixed some of the issues that would previously occur if the user specified a node that couldn't be found, etc so I think we're in a much better position to make this change now than we were in the past.

iksaif:
Ok, the code you suggested looks good

PR #2215, opened 2018-01-31
Update config-carbon.rst

wfhu:
I tried to change the config file 'storage-aggregation.conf', added a section at the bottom, and it takes no effect. So after I read all the documents in full, I think the order of sections for file 'storage-aggregation.conf' also matters. So I think it is important to add the order explanation to the 'storage-aggregation.conf' document sections. 

I have been trapped by this order problem for a while.

I think it would be helpful to other people after this clarification.

### DIFF
--- a/docs/config-carbon.rst
+++ b/docs/config-carbon.rst
@@ -100,6 +100,7 @@ Important notes before continuing:
 
 * This file is optional.  If it is not present, defaults will be used.
 * The sections are applied in order from the top (first) and bottom (last).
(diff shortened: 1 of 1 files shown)


deniszh:
Yes, thanks for clarification, @wfhu !
LGTM, but I'm not native English speaker. Maybe someone else can review? @DanCech @cbowman0 ?

gwaldo:
The clarification is helpful!  Thank you!

I think that the two sentences can be combined into one, but I'm ok with the change as-is.

Issue #2222, opened 2018-02-05
Unable to visualize available metric points in a whisper file


raghavan20:
Given a whisper file with the archive `(1s:14d)`

```
whisper-info.py direction.wsp 
maxRetention: 1209600
xFilesFactor: 0.5
aggregationMethod: average
fileSize: 14515228

Archive 0
retention: 1209600
secondsPerPoint: 1
points: 1209600
size: 14515200
offset: 28
```

having data points for 16 July 2017.
```
...
321184: 1500199984,          0
321187: 1500199987,          0
321190: 1500199990,          0
321193: 1500199993,          1
321198: 1500199998,          0
...

$ for i in 1500199984 1500199987 1500199990 1500199993 1500199998 ; do  echo "$i - `date -d @$i`"; done
1500199984 - Sun 16 Jul 11:13:04 BST 2017
1500199987 - Sun 16 Jul 11:13:07 BST 2017
1500199990 - Sun 16 Jul 11:13:10 BST 2017
1500199993 - Sun 16 Jul 11:13:13 BST 2017
1500199998 - Sun 16 Jul 11:13:18 BST 2017
```
When viewing the same data from `graphite-web`, there are not results.

```
http://localhost/render?from=10%3A00_20170716&until=14%3A00_20170716&width=400&height=250&target=direction&_uniq=0.24008196147297345&title=DEVICE_3375.direction&rawData=1

empty result
```

Can someone please explain why this happens? Is graphite-web performing some kind of aggregation and considering most data points as null? For some time durations (over several seconds) in the whisper file, all available slots are not filled.

DanCech:
Right now whisper always calculates available data from the current time, so it won't be able to retrieve data older than 14 days.  You can work around it via the render API if you're running master or the 1.1.x branch by passing `now=14%3A00_20170716` in your render request.

raghavan20:
Sorry, I do not understand your comment completely as I managed to read data points in the whisper file using `whisper-dump` utility.

```
# Searching for interested data points based on timestamp
whisper-dump.py direction.wsp  | grep 150019
```

why do you say that i cannot visualize data older than 14 days?

deniszh:
@raghavan20 : You have single `1s:14d` archive in whisper. That means "Store data every 1 second and keep last 14 days of data". So, Graphite by default takes 14 days from current date and renders data - but there's no data for last 14 days - then it shows nothing.
[...]

PR #2223, opened 2018-02-05
handle case when node.fetch returns None

DanCech:
This PR fixes an exception thrown when a reader returns None

#2222


deniszh:
Ah, that's how `None` coming instead of `[]`. Thanks!
Will backport it to 1.1.x and prepare release notes of 1.1.2, now we have more then enough patches for it.

PR #2225, opened 2018-02-07
Add seasonality parameter for holt-winters functions

LTMXcitrus:
This PR adds the possibility to specify the seasonality when asking for a Holt-Winters prediction.
Before, the seasonality was to 1 day by default.


LTMXcitrus:
Sorry about that, I'm a python newbie and I must have missed this...
Should be ok now (all tests are ok on my PC)

DanCech:
Looking much better, the last thing that needs to be done is to update the parameter definitions, you'll want to add an entry for seasonality after the `Param('bootstrapInterval', ParamTypes.interval, default='7d', suggestions=['7d', '30d']),` line for each function.  This is used to expose the parameter to make the new parameter available in systems like Grafana.

LTMXcitrus:
Thanks for your help, it's done as well :+1:

DanCech:
Thanks @LTMXcitrus 🎉

PR #2226, opened 2018-02-09
functions: fix holtWinterAberration with multiple targets

iksaif:


DanCech:
@deniszh as far as I can see the function signature didn't change, it still accepts a single series list.

deniszh:
Ah, nvm, I'm blind. Friday 🤷‍♂️

iksaif:
Two time series, but still one single `seriesList` :)

PR #2240, opened 2018-02-22
Add metrics for Graphite Web

iksaif:
Ironically, graphite-web exposes no metrics at all, which makes it hard to
monitor things (people can implement things in their webservers but this
won't have the same level of details)

This takes advantage of Prometheus's client library and django integration
to add django-related metrics and allow for additional metrics to be added
later. Bonus point, we get Prometheus support too.


webapp/graphite/urls.py
```diff
@@ -33,8 +33,16 @@
     url('^s/(?P<path>.*)', shorten, name='shorten'),
     url('^S/(?P<link_id>[a-zA-Z0-9]+)/?$', follow, name='follow'),
     url('^$', browser, name='browser'),
+
 ]
 
+try:
+    graphite_urls.append(
+        url('^prometheus/', include('django_prometheus.urls')),
```
deniszh: should it be something like `prometeus-metrics/` ? Thinking about possible (logical) clash with https://github.com/graphite-project/graphite-web/pull/2195

iksaif: `prometheus-metrics/metrics` isn't nice. Maybe having `prometheus/metrics`, `prometheus/write`, `prometheus/read` isn't really an issue ? I could just create a prometheus apps that will later be extended to add `/read` and redirect its `/metrics` to django-prometheus

DanCech:
I like the idea a lot, but I'm much less keen on coupling ever more tightly with django, and I would like to see instrumentation of at least the basic steps in the find, render and processing pipeline so we can get some of the same stats that right now are just in the logs.

iksaif:
> but I'm much less keen on coupling ever more tightly with django,

Per view/method/service stats really need to be done at the framework level. If it ever moves to another framework the metrics will change, but that's fine. Using django-prometheus gives us most of the metrics we want for free

> find, render and processing pipeline

Sure, I'll add something

FYI, here is how it currently looks like: [metrics.txt](https://github.com/graphite-project/graphite-web/files/1747883/metrics.txt)

deniszh:
I also not really like more Django, but it's better to have some metrics then no metrics, right?
Maybe default prefiux should be switched to `graphite.django.` ?

iksaif:
The django related metrics already have Django in their names. Would not
make much sense for process_fds for example :)

iksaif:
Here it is with a few basic metrics

deniszh:
Yes, but now we have hard dependency on `prometheus_client` which make me sad.

DanCech:
Yeah, those metric names are horrible also.  I think that it makes more sense for us to build this ourselves or to use a more graphite-friendly library.

iksaif:
Maybe.. I'll experiment with django-prometheus insternally then, it's doable to integrate it using local_settings.py only

Issue #2243, opened 2018-02-26
Tags autoComplete timing-out on clustered setup (infinite loop!)


Quentin-M:
Hi there,

I am using a graphite cluster with 4 carbon-cache processes, on 4 different nodes, that run alongside 4 graphite-web instances. Fronting those, 2 cache-relay instances.

When starting to type a query in Grafana (first node field), Grafana issues two API calls:
- `metrics/find?from=X&until=Y&query=*`
- `tags/autoComplete/tags` (or `tags/autoComplete/tags?tagPrefix=%2A&limit=100`)

However, it looks like something is really wrong with the tags request. First of all, the request takes forever and eventually time-out. But also, after a few attempts in grafana, the logs start spamming with timeouts for tags and tracebacks reading the tags from the other servers (see below), and it keeps going forever, even though no more requests are being made by grafana whatsoever. It seems that there is an endless loop somehow, like if `local=1` was not respected, or something along those lines. Then indeed, the web servers stops being able to even respond to `metrics/find` requests.

However, if in Grafana, we set the graphite version to <1.1, no `tags/autoComplete/tags` queries are made, and everything works wonderfully: auto-completion is snappy and graphs are showing perfectly.

Furthermore, when curling a single time the `tags/autoComplete/tags` API endpoint (see below), the logs go bonkers yet again, and basically never stops hard trying. According to the logs below, it'd seem like, indeed, `local=1` (as well as the msgpack`) parameters are missing from the queries, which would create an infinite loop?

```
curl 'localhost:8080/tags/autoComplete/tags?tagPrefix=%2A&limit=100' 
```

```
2018-02-26,07:54:21.246 :: RemoteFinder[graphite-cache-3:8080] Error response 500 from http://graphite-cache-3.graphite-cache.dev.svc.cluster.local:8080/tags/autoComplete/tags?tagPrefix=%2A&limit=100
NoneType: None
2018-02-26,07:54:21.405 :: RemoteFinder[graphite-cache-3:8080] Error response 500 from http://graphite-cache-3:8080/tags/autoComplete/tags?tagPrefix=%2A&limit=100
NoneType: None
2018-02-26,07:54:21.421 :: RemoteFinder[graphite-cache-2:8080] Error response 500 from http://graphite-cache:8080/tags/autoComplete/tags?tagPrefix=%2A&limit=100
NoneType: None

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #2244, opened 2018-02-26
w/g/f/remote.py: add missing local field to auto_complete_{tags,values}

Quentin-M:
This PR makes the `tags/autoComplete/tags` and `tags/autoComplete/values` API calls work properly in a clustered configuration of graphite-web. It adds the `localOnly` query parameter as needed in order to avoid an infinite recursion of API calls across the cluster, similarly to `/metrics/find`.

Fixes https://github.com/graphite-project/graphite-web/issues/2243
Follows https://github.com/graphite-project/graphite-web/pull/2128

### DIFF
--- a/webapp/graphite/finders/remote.py
+++ b/webapp/graphite/finders/remote.py
@@ -189,6 +189,7 @@ class RemoteFinder(BaseFinder):
         fields = [
             ('tagPrefix', tagPrefix or ''),
             ('limit', str(limit)),
+            ('local', self.params.get('local', '1')),
         ]
         for expr in exprs:
             fields.append(('expr', expr))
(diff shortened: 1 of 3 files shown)


Quentin-M:
```
FAIL: test_WhisperReader_get_intervals (tests.test_readers_whisper.WhisperReadersTests)
  File "/home/travis/build/graphite-project/graphite-web/webapp/tests/test_readers_whisper.py", line 129, in test_WhisperReader_get_intervals
    self.assertEqual(int(interval.start),ts - 60)
AssertionError: 1519637574 != 1519637573
```

^ Don't think my PR has anything with it.

PR #2254, opened 2018-03-06
fix usage of "unicode" in glyph.py for Python 3

JelleZijlstra:


JelleZijlstra:
The codacy failure is due to an issue that was already present in the previous code. I'm happy to fix it of a maintainer requests it, but wanted to limit the scope of this PR.

DanCech:
Please go ahead and switch it to use isinstance

JelleZijlstra:
Done, thanks for the quick review.

PR #2266, opened 2018-03-27
graphite functions: handle max on empty sequences better

JelleZijlstra:
I was seeing errors due to code like safeMax(series) > 0, where safeMax
returned None.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -130,15 +130,19 @@ def safeLast(values):
   for v in reversed(values):
     if v is not None: return v
 
-def safeMin(values):
+def safeMin(values, default=None):
   safeValues = [v for v in values if v is not None]
   if safeValues:
     return min(safeValues)
+  else:
+    return default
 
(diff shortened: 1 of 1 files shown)


webapp/graphite/render/functions.py
```diff
@@ -3257,7 +3261,7 @@ def sortByMinima(requestContext, seriesList):
     &target=sortByMinima(server*.instance*.memory.free)
 
   """
-  newSeries = [series for series in seriesList if safeMax(series) > 0]
+  newSeries = [series for series in seriesList if safeMax(series, default=0) > 0]
```
DanCech: This doesn't seem like it would actually change the behavior of the function

JelleZijlstra: It does, because previously `safeMax()` might return None, and `None > 0` is a TypeError in Python 3.

DanCech: Ah ok, I didn't realize that was a problem on 3.x

PR #2281, opened 2018-04-14
fix typo: matric -> metric in feeding-carbon.rst

ngash:


docs/feeding-carbon.rst
```diff
@@ -32,7 +32,7 @@ On Unix, the ``nc`` program (``netcat``) can be used to create a socket and send
 If you use the OpenBSD implementation of ``netcat``, please follow this example:
 
   .. code-block:: none
- 
+
```
deniszh: Not sure if an empty line is safe/needed here...

ngash: According to [0] and [1], a blank line indicates the directive's content. I'm thinking *blank line* also includes lines with whitespace only.

[0] http://www.sphinx-doc.org/en/stable/rest.html#directives
[1] http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#directives

DanCech: It's fine, not sure why this file had the trailing whitespace but it isn't needed (or desired).

PR #2294, opened 2018-05-21
handle highest/lowest thresholds specified as numeric strings

DanCech:
If a user specifies the `n` parameter to any of the `highest*` or `lowest*` functions as a numeric string `'20'` rather than a bare number `20` they will receive the error `TypeError: slice indices must be integers or None or have an __index__ method` since the string value can't be used to slice the list of series.

This PR implements a very simple fix by casting the provided value to an integer before using it to slice the series list, which will avoid this issue.

Longer-term a better solution to this kind of problem is likely going to be to use the function definition to perform validation and normalization in `evaluateTokens` before calling the actual function.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2697,7 +2697,7 @@ def highest(requestContext, seriesList, n=1, func='average'):
   Draws the 5 servers with the highest number of busy threads.
   """
   seriesList.sort(key=keyFunc(getAggFunc(func)), reverse=True)
-  return seriesList[:n]
+  return seriesList[:int(n)]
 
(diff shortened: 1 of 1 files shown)


deniszh:
Well, this patch will fix that specifiic usecase, indeed.
But I'm just wondering what exact use case it is? `n` in `lowest()` is not a time interval, it's a number. Graphite is written in Python, in Python `string` and `integer` is different entities, so, why on earth someone putting string into integer field in function? :)

DanCech:
It can happen in Grafana if you try to use a template variable to make the threshold select-able, since it wraps the value with quotes by default.

deniszh:
Ah, right. OK, thanks.

Issue #2218, opened 2018-02-01
Bug: Metrics tree not loaded when CLUSTER_SERVERS member is unavailable - fixed just in 1.0


mcbsdEU:
Hi,

I refer to an issue #1791, this has been fixed in past, but seems just in 1.0.x.
1.1.x has different file structure and fix is not valid anymore.
In order to bypass the issue in very dirty way, I had to modify:
..graphite/finders/remote.py:
line 262: bypass Exception by inserting "return" right after:
```
        except BaseException as err:
            return
            self.fail()
            log.exception("RemoteFinder[%s] Error requesting %s: %s" % (self.host, url_full, err))
            raise Exception("Error requesting %s: %s" % (url_full, err))
```

line 116: same workaround as above:
```
                    result, buffer_size=settings.REMOTE_BUFFER_SIZE))
            except Exception as err:
                return
                self.fail()
                log.exception(
                    "RemoteFinder[%s] Error decoding find response from %s: %s" %
                    (self.host, result.url_full, err))
                raise Exception("Error decoding find response from %s: %s" % (result.url_full, err))
```

line 123: comment the  "finally" clause:
```
            #finally:
            #    result.release_conn()

            cache.set(cacheKey, results, settings.FIND_CACHE_DURATION)
```

It would be very very helpful to have this fixed in 1.1.x
It has huge importance for High Availability situations, when one node of cluster is down because with a current bug, whole solution is not available because of one non-functional node.

Thanks in advance!

FYI:
This is snippet of the error for future generations (affected node has no httpd running, otherwise it would shout out timeout error):
```
Exception: Error requesting http://a.b.c.d:80/metrics/find/?local=1&format=pickle&query=%2A: HTTPConnectionPool(hos
t='a.b.c.d', port=80): Max retries exceeded with url: /metrics/find/?local=1&format=pickle&query=%2A (Caused by New
ConnectionError('<urllib3.connection.HTTPConnection object at 0x56013b692490>: Failed to establish a new connection: [Errn
o 111] Connection refused',))
```

deniszh:
@mcbsdEU : Thanks for reporting!
@DanCech @iksaif : It's about scenarion when one node of the cluster is down.
Previous fix of @iksaif is https://github.com/graphite-project/graphite-web/commit/c6cf8244862ec15bf25b4cb4b01a29e45f59e2eb
Do you  have an ideas how to fix that properly in current codebase?

DanCech:
[...]

PR #2303, opened 2018-05-30
Fail hard

clusterfudge:
I have a monitoring system that reads from a graphite cluster with 10 stateless "frontend" instances, and 50 stateful "backend" instances. No replication, so a single host being unavailable results in a subset of data being unavailable. Several alert rules query multiple timeseries (and compute functions against them). When a request fails to a backend instance, the frontend ignores the failure and serves (or computes with) partial data, resulting in incorrect data served to upstream monitoring systems. The net result is false alarms, and developers getting paged in the middle of the night.

* What I attempted
Calculate error ratios using ```asPercent(fail, sumSeries(pass, fail))```

* Expected Result
The correct error ratio on happy path, or an HTTP 500 when one or more of backends is unavailable.

* Witnessed Result
In cases where the backend responsbile for either ```fail``` or ```pass``` is unavailable, graphite-web is swallowing the error and returning the incorrect result. In the case of ```pass``` being unavailable, the query gets transformed into:
```
asPercent(fail, sumSeries(fail))
```
Which returns 100 for all data points.

* System Details
Ubuntu 16.04 LTS
Python 2.7.6
Graphite Component Versions: 1.1.2


clusterfudge:
@DanCech FYI

clusterfudge:
Related to conversation in https://github.com/graphite-project/graphite-web/issues/2218#issuecomment-362283812

deniszh:
I'm 👍 but not sure does default should be True here. It's changing existing behavior and other users use replication, so, current behavior can be acceptable.
Also "All requests failed" error message is misleading here, probably it's better to introduce separate `if` with distinct error message.

clusterfudge:
I actually intended False to be the default to preserve existing behavior (late night coding, always a winner). I'll update that, and rewrite the conditional to be more straightforward. Thanks for the feedback!

DanCech:
I agree with Denis, with those changes this should be merge-able.

clusterfudge:
@deniszh I believe the default in the current iteration is that in settings, we'll default to False (preserving existing behavior), and I've update the local_settings.py.example with a value of true, but commented out (so the action to change behavior is uncommenting the line). Is that not the correct approach here?

clusterfudge:
@DanCech and @deniszh unit tests added and local_settings.py updated. Please take a look. Thanks!

deniszh:
LGTM

DanCech:
Actually, thinking about this a little more the name of the config setting should probably be changed since it's actually applicable to all stores.  One possible option would be `STORE_FAIL_ON_ERROR` but I'm open to any suggestions.

clusterfudge:
I've continued digging on another edge case that's bugging my team, and I think there's another issue I want to capture this (we currently only capture failures, not timeouts). Placing on hold.

clusterfudge:
@DanCech @deniszh This is now ready for review again, thanks for your patience. 

I understand based on our conversation (and the comments in #2218 ) that there's a desire to yield a subset of metrics when not all backends are available. I'm of the mindset (and it seems that as of a few months ago @DanCech [agreed](https://github.com/graphite-project/graphite-web/issues/2218#issuecomment-362283812) ) that this is incorrect, and is more akin to your database silently lying to you. While users manually browsing metrics may find this behavior acceptable, it makes it near impossible to build tools on top of the graphite api without consistency guarantees. 

My recommendation would be to continue a full audit of the codebase and abandon anywhere we silently swallow errors and return partial results. I'm hoping this is the only place, but I'm honestly not sure. I'm still very new to the source (even though I've been loving graphite since 2011, so I guess egg on my face :)).

deniszh:
I'm agreed with @DanCeh here that Graphite should not lie to a user and return broken data to a user by default. But it can be viable option to enable because e.g. if you copy your metrics to 2 servers in parallel and one of the servers gone - you still can use another copy - it's a perfectly valid use case. Unfortunately, it's almost impossible to be sure if all your data valid in all cases, so, IMO, we can have an option in properties to enable this behavior - if a user knows what he/she's doing.

clusterfudge:
@DanCech @deniszh anything blocking us here?

PR #2304, opened 2018-06-07
Allow history tracking in composer

yuzawa-san:
Adds forward / backward functionality too the composer for easier navigation. Also, the url updates whenever the composer is modified, allowing the (long) url to be copy-pasted from the address bar.

Test locally in Firefox, Chrome, and Safari.


deniszh:
Looks good - thanks!
I'm not a big JS guru, though...

Issue #2313, opened 2018-07-01
Fetch timing out for large queries on 1.1.2


JosephFY:
I am getting the bellow error when doing a large query in our new graphite instance
```
Traceback (most recent call last):
  File &quot;/usr/lib64/python2.7/site-packages/django/core/handlers/base.py&quot;, line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File &quot;/graphite/webapp/graphite/render/views.py&quot;, line 117, in renderView
    data.extend(evaluateTarget(requestContext, targets))
  File &quot;/graphite/webapp/graphite/render/evaluator.py&quot;, line 15, in evaluateTarget
    prefetchData(requestContext, pathExpressions)
  File &quot;/graphite/webapp/graphite/render/datalib.py&quot;, line 257, in prefetchData
    for result in STORE.fetch(pathExpressions, startTime, endTime, now, requestContext):
  File &quot;/graphite/webapp/graphite/storage.py&quot;, line 179, in fetch
    &#39;fetch for %s&#39; % str(patterns))
  File &quot;/graphite/webapp/graphite/storage.py&quot;, line 121, in wait_jobs
    raise Exception(message)
Exception: Timed out after 6.000316s for fetch for [u&#39;Devices.USA.*.*.*.*.*.*_device_status&#39;]

</pre>
</div>

</center>
```

This is only happening on 1.1.2 , I have another graphite instances running under 0.9.13 and I can run the query on without any issue, the query returns values for 8k+ metrics.

When eliminating the wildcards to get ~1000 metrics, the query runs smoothly 

tried to increase the ``` CARBONLINK_TIMEOUT ``` to 10 but no luck. 

Any ideas ?

deniszh:
You need to increase fetch timeout instead. In latest Graphite variable named `FETCH_TIMEOUT` but probably in 1.1.2 it's still named `REMOTE_FETCH_TIMEOUT`.

DanCech:
This is interesting work, it's been a while now but I seem to recall doing benchmarking at the time that didn't show anything like a 500x slowdown.  If you can share your benchmarking setup I'd be keen to see if we can figure out what's going on and whether there's a way to have the best of both worlds.

JosephFY:
[...]

PR #2314, opened 2018-07-02
Efficient reading

clusterfudge:
Avoid using streaming read where possible.


webapp/graphite/readers/remote.py
```diff
@@ -106,3 +111,65 @@ def fetch_multi(self, startTime, endTime, now=None, requestContext=None):
...
+    # Here be Snap monkey-patch code
+    #
+    def deserialize(self, result):
+        # avoid buffered reader if possible, instead using in-memory unpacking for small->REMOTE_BUFFER_SIZE payloads
+        try:
+            start = time.time()
+            measured_reader = MeasuredReader(BufferedHTTPReader(result, settings.REMOTE_BUFFER_SIZE))
+            first_chunk = self._fill_first_buffer(measured_reader, settings.REMOTE_BUFFER_SIZE)
+            if len(first_chunk) < settings.REMOTE_BUFFER_SIZE:
+                log.info("Using inline deserializer for small payload")
```
deniszh: I think it should be `log.debug` - here, and in other places too.

deniszh:
@clusterfudge @DanCech @iksaif - what we going to do with this change?
Reminding that is related to the problem described in https://github.com/graphite-project/graphite-web/issues/2313#issuecomment-401876110
I also notice some performance degradation on big requests but I can't prove it with numbers, though.
OTOH I don't really like consequences of mandatory REMOTE_BUFFER_SIZE. Can we probably make avoiding streaming deserialization optional somehow? I.e if REMOTE_BUFFER_SIZE > 0 use it, if not - fallback to stream implementation?

iksaif:
I agree with making REMOTE_BUFFER_SIZE optional

clusterfudge:
@deniszh yeah, I like the idea of 0 to disable (vs using an absurdly high number to avoid). I'll refresh this PR to reflect.

deniszh:
Hi @clusterfudge ,
Could you please fix this PR soon, if you want that change still be included in upcoming 1.1.4 release. Thanks!

clusterfudge:
Hey @deniszh , I spent a bit of time polishing yesterday and should be able to get it out tonight. What's the target cutoff for 1.1.4?

deniszh:
Soon ™️ :)

deniszh:
@clusterfudge : OK, you mentioned before that if request size is more then REMOTE_BUFFER then it's almost guaranteed to fail, but I'm not really getting why. Is it still true?
If yes - then we need to inform users about changing behavior.

clusterfudge:
@deniszh That was an artifact of the prior implementation and is no longer relevant. In this implementation, we use the streaming deserializers if we have REMOTE_BUFFER_SIZE > 0, otherwise we use the inline deserialization. This does not change the behavior of existing users, but allows people to opt-out of streaming serialization to increase performance on large payloads.
TL/DR; that comment is no longer relevant.

deniszh:
Ah, that's what I suspect - but thanks for the clarification.
Merging.

Quentin-M:
This PR appears to break metrics discovery: https://github.com/graphite-project/graphite-web/issues/2352.

PR #2317, opened 2018-07-11
html-encode text passed to Ext.Msg

DanCech:
`Ext.Msg.alert` doesn't automatically escape HTML special characters in the message, which can lead to display issues and could be a potential area for XSS.

The new `htmlEncode` function added errs on the side of caution by encoding all characters other than alphanumeric and space.


deniszh:
@cbowman0 - could you please take a peek on this PR, please? Thanks a lot!

Issue #2318, opened 2018-07-17
graphite-web 1.1.3 render with multiple targets returns null or no data


TimWhalen:
OS: centos-7
python: 2.7.5
django: 1.11
graphite-web: 1.1.3
rrdtool: 1.7.0

There are three queries in the examples below.  Two are raw data from rrds and the third is a divideSeries of the first two.  When you request the raw data or the divideSeries you get the
correct results.  If you request both raw data and the divideSeries you get null values or no
datapoints depending on the order of targets:

1. raw data
`curl 'http://xvba-e25.xv.dc.openx.org/graphite/render?format=json&from=1530050400&until=1530051000&target=market_opportunity_svc.md.internal_qualify_prequalify_micros.host.xvbd-e62.counter.value&target=market_opportunity_svc.md.get_opportunities_success.host.xvbd-e62.counter.value'`
```
[{
	"datapoints": [
		[9054705.109444445, 1530050400],
		[9120871.425, 1530050460],
		[9177432.971388888, 1530050520],
		[9045113.327777779, 1530050580],
		[9172009.116666667, 1530050640],
		[9072210.63888889, 1530050700],
		[9069803.035555555, 1530050760],
		[9143143.27388889, 1530050820],
		[9131529.015555555, 1530050880],
		[9077973.385833332, 1530050940]
	],
	"target": "market_opportunity_svc.md.internal_qualify_prequalify_micros.host.xvbd-e62.counter.value",
	"tags": {
		"name": "market_opportunity_svc.md.internal_qualify_prequalify_micros.host.xvbd-e62.counter.value"
	}
}, {
	"datapoints": [
		[7462.465277777777, 1530050400],
		[7626.964166666668, 1530050460],
		[7703.191944444445, 1530050520],
		[7513.552500000001, 1530050580],
		[7721.190833333332, 1530050640],
		[7547.9766666666665, 1530050700],
		[7571.581944444444, 1530050760],
		[7595.705277777778, 1530050820],
		[7630.304722222223, 1530050880],
		[7562.694444444443, 1530050940]
	],
	"target": "market_opportunity_svc.md.get_opportunities_success.host.xvbd-e62.counter.value",
	"tags": {
		"name": "market_opportunity_svc.md.get_opportunities_success.host.xvbd-e62.counter.value"
	}
}]
```

2. divideSeries
[...]

PR #2322, opened 2018-07-25
convert prefetched values generator to reusable iterator

TimWhalen:
One way to address the issue I opened last week: https://github.com/graphite-project/graphite-web/issues/2318

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -16,6 +16,7 @@ from __future__ import division
 import collections
 import re
 import time
+import types
 from six import text_type
 
 from django.conf import settings
@@ -270,6 +271,13 @@ def prefetchData(requestContext, pathExpressions):
         result['values'],
       ),
     ))
+
+  # Several third-party readers including rrdtool and biggraphite return values in a
+  # generator which can only be iterated on once. These must be converted to a list.
+  for pathExpression, items in prefetched.items():
+    for i, (name, (time_info, values)) in enumerate(items):
+      if isinstance(values, types.GeneratorType):
+        prefetched[pathExpression][i] = (name, (time_info, list(values)))
 
   if not requestContext.get('prefetched'):
     requestContext['prefetched'] = {}


DanCech:
Need to dig into this a bit and make sure there's no performance regression

iksaif:
I'm pretty sure this *will* have performance regressions. Doing it in two loops (first getting all the generators, then expanding them) would maybe work.

TimWhalen:
Thanks, I'll move the generator expansion to render.evaluate.evaluateTargets and
I'll try to add a meaningful test too.  If only the rrdtool backend is returning values in
a generator maybe it'd be better to address this there?

iksaif:
FYI: It's not only RRD, it's also a bunch for third-party readers (mostly biggraphite). I won't have a stable  internet connection in the next three weeks so I can't validate your patch, but it looks like it would work (maybe add some code comments to explain why it is done this way).

Thanks

deniszh:
Is it OK to merge that? @DanCech @iksaif ?

iksaif:
I think it's still missing comments to explain why it is done this way. Else it's likely that the code will be changed again at some point.

TimWhalen:
I've added a brief comment explaining the change.

webapp/graphite/render/datalib.py
```diff
@@ -271,6 +272,13 @@ def prefetchData(requestContext, pathExpressions):
       ),
     ))
 
+  # Several third-party readers including rrdtool and biggraphite return values in a
+  # generator which can only be iterated on once. These must be converted to a list.
+  for pathExpression, items in prefetched.items():
+    for i, (name, (time_info, values)) in enumerate(items):
+      if isinstance(result['values'], types.GeneratorType):
```
DanCech: where is `result['values']` coming from? assuming this should be `values`

TimWhalen: You are correct, that was a mistake introduced when I moved the check into its own loop to address an earlier change request.  Sorry about that.  I've updated the pr again.

PR #2324, opened 2018-07-31
Fix get_real_metric_path for paths where an intermediate directory is a symlink

yadsirhc:
Take 2


yadsirhc:
@deniszh  No dice.  From everything else I'm reading it looks like someone from the graphite project will need to engage Travis CI support.

Issue #2320, opened 2018-07-23
get_real_metric_path still does not properly handle some symlinks


yadsirhc:
* What I attempted: /home/graphite-web/storage/whisper is a symbolic link pointing to /path/foo/whisper and /path/foo/whisper/collectd is a symbolic link pointing to /path/foo/collectd, which contains host metrics
* Expected Result: all whisper files are found after looking up real metric path
* Witnessed Result: no whisper files are found after looking up real metric path because the beginning of the real metric path gets truncated
* System Details:
  * OS / Version: Ubuntu 16.04.2 LTS
  * Python Version: 2.7.12
  * Django Version: 1.8.7
  * graphite-web Version: 1.1.3

Related to #1738  and #2012

yadsirhc:
Still working on test case updates, but they have been a bit painful to run in my environment compared to the test harness I had for rapid prototyping of the fix.

PR #2326, opened 2018-07-31
Fix get_real_metric_path for paths where an intermediate directory is a symlink 

yadsirhc:
Take 4, last try for now.  I was informed that I needed an active Travis CI account w/ the same username as my GitHub account for the Travis CI PR build to occur successfully. I'm skeptical, but I've linked my accounts via OAuth and will see if the results differ afterward now.

### DIFF
--- a/webapp/graphite/finders/__init__.py
+++ b/webapp/graphite/finders/__init__.py
@@ -11,8 +11,8 @@ def get_real_metric_path(absolute_path, metric_path):
     if absolute_path != real_absolute_path:
         # replace left side base_fs_path that contains sym link with real fs path
         relative_fs_path = metric_path.replace('.', os.sep)
-        absolute_path_no_ext, _ext = os.path.splitext(absolute_path)
-        base_fs_path = os.path.dirname(absolute_path_no_ext[:-len(relative_fs_path)])
+        real_absolute_path_no_ext, _ext = os.path.splitext(real_absolute_path)
+        base_fs_path = os.path.dirname(real_absolute_path_no_ext[:-len(relative_fs_path)])
         real_base_fs_path = os.path.realpath(base_fs_path)
         real_relative_fs_path = real_absolute_path[len(real_base_fs_path):].lstrip(os.sep)
         return fs_to_metric(real_relative_fs_path)
--- a/webapp/tests/test_finders.py
+++ b/webapp/tests/test_finders.py
@@ -302,16 +302,27 @@ class StandardFinderTest(TestCase):
     def dummy_realpath(path):
         return path.replace("some/symbolic/path", "this/is/the/real/path")
 
+    def dummy_realpath2(path):
+        return path.replace("foo/bar", "bar")
+
     @patch('os.path.realpath', wraps=dummy_realpath)
-    def test_get_real_metric_path_symlink_outside(self, dummy_realpath):
+    def test_get_real_metric_path_symlink_before(self, dummy_realpath):
         input_abs_path='/some/symbolic/path/graphite/whisper/Env/HTTP/NumConnections.wsp'
         input_metric_path='Env.HTTP.NumConnections'
         expected_metric_path='Env.HTTP.NumConnections'
         output_metric_path = get_real_metric_path(input_abs_path, input_metric_path)
         self.assertEqual(output_metric_path, expected_metric_path)
 
+    @patch('os.path.realpath', wraps=dummy_realpath2)
+    def test_get_real_metric_path_symlink_inside(self, dummy_realpath2):
+        input_abs_path='/opt/graphite/storage/whisper/foo/bar/Env/HTTP/NumConnections.wsp'
+        input_metric_path='bar.Env.HTTP.NumConnections'
+        expected_metric_path='bar.Env.HTTP.NumConnections'
+        output_metric_path = get_real_metric_path(input_abs_path, input_metric_path)
+        self.assertEqual(output_metric_path, expected_metric_path)
+
     @patch('os.path.realpath', wraps=dummy_realpath)
-    def test_get_real_metric_path_symlink_inside(self, dummy_realpath):
+    def test_get_real_metric_path_symlink_after(self, dummy_realpath):
         input_abs_path='/opt/graphite/storage/whisper/some/symbolic/path/NumConnections.wsp'
         input_metric_path='some.symbolic.path.NumConnections'
         expected_metric_path='this.is.the.real.path.NumConnections'


yadsirhc:
@deniszh It finally went through successfully on Travis CI after this most recent PR attempt.

yadsirhc:
Fix for #2320

deniszh:
Thanks! But I will have PC and stable internet in a week or so... Maybe be
@DanCech could check?

deniszh:
Can I get more eyes on this PR, pretty please? @DanCech @iksaif @cbowman0 @piotr1212 ?
Thanks!

bwoebi:
@deniszh Why did you merge that?

The tests are just passing by chance; e.g. if the symlink name would be much longer, then the tests would break.

deniszh:
@bwoebi : please check https://github.com/graphite-project/graphite-web/issues/2320 as original issue

bwoebi:
Example:

```
metric_path = "foo.bar"
absolute_path = "/graphite/storage/whisper/foo/bar"
real_absolute_path = "/graphite/storage/whisper/foo/somethingTotallyDifferent"
# i.e. the path at absolute_path symlinks to real_absolute_path
```

Resolves to `somethingTotallyDifferent` instead of `foo.somethingTotallyDifferent`.

Without this patch, this works.

For the original issue (#2320) - I don't think you can fix this at all. In his original example, he has:
/home/graphite-web/storage/whisper -> /path/foo/whisper
/path/foo/whisper/collectd -> /path/foo/collectd

So, he is moving with his symbolic link _outside_ the databases root tree (i.e. /path/foo/whisper). How can you properly compare the roots?
If I understood the purpose of symlink resolving correctly, this is to have the right resolved metric named to pass to carbon-cache. So, we have to assume that the carbon-cache metrics root is identical to the root we see here (the whisper folder). But as he is moving outside of the whisper folder, so this just cannot work out.

I mean, given his example, how do you know whether his expected metric is `collectd`, `foo.collectd` or `path.foo.collectd`? You do not have any root path to compare it to...

For me the symlink resolving has one invariant: symlinks within the metrics root must not point to locations outside of the metrics root and he is violating that.

FliegenKLATSCH:
@deniszh @iksaif  Any news on this topic? Would be nice if this could be reverted.

deniszh:
@FliegenKLATSCH : why it should be reverted?

FliegenKLATSCH:
Let me try to find different words for @bwoebi example:
If we have a symlink:
`/graphite/storage/whisper/foo/bar -> /graphite/storage/whisper/foo/somethingTotallyDifferent `
With this patch, the resolved metrics path is `somethingTotallyDifferent` whereas it should actually resolve to `foo.somethingTotallyDifferent`. And therefore none of the requests to `foo.bar` metrics (the symlinked ones) will have any cache hit.

FliegenKLATSCH:
Ping @deniszh

deniszh:
Hello @FliegenKLATSCH,
I understood @bwoebi logic, but that didn't answer my question. I still think that fixes need to be done not because of formal correctness, but for fixing user's issues. That's why I'm asking "issue with a bug description".
@yadsirhc had  problem, he made issue, made PR, make tests - he did that because he had problem to solve. He made mistake? Totally possible, but he solved his issue.
Do we have problem to solve here? Then we need at least proper issue, which is definitely not this one.
Thanks for understanding. If other maintainers disagree - I will not be against reverting this change, but that's not how I prefer to do changes, so, I would not do that.

ploxiln:
agreed - if you want a change, please open a pull request, with a full and self-contained description of the issue

deniszh:
@ploxiln : PR is not required, not all users of software are capable of produce one, but [bug] issue should be enough.

piotr1212:
Current (master/head) behaviour is broken. I tried with a symlink pointing within the root and one pointing outside the root, both giving clearly incorrect results (broken/half paths like /opt/graphite/some/metric -> hite/some/metric).

I need to do some more testing but I am quite confident to say that cache lookups for metrics which have a symlink in them will fail.

I don't see an easy way to do this correctly for metrics which are outside the root. One way could be to define multiple root's in config or just don't do cache queries for metrics outside the root (and log it).

FliegenKLATSCH:
@piotr1212 Thanks. Try without this patch and symlink within the root will work again.
Symlinks outside of the root are a different story (I don't know if that is a valid use case).

piotr1212:
`STORAGE_DIR = '/opt/graphite/storage'`


```
$ tree /opt/graphite/
/opt/graphite/
├── outside_storage
│   └── metric.wsp
└── storage
    └── whisper
        ├── carbon
        ├── other
        │   ├── link_inside -> ../some/
        │   └── link_outside -> /opt/graphite/outside_storage/
        └── some
            └── metric.wsp

```

web request for other.link_inside.metric results in a cache query:
head: `cache INFO CarbonLink sending request for storage.whisper.some.metric to ('127.0.0.1', None)`
old: `cache INFO CarbonLink sending request for some.metric to ('127.0.0.1', None)`

Old is correct, carbon-cache does not know about the symlink, only has cache for `some.metric`


web request for other.link_outside.metric:
haed: `cache INFO CarbonLink sending request for graphite.outside_storage.metric to ('127.0.0.1', None)`
old: `cache INFO CarbonLink sending request for metric to ('127.0.0.1', None)`

Should be ??

Trying to reproduce original bugreport, created:
```
$ tree /opt/graphite/
/opt/graphite/
├── outside_storage
│   ├── metric.wsp
│   └── outside_storage2 -> /opt/graphite//outside_storage2
├── outside_storage2
│   ├── metric.wsp
│   └── more_levels
│       └── outside_3 -> ../../outside_storage3
├── outside_storage3
│   └── more_dirs
│       └── metric.wsp
└── storage
    └── whisper
        ├── other
        │   ├── link_inside -> ../some/
        │   └── link_outside -> /opt/graphite/outside_storage/
        └── some
            └── metric.wsp
```
I can see all metrics. So I can't reproduce the old bugreport

metrics

conclusion: I cannot reproduce the old bug before the patch, the patch breaks cache lookups for metrics with symlinks. Maybe @yadsirhc could comment?

ploxiln:
I think there's agreement that this PR is wrong, while recommending that a new issue be opened, instead of continuing to comment on a closed PR.

yadsirhc:
I no longer have access to the environment where I was testing these changes, but it's clear that there are corner cases which were not considered.

> For me the symlink resolving has one invariant: symlinks within the metrics root must not point to locations outside of the metrics root and he is violating that.

If that's a true invariant, it's one which I was not aware of and never came up along the way, and for some storage appliances, it's not necessarily viable.

yadsirhc:
@pfree, you may be interested in picking this back up.

deniszh:
Thanks @piotr1212 for spending time on that!
@yadsirhc @bwoebi @FliegenKLATSCH : this PR is reverted now.

bwoebi:
Thanks! @deniszh

PR #2331, opened 2018-08-14
Latest whitenoise is not supported by graphite

ellisvlad:
This is a temporary fix to use an older version


deniszh:
Going to merge this to fix master

Issue #2332, opened 2018-08-17
graphite-web might be incompatible with whitenoise 4


ybizeul:
```
[Fri Aug 17 15:44:30.702758 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300] mod_wsgi (pid=20006): Target WSGI script '/opt/graphite/conf/graphite.wsgi' cannot be loaded as Python module.
[Fri Aug 17 15:44:30.702793 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300] mod_wsgi (pid=20006): Exception occurred processing WSGI script '/opt/graphite/conf/graphite.wsgi'.
[Fri Aug 17 15:44:30.702813 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300] Traceback (most recent call last):
[Fri Aug 17 15:44:30.702840 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300]   File "/opt/graphite/conf/graphite.wsgi", line 4, in <module>
[Fri Aug 17 15:44:30.702886 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300]     from graphite.wsgi import application
[Fri Aug 17 15:44:30.702894 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300]   File "/opt/graphite/webapp/graphite/wsgi.py", line 29, in <module>
[Fri Aug 17 15:44:30.702907 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300]     from whitenoise.django import DjangoWhiteNoise
[Fri Aug 17 15:44:30.702914 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300]   File "/usr/local/lib/python2.7/dist-packages/whitenoise/django.py", line 2, in <module>
[Fri Aug 17 15:44:30.702925 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300]     '\\n\\n'
[Fri Aug 17 15:44:30.702942 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300] ImportError: 
[Fri Aug 17 15:44:30.702945 2018] [wsgi:error] [pid 20006] [remote 127.0.0.1:21300] 
[...]

PR #2333, opened 2018-08-18
Add support for WhiteNoise 4

piotr1212:
https://github.com/graphite-project/graphite-web/issues/2332
https://github.com/graphite-project/graphite-web/pull/2331

This should fix support for WhiteNoise 4. Note that you'll need Django > 1.10 (I think).

I've only tested with gunicorn. Had to rerun manage.py collectstatic.


deniszh:
Why Django 1.10 is required? Is it needed for whitenoise 4 ?

piotr1212:
Rechecked and seems to work with Django 1.8. Not sure what I did wrong before..

Issue #2337, opened 2018-08-22
Query crashes for avg_zero aggregation method


eachirei:
What you attempted: Hitcount query
Expected Result: Successful result
Witnessed Result: Error described below
System Details
OS / Version: Docker container from ubuntu:14.04
Python Version: 2.7.6
Django Version: 1.11.10
Graphite Component Versions: 1.1.3

When querying for data sometimes the graphite request returns the following error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/exception.py", line 41, in inner
    response = get_response(request)
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 187, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 185, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/var/lib/graphite/webapp/graphite/render/views.py", line 117, in renderView
    data.extend(evaluateTarget(requestContext, targets))
  File "/var/lib/graphite/webapp/graphite/render/evaluator.py", line 15, in evaluateTarget
    prefetchData(requestContext, pathExpressions)
  File "/var/lib/graphite/webapp/graphite/render/datalib.py", line 262, in prefetchData
    for result in STORE.fetch(pathExpressions, startTime, endTime, now, requestContext):
  File "/var/lib/graphite/webapp/graphite/storage.py", line 179, in fetch
    'fetch for %s' % str(patterns))
  File "/var/lib/graphite/webapp/graphite/storage.py", line 132, in wait_jobs
    raise Exception(message)
Exception: All requests failed for fetch for [u'stats.request.*.all'] (1)

fetch for [u'stats.request.*.all']: Invalid consolidation function: 'avg_zero': Traceback (most recent call last):

  File "/var/lib/graphite/webapp/graphite/worker_pool/pool.py", line 43, in run
    self.result = self.func(*self.args, **self.kwargs)

  File "/var/lib/graphite/webapp/graphite/finders/utils.py", line 140, in fetch
    now=now, requestContext=requestContext

  File "/var/lib/graphite/webapp/graphite/node.py", line 30, in fetch
    result = self.reader.fetch(startTime, endTime, now, requestContext)

  File "/var/lib/graphite/webapp/graphite/readers/whisper.py", line 74, in fetch
[...]

PR #2338, opened 2018-08-23
 - added avg_zero support for consolidate function

eachirei:
https://github.com/graphite-project/graphite-web/issues/2337

### DIFF
--- a/webapp/graphite/readers/utils.py
+++ b/webapp/graphite/readers/utils.py
@@ -37,6 +37,11 @@ def merge_with_cache(cached_datapoints, start, step, values, func=None, raw_step
 
     # Similar to the function in render/datalib:TimeSeries
     def consolidate(func, values):
+        if func == 'avg_zero':
+            zero_usable = [0 if v is None else v for v in values]
+            if not zero_usable:
+                return None
+            return float(sum(zero_usable)) / len(zero_usable)
         usable = [v for v in values if v is not None]
         if not usable:
             return None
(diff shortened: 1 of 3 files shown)


piotr1212:
lgtm, Could you please add it to the tests? somewhere here https://github.com/graphite-project/graphite-web/blob/master/webapp/tests/test_functions.py#L1443-L1449

deniszh:
Thanks, @eachirei !
But please add tests, as @piotr1212 said.

eachirei:
Done! Not sure if I should add one of these as well: https://github.com/graphite-project/graphite-web/blob/824800a4218c8cb9ab56ca67277b0f7774a7ff1d/webapp/tests/test_readers_util.py#L179-L210

DanCech:
This looks good, we should add `avg_zero` to the `consolidateBy` function docs and also add it to `aggFuncs` in `functions.py` so that it can be used consistently throughout the system.

DanCech:
On further review, this only adds support for `merge_with_cache` consolidation at read time for whisper files using `avg_zero` consolidation.

To support `avg_zero` in `consolidateBy` it would need to be added to `__consolidatingGenerator` in `TimeSeries` which will take some effort as that is geared entirely toward discarding `None` values.  The test passes because it just verifies that `series.consolidationFunc` is set, not that it actually functions.

To support `avg_zero` in functions that perform consolidation is possible, and doesn't require too much work.

At this point I'd accept this as a bugfix if the change in `webapp/tests/test_functions.py` is removed, then support for `avg_zero` in `consolidateBy` and the aggregation functions could be added separately.

deniszh:
Superceded by https://github.com/graphite-project/graphite-web/pull/2380

PR #2342, opened 2018-08-28
Adds the missing avg_zero consolidation method

ArneBachmann:


DanCech:
There is an existing PR for this #2338, see the commentary there for what else is needed if you want to expand this PR

PR #2344, opened 2018-08-29
[1.1.x] Allow history tracking in composer | replace raise StopIteration with return pep-0479 | carbonlink: set the type of the recv buffer explicitly to bytes | clarify 'maxDataPoints' | Fail Hard p1 | Fail 

deniszh:
 - Allow history tracking in composer (#2304)
 - replace raise StopIteration with return pep-0479 (#2300)
 - carbonlink: set the type of the recv buffer explicitly to bytes (#2301)
 - clarify 'maxDataPoints' (#2302)
 - Fail Hard p1 (#2303)
 - Fail Hard p2 (#2303)
 - Fail Hard p3 (#2303)
 - Typo (#2303)
 - Comments from @deniszh (#2303)
 - adding unit tests (#2) (#2303)
 - Commented out default should match default in settings.py (#2303)
 - Naming suggesting from DanCech. (#2303)
 - remove unused variable (#2304)
 - Fail if any job does not yield a result (#2303)


deniszh:
Part 2 of 1.1.4 backport

deniszh:
will merge that, because part 3 is not mergeable until I merge this to 1.1.x

PR #2350, opened 2018-09-06
Do not use generator for jobs need to be run in pool

fengyehong:
As generators are lazy, in context of pool exceution it will return
immediately and exceptions can not be caught as expected.

### DIFF
--- a/webapp/graphite/finders/remote.py
+++ b/webapp/graphite/finders/remote.py
@@ -124,6 +124,11 @@ class RemoteFinder(BaseFinder):
 
             cache.set(cacheKey, results, settings.FIND_CACHE_DURATION)
 
+        # We don't use generator here, this function may be run as a job in a thread pool, using a generator has the following risks:
+        # 1. Generators are lazy, if we don't iterator the returned generator in the job, the real execution(network operations,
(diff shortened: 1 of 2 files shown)


fengyehong:
I was testing the `STORE_FAIL_ON_ERROR = False` flag, I have 4 storage nodes(with graphite-web and carbon running), one relay node, and a separate master graphite-web node.
If i shutdown one of the storage node, and query the master graphite-web `/metrics/find?query=*`, I still get HTTP 500 returned, which body contains the detail: connection refused to the storage ndoe.
Depending on the `REMOTE_RETRY_DELAY` and `FIND_CACHE_DURATION` parameters, refresh may return successfully, but after the retry dealy or cache duration, I get HTTP 500 again.

I check the code and find that as RemoteFinder.find_nodes is a generator, exceptions in this function can not be found by Store.wait_jobs

DanCech:
This seems reasonable to me

iksaif:
Please add a comment in the code explaining why you don 't use a generator, else in ~6 month somebody will try to switch back to a generator without knowing why it wasn't one in the first place :)

fengyehong:
updated

iksaif:
👍

Issue #2352, opened 2018-09-10
`REMOTE_BUFFER_SIZE = 0` ==> Metrics discovery broken


Quentin-M:
Hi,

We recently upgraded our cluster to 1.1.4 in order to profit from https://github.com/graphite-project/graphite-web/pull/2314 and thus work around the potential 500x slowdown (was it not tested?) described in https://github.com/graphite-project/graphite-web/issues/2313#issuecomment-401876110. We then set `REMOTE_BUFFER_SIZE = 0` as documented.

But now, our clusters (populated or empty) fail to do metrics discovery for the most part. Here is an example:

```$ curl 'localhost:8080/metrics/find?from=1536617499&query=*&until=1536618399'

<body style="background-color: #666666; color: black;">
<center>
<h2 style='font-family: "Arial", sans-serif'>
<p>Graphite encountered an unexpected error while handling your request.</p>
<p>Please contact your site administrator if the problem persists.</p>
</h2>
<br/>
<div style="width: 50%; text-align: center; font-family: monospace; background-color: black; font-weight: bold; color: #ff4422;">

</div>

<div style="width: 70%; text-align: left; background-color: black; color: #44ff22; border: thin solid gray;">
<pre>
Traceback (most recent call last):
  File &quot;/opt/graphite/webapp/graphite/finders/remote.py&quot;, line 112, in find_nodes
    result, buffer_size=settings.REMOTE_BUFFER_SIZE), encoding=&#39;utf-8&#39;)
  File &quot;/opt/graphite/webapp/graphite/umsgpack.py&quot;, line 860, in _unpack3
    return _unpack(fp, options)
  File &quot;/opt/graphite/webapp/graphite/umsgpack.py&quot;, line 773, in _unpack
    code = _read_except(fp, 1)
  File &quot;/opt/graphite/webapp/graphite/umsgpack.py&quot;, line 589, in _read_except
    raise InsufficientDataException()
graphite.umsgpack.InsufficientDataException

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File &quot;/usr/local/lib/python3.6/site-packages/django/core/handlers/exception.py&quot;, line 35, in inner
    response = get_response(request)
  File &quot;/usr/local/lib/python3.6/site-packages/django/core/handlers/base.py&quot;, line 128, in _get_response
    response = self.process_exception_by_middleware(e, request)
[...]

PR #2354, opened 2018-09-11
move deserialize into remote finder

DanCech:
This addresses the issue outlined in #2352 by moving the `deserialize` functions into the remote finder and using them in `find_nodes`.

While this code passes all tests I have not actually tried it yet and would appreciate feedback from anyone who has time to experiment with it.

### DIFF
--- a/webapp/graphite/finders/remote.py
+++ b/webapp/graphite/finders/remote.py
@@ -106,21 +106,7 @@ class RemoteFinder(BaseFinder):
                 headers=query.headers,
                 timeout=settings.FIND_TIMEOUT)
 
-            try:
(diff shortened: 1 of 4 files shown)


deniszh:
just FYI: trying to create simple docker compose to test cluster from 3 nodes

DanCech:
The tags stuff also uses the finder `request` method, so I'm not going to try and shoehorn the de-serialization in there also at the moment.

DanCech:
As far as I can tell this is working correctly, any objections to merging it @deniszh ?

deniszh:
No objections. :)

PR #2357, opened 2018-09-17
Code and unit test to support non-streaming serialization on RemoteFinder

clusterfudge:
#2314 did not properly cover RemoteFinder. This change adds a test and the associated code for toggling between streaming and in-memory deserialization


deniszh:
@clusterfudge : you fixing only find/remote.py here, right? And render/remote.py  will still have old code?
IMO we should did it in similar way as @DanCech did in #2354, but probably move deserializer in some shared file - user_util.py or something else.

DanCech:
Superseded by #2354

PR #2330, opened 2018-08-13
Multi-instance settings

zasca:
#2329

Issue #2329, opened 2018-08-13
Multiple instances on one host


zasca:
In some cases, there is a need to run multiple graphite instances with different configurations on the same host. For example, there are two graphite endpoints *:8080, *:8081. From the first endpoint should be access to all metrics, but from the second endpoint should be limited access only to some metrics from certain sub directory.

Graphite does not have multi-instance support out of the box. And in this case, you need to create multiple environments with `virtualenv` or duplicate graphite directory with all source files by hand, while only the different settings are required. I think that such a decision is not optimal.

I suggest:
 1. to pass settings module name via certain environment variable in wsgi file
 2. modify the settings module so it first would be tries to load a custom settings file

PR #2358, opened 2018-09-17
Multi-instance settings support [light version]

zasca:
Alternate PR version for #2329 without architectural changes.
The first version is #2330

### DIFF
--- a/webapp/graphite/settings.py
+++ b/webapp/graphite/settings.py
@@ -207,7 +207,7 @@ DATABASES = None
 FLUSHRRDCACHED = ''
 
 ## Load our local_settings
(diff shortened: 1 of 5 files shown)


webapp/graphite/settings.py
```diff
@@ -206,10 +207,11 @@
 FLUSHRRDCACHED = ''
 
 ## Load our local_settings
+SETTINGS_MODULE = os.environ['GRAPHITE_SETTINGS_MODULE']
```
deniszh: You should use `get()` or wrap up this in try...except - it breaks tests currently.

zasca: Thanks, Denis.
Fixed.

PR #2368, opened 2018-10-19
Increase performance of localdatabase tag db

DanCech:
This PR updates the local (mysql, postgres, sqlite) tagdb to use the hash of the series when locating records, which provides a significant performance boost by supporting use of the unique index on `tags_series.hash`.


deniszh:
Did I understand correctly that it will be not compatible with previous version? Should we provide migration script then?

deniszh:
Ah, right, hash was there, but not used for selecting. Cool, looks good then. :)

Issue #2369, opened 2018-10-29
carbonlink query doesn't work for tagged series


andreasferber:
When fetching data from a tagged series, graphite-web as evidenced by the carbon-cache query.log uses the encoded series path (something like `_tagged.301.220.3012209b6ac3b582a8fd1a3317a5f40407fa446943af8f24ed32beff3f13b620` in my case with `TAG_HASH_FILENAMES=True`) to query the carbon-cache for in-memory data. However, AFAICT carbon-cache organizes its memory cache not by the encoded path, but by the path it received from the metric sender (after normalizing it, ie. sorting the tags and converting OpenMetrics tag syntax to Carbon syntax), something like `my.metric;bar=a;foo=b`. Hashing to get the filename is only done when data is finally written to disk. Therefore, the carbonlink query will fail to retrieve any data currently floating in the carbon-cache.

In theory this could be fixed on either end by changing the behaviour to match the other end. However, if you were to change this in carbon-cache, you would run into a related issue, namely that the relay doesn't take tagging into account when consistent-hashing (I'm going to open another bug report on carbon for that). Since it is essential that the hash used for consistent-hashing is the same in both graphite-web and carbon-relay, changing carbon-cache would mean that carbon-relay would also have to be aware of the TAG_HASH_FILENAMES setting for one thing, and it would have to calculate the final encoded path for the series in order to decide to which carbon-cache the metric should be sent.

OTOH, fixing it in graphite-web should be straightforward, since at the point where carbonlink is queried the human-friendly name of the series must have already been queried from the TAGDB (otherwise the series wouldn't have been found in the first place).

deniszh:
Hi @andreasferber ,
Valid concern, indeed.

andreasferber:
As suspected, the fix was pretty simple indeed, PR coming up shortly.

PR #2370, opened 2018-10-29
Fix carbonlink queries for tagged series.

andreasferber:
Fixes graphite-project/graphite-web#2369


andreasferber:
I don't really understand why the TOXENV=lint check failed, none of the files mentioned in the build log were even touched by me. Looks to me like an issue with the check itself, although other PRs seem to pass it...

DanCech:
That may be due to a flake8 update, I'll check it out

PR #2373, opened 2018-10-30
fix static content directory in graphite-vhost.conf

uwemaurer:


deniszh:
Hello @uwemaurer ,

You should run "collectstatic" command from Django to put files in /opt/graphite/static

uwemaurer:
Thanks for the tip.  This should be mentioned much more prominently in the documentation, since it doesn't work without running this.

https://graphite.readthedocs.io/en/latest/config-local-settings.html?highlight=collectstatic#filesystem-paths

It worked with these commands:

pip  install django-admin
PYTHONPATH=/opt/graphite/webapp django-admin.py collectstatic --noinput --settings=graphite.settings

PR #2380, opened 2018-11-05
Consolidation support avg zero

DanCech:
This is a continuation of #2338 which adds support in consolidateBy and aggregation functions, as well as making the behavior consistent with whisper when an interval has no non-null datapoints (it aggregates to None).

I have not yet had time to write tests, but we should have some before we merge this.


deniszh:
Looks good! but should avg_zero be mentioned somewhere in docs?

deniszh:
@piotr1212 : could you please check this PR, if possible? Thanks!

PR #2383, opened 2018-11-09
RRD path fix and tests

piotr1212:
Fix from https://github.com/graphite-project/graphite-web/pull/2377 and added tests.

### DIFF
--- a/webapp/graphite/readers/rrd.py
+++ b/webapp/graphite/readers/rrd.py
@@ -1,7 +1,6 @@
 import sys
 import os
 import time
-import six
 
 # Use the built-in version of scandir/stat if possible, otherwise
 # use the scandir module version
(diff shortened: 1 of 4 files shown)


webapp/graphite/readers/rrd.py
```diff
@@ -25,8 +24,12 @@ class RRDReader(BaseReader):
 
     @staticmethod
     def _convert_fs_path(fs_path):
-        if isinstance(fs_path, six.text_type):
-            fs_path = fs_path.encode(sys.getfilesystemencoding())
+        # Only Python 2 'unioode' needs to be converted to str/bytes
```
DanCech:
```suggestion
        # Only Python 2 'unicode' needs to be converted to str/bytes
```

deniszh:
Thanks a lot, @piotr1212 ! Merging this.

PR #2409, opened 2019-01-08
Remove /opt/graphite prefix and use setuptools

piotr1212:
See https://github.com/graphite-project/carbon/pull/835 for reasoning


webapp/graphite/settings.py
```diff
@@ -35,9 +35,9 @@
 DATE_FORMAT = '%m/%d'
 
 # Filesystem layout
-WEB_DIR = dirname( abspath(__file__) )
+WEB_DIR = dirname(abspath(__file__))
 WEBAPP_DIR = dirname(WEB_DIR)
-GRAPHITE_ROOT = dirname(WEBAPP_DIR)
+GRAPHITE_ROOT = '/opt/graphite'
```
piotr1212: This should be changed to prefix, Ignore this commit, I'll change/rebase later.

piotr1212:
I'm not finished yet but made some changes to the docs which I would like to get some feedback for. I've tried to simplify the docs, IMO there were too much separate pages which you had to jump back and forward to.

I've changed the default install (settings.py) so that running collectstatic is not needed. The static files can be served directly from the app with whitenoise. Serving from whitenoise should be fast enough for most installations. this eliminates the need for configuring the static dir in the webserver (simplifies installation). From what I've read the whole purpose of collectstatic is for organisations which run multiple django apps and have separated their static files from code (in repo), so that they can update static files without having to deploy code and vice versa. As graphite's static files haven't changed in years and they are in the code repo I don't see a point to require collectstatic in the default install. Users can still run collectstatic if they want/need.

please ignore the GRAPHITE_ROOT commit, ill remove it later.

webapp/graphite/local_settings.py.example
```diff
@@ -94,6 +94,28 @@ DEFAULT_XFILES_FACTOR = 0
 #FIND_TIMEOUT = 3.0  # Timeout for metric find requests
 #FETCH_TIMEOUT = 3.0  # Timeout to fetch series data
 
+
+#####################################
+# Static Files #
+#####################################
+#
+# Graphite servers static files (.html, .css, .js) directly from the
```
ploxiln: "servers" -> "serves"

piotr1212: thanx

setup.py
```diff
@@ -1,93 +1,38 @@
...
 
 storage_dirs = []
 
-for subdir in ('whisper/dummy.txt', 'ceres/dummy.txt', 'rrd/dummy.txt', 'log/dummy.txt', 'log/webapp/dummy.txt'):
-  storage_dirs.append( ('storage/%s' % subdir, []) )
-
-webapp_content = defaultdict(list)
+for subdir in ('whisper/dummy.txt', 'ceres/dummy.txt', 'rrd/dummy.txt',
+               'log/dummy.txt', 'log/webapp/dummy.txt'):
+    storage_dirs.append(('storage/%s' % subdir, []))
```
ploxiln: since you're editing this, might as well clean it up a bit:

```python
for subdir in ('whisper', 'ceres', 'rrd', 'log', 'log/webapp'):
    storage_dirs.append(('storage/%s/dummy.txt' % subdir, []))
```

also not sure if bot "log" and "log/webapp" are needed

ploxiln:
I think a rebase went a bit wrong, you ended up with a copy of the commit "fix dashboard graph metric list icon paths with URL_PREFIX" from the master branch
in master: 0a037db4b2d864734e14dd6302bc71194f53e8d3
in this branch: 1ba4da55c08035cccfcdaae2220f8d384dbd1929

setup.py
ploxiln: it also might be nice to use ".keep" instead of "dummy.txt"

webapp/graphite/logger.py
```diff
@@ -28,6 +31,19 @@ def emit(self, record):
...
+# this can't be put in util.py because of circular depency between log.py
+# and util.py
+def _mkdir_p(path):
+    try:
+        os.makedirs(path)
+    except OSError as exc:  # Directory already exists
+        if exc.errno == errno.EEXIST and os.path.isdir(path):
+            pass
+        else:
+            raise
```
ploxiln: could simplify to:

```python
except OSError as exc:
    if not (exc.errno == errno.EEXIST and os.path.isdir(path)):
        raise
```

setup.py
piotr1212: cleaned up. Not sure what happens wit the .keep vs dummy.txt when updating. Will leave it like this for now.

piotr1212:
> I think a rebase went a bit wrong, you ended up with a copy of the commit "fix dashboard graph metric list icon paths with URL_PREFIX" from the master branch
in master: 0a037db
in this branch: 1ba4da5

I think I merged instead of rebased. Anyway, cleaned up now.

docs/releases/1_2_0.rst
```diff
@@ -0,0 +1,132 @@
...
+^^^^^^
+* TODO
+
+Whisper
+^^^^^^^
+* TODO
+
+Carbonate
+^^^^^^^^^
+kA None
```
ploxiln: very minor - this "kA" looks unintentional

ploxiln:
All looks good to me.

I had another thought about storage dirs ... I think the original idea behind using /opt/graphite is those storage and log dirs, which would be awkward in the python site-packages directory. Maybe the thing to do is divorce them from the graphite application root, and default to /opt/graphite/storage and /opt/graphite/log regardless of the install prefix? And not mention them in setup.py at all? I suppose the downside is they would not be created by install. Just an idea - I suppose I always customize these dirs anyway.

piotr1212:
good point. I'll have a look. But busy at the moment with more higher prio stuff.

ploxiln:
This is a big painful change, but I think it's still relatively important. Just sayin' to appease the stale bot :)

adamboutcher:
If this will fix the pip install, can we get it merged?

deniszh:
Unfortunately, it's not that easy. That would break backward compatibility,
and need more changes in carbon. Also, not sure if that fix issue too :/

ср, 19 июл. 2023 г., 18:36 Adam ***@***.***>:

PR #2411, opened 2019-01-15
Fix #1814: check if series is not empty before deleting first element

### LINKED ISSUE #1814: "index out of range" when JSON format consolates series from movingAverage() and friends
Hello.

I see an error with all the `moving*()` functions when getting JSON-format data and specifying `maxDataPoints`.

```
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 149, in get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 147, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/opt/graphite/webapp/graphite/render/views.py", line 154, in renderView
    del series[0]
IndexError: list assignment index out of range
```

The problematic query (generated by Grafana) looks like this:

```
/render/?from=-24hours&target=movingAverage(*.some.metric%2C"10day")&maxDataPoints=462&format=json
```

The problem does not happen all of the time. It occasionally works, but usually doesn't.

It looks like the consolidation is trying to remove more data points from a series than are really there. That leads me to believe that the logic around that is probably flawed.

One of the series that I'm returning is empty and it's sufficient to trigger the error to include that series and one other non-empty one (but not the empty one on its own).

As well as coming and going over time it comes and goes when I tweak the `from`, `maxDataPoints` and `windowSize` parameters in the query. That leads me to believe that there is some edge case in the logic causing the problem (for example where something is an exact multiple of something else).

I can avoid the problem by modifying `renderView` to check that the series is non-empty before doing the `del`, but I'm pretty sure that there is a deeper logic problem than that, so I'm reluctant to submit a PR with that "fix" until someone else has had a go at understanding what's going on. I've spent the day staring at it and I'm out of clues.

AlexZzz:
Hi all.
This is not the best way to fix this bug probably, but it works.

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -214,7 +214,8 @@ def renderViewJson(requestOptions, data):
             series.start = series.start + nudge
             valuesToLose = int(nudge/series.step)
             for r in range(1, valuesToLose):
-              del series[0]
+              if series:
+                del series[0]
             series.consolidate(valuesPerPoint)
 
       datapoints = series.datapoints()
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -204,7 +204,7 @@ def renderViewJson(requestOptions, data):
         if maxDataPoints == 1:
           series.consolidate(len(series))
         else:
-          numberOfDataPoints = timeRange/series.step
+          numberOfDataPoints = len(series)
           if maxDataPoints < numberOfDataPoints:
             valuesPerPoint = math.ceil(float(numberOfDataPoints) / float(maxDataPoints))
             secondsPerPoint = int(valuesPerPoint * series.step)
@@ -214,8 +214,7 @@ def renderViewJson(requestOptions, data):
             series.start = series.start + nudge
             valuesToLose = int(nudge/series.step)
             for r in range(1, valuesToLose):
-              if series:
-                del series[0]
+              del series[0]
             series.consolidate(valuesPerPoint)
 
       datapoints = series.datapoints()


piotr1212:
If it breaks other functions as written here https://github.com/graphite-project/graphite-web/issues/1814#issuecomment-439843600 then I don't think this should be merged.. 

Have you tried replacing https://github.com/graphite-project/graphite-web/pull/2411/files#diff-e87e8c4fda2bf408f4864aad6741df42L207
`numberOfDataPoints = timeRange/series.step` with `numberOfDataPoints = len(series)`. I think I once run into an issue that this was not the same (which it should..).

DanCech:
I believe that the reason `timeRange` is used is so that all series get consolidated to the same interval, rather than series that are only active for part of the interval being returned at a higher resolution.

piotr1212:
> I believe that the reason timeRange is used is so that all series get consolidated to the same interval, rather than series that are only active for part of the query time range being returned at a higher resolution.

How does that work out if `data` has series with different step?

> Anyway, I'm not sure how to fix this in a normal way. We don't use a lot of aggregation functions, so probably we can't test it well

I think the real issue is that sometimes the number of datapoints in a serie does not match the expected number (timerange/step). I've run into this before and couldn't find where it is caused. I guess this happens somewhere deeper in whisper or in the code that merges whisper and carbonlink data. Not sure..

@rickrian does it show only one datapoint for series which would otherwise not work at all or for all series (which did work before the patch)?

DanCech:
> > I believe that the reason timeRange is used is so that all series get consolidated to the same interval, rather than series that are only active for part of the query time range being returned at a higher resolution.
> 
> How does that work out if `data` has series with different step?

The step of the underlying series will just impact how many raw points go into each aggregated interval.

> > Anyway, I'm not sure how to fix this in a normal way. We don't use a lot of aggregation functions, so probably we can't test it well
> 
> I think the real issue is that sometimes the number of datapoints in a serie does not match the expected number (timerange/step). I've run into this before and couldn't find where it is caused. I guess this happens somewhere deeper in whisper or in the code that merges whisper and carbonlink data. Not sure..
> 
> @rickrian does it show only one datapoint for series which would otherwise not work at all or for all series (which did work before the patch)?

I suspect it could happen if you have a series that returned points for only part of the interval.  If you look at lines 197-199 it's trying to figure out the total time range covered by all of the series and setting that as `timeRange`.

AlexZzz:
Does the latest change break something?
If it is not, might these changes be merged?

piotr1212:
ok, for me. @DanCech What do you think?

DanCech:
I don't think this is a good idea as it stands, because of the change it introduces in how the final step is calculated.

PR #2421, opened 2019-02-01
FIX /static/ alias for Apache config

sbancal:
/static/ alias in Apache config doesn't match with other configs (like nginx) ... and doesn't match the files. Fixing it.

### DIFF
--- a/docs/config-webapp.rst
+++ b/docs/config-webapp.rst
@@ -161,9 +161,9 @@ Finally, configure the apache vhost. (You can find example of Graphite vhost con
 
         WSGIScriptAlias / /opt/graphite/conf/graphite.wsgi
 
-        Alias /static/ /opt/graphite/static/
+        Alias /static/ /opt/graphite/webapp/content/
 
-        <Directory /opt/graphite/static/>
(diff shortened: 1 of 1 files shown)


deniszh:
Sorry, @sbancal , but that's wrong. Not sure about which Nginx config do you mean, but `/opt/graphite/static/` is right path, jou just need to run `collectstatic` Django command, as described in documentation - see STATIC_ROOT in https://graphite.readthedocs.io/en/latest/config-local-settings.html

piotr1212:
In my attempt to make installation easier, I'm trying to get rid of the whole collectstatic mess and serve static files with whitenoise from the application as default. Then you also won't need to configure the webserver for serving static files, which makes it even easier.

We'll still have an option to serve static files from a webserver, but using whitnoise as default should be good enough for most installations, especially since I expect most users running grafana by now and not using graphite dashboards.

https://github.com/graphite-project/graphite-web/pull/2409

sbancal:
@deniszh 
You're right, running `collectstatic` fixes that creates that missing folder. I was running all side of the doc ... but didn't see that one ...

I was talking of the lines 79 and 257 in that same file.

PR #2424, opened 2019-02-06
fix dashboard graph metric list icon paths with URL_PREFIX

ploxiln:
I configure graphite-web with `URL_PREFIX = "/graphite"` and these icon urls ended up wrong. This seems to fix it. All other icons in the dashboard and composer seem to be working. But maybe there's an even more correct way, related to the "static" template function? Should it be saved to a variable name like the others here:

https://github.com/graphite-project/graphite-web/blob/1.1.5/webapp/graphite/templates/dashboard.html#L27

EDIT: I went ahead and re-did this the global var way

----
for reference where this is:

before:

screen shot 2019-02-05 at 20 37 27

after:

screen shot 2019-02-05 at 20 39 01

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -4,7 +4,7 @@
 /*global AUTOCOMPLETE_DELAY CALENDAR_ICON CLOCK_ICON CONTEXT_FIELD_WIDTH FINDER_QUERY_DELAY*/
 /*global HELP_ICON NEW_DASHBOARD_REMOVE_GRAPHS REFRESH_ICON REMOVE_ICON RESIZE_ICON*/
 /*global SHARE_ICON UI_CONFIG initialState initialError permissions queryString userName*/
-/*global permissionsUnauthenticated schemes*/
+/*global UP_ICON DOWN_ICON TRASH_ICON permissionsUnauthenticated schemes*/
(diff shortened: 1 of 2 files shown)


ploxiln:
I also found this way of using getting the static root:
https://github.com/graphite-project/graphite-web/blob/1.1.5/webapp/content/js/composer_widgets.js#L154

ploxiln:
one test matrix entry failing: `TOXENV=py27-django111-pyparsing2-mysql`

due to:

```
...
  File ".../site-packages/django/db/backends/base/base.py", line 189, in connect
    self.connection = self.get_new_connection(conn_params)
  File ".../site-packages/django/db/backends/mysql/base.py", line 276, in get_new_connection
    conn.encoders[SafeBytes] = conn.encoders[bytes]
KeyError: <type 'str'>
```

ploxiln:
I think this is pretty good. I'll squash down to one commit if I get the go-ahead :)

piotr1212:
mysqlclient 1.4 changed some internal api's which Django < 2 was depending on, that is why those tests fail. This will be restored in the next mysqlclient release so didn't bother add a constraint for that time.

ploxiln:
Thanks. squashed

deniszh:
Thanks a lot! Merged

PR #2126, opened 2017-11-25
When creating DB do not use --run-sync-db

AlejandroRivera:
With `--run-sync-db`:
```
$ sudo PYTHONPATH=/opt/graphite/webapp django-admin.py migrate --settings=graphite.settings --run-syncdb
Operations to perform:
  Synchronize unmigrated apps: account, render, staticfiles, whitelist, metrics, url_shortener, dashboard, composer, events, browser
  Apply all migrations: admin, contenttypes, tagging, auth, sessions
Synchronizing apps without migrations:
  Creating tables...
    Creating table account_profile
    Creating table account_variable
    Creating table account_view
    Creating table account_window
    Creating table account_mygraph
    Creating table dashboard_dashboard
    Creating table dashboard_template
    Creating table events_event
    Creating table url_shortener_link
    Running deferred SQL...
Traceback (most recent call last):
  File "/usr/local/bin/django-admin.py", line 5, in <module>
    management.execute_from_command_line()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 172, in handle
    self.sync_apps(connection, executor.loader.unmigrated_apps)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 293, in sync_apps
    cursor.execute(statement)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "/usr/local/lib/python2.7/site-packages/django/db/utils.py", line 95, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/usr/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
django.db.utils.ProgrammingError: relation "auth_user" does not exist
```

Without `--run-sync-db`:
```
[...]

PR #2425, opened 2019-02-07
docs: for sql db migration to 1.1 recommend --fake-initial

### LINKED ISSUE #2412: docs: migration from 1.0 to 1.1 may need --fake-initial
I initially created the sqlite database for graphite-web 1.0.2 according to the documentation recommendation:

https://graphite.readthedocs.io/en/latest/config-database-setup.html#webapp-database-setup

```
 PYTHONPATH=... django-admin migrate --settings=graphite.settings --run-syncdb
```

Recently I attempted to upgrade to graphite-web 1.1.5, and followed the documented command in the 1.1.1 release notes:

https://graphite.readthedocs.io/en/latest/releases/1_1_1.html#upgrading

```
PYTHONPATH=... django-admin migrate --noinput --settings=graphite.settings --run-syncdb
```

which resulted in failure:

```
Operations to perform:
  Apply all migrations: account, admin, auth, contenttypes, dashboard, events, sessions, tagging, tags, url_shortener
Running migrations:
  Applying account.0001_initial...Traceback (most recent call last):
  File "/usr/bin/django-admin", line 11, in <module>
    load_entry_point('Django==1.11.18', 'console_scripts', 'django-admin')()
  ...
  File "/usr/lib/python2.7/site-packages/django/core/management/commands/migrate.py", line 204, in handle
    fake_initial=fake_initial,
  File "/usr/lib/python2.7/site-packages/django/db/migrations/executor.py", line 115, in migrate
    state = self._migrate_all_forwards(state, plan, full_plan, fake=fake, fake_initial=fake_initial)
  File "/usr/lib/python2.7/site-packages/django/db/migrations/executor.py", line 145, in _migrate_all_forwards
    state = self.apply_migration(state, migration, fake=fake, fake_initial=fake_initial)
  ...
  File "/usr/lib/python2.7/site-packages/django/db/backends/utils.py", line 62, in execute
    return self.cursor.execute(sql)
  File "/usr/lib/python2.7/site-packages/django/db/backends/sqlite3/base.py", line 326, in execute
    return Database.Cursor.execute(self, query)
django.db.utils.OperationalError: table "account_mygraph" already exists
```

I found another that hit this too, and didn't seem to solve it directly:
https://github.com/graphite-project/docker-graphite-statsd/issues/32

Anyway, I think I know why this happens. This commit was merged just before 1.1.0 - d1912d4bc0f0f878570120173e9d0845a979938f - and it added migrations for existing django app models. Previously, tables were created for these models due to the option `--run-syncdb`, which according to django docs:

> Allows creating tables for apps without migrations. While this isn’t recommended ...

... but now there are un-applied migrations, which conflict with tables already in the database. I think this would happen to anyone who initialized the db before 1.1 and then followed the upgrade instructions in 1.1.1.

Luckily, `django-admin migrate` has another relevant option `--fake-initial`:

> Allows Django to skip an app’s initial migration if all database tables with the names of all models created by all CreateModel operations in that migration already exist. This option is intended for use when first running migrations against a database that preexisted the use of migrations. This option does not, however, check for matching database schema beyond matching table names ...

Running migrate with `--fake-initial` I get:

```
$ ... django-admin migrate --no-input --settings=graphite.settings --fake-initial
Operations to perform:
  Apply all migrations: account, admin, auth, contenttypes, dashboard, events, sessions, tagging, tags, url_shortener
Running migrations:
  Applying account.0001_initial... FAKED
  Applying auth.0008_alter_user_username_max_length... OK
  Applying dashboard.0001_initial... FAKED
  Applying events.0001_initial... FAKED
  Applying tags.0001_initial... OK
  Applying url_shortener.0001_initial... FAKED
```

So it seems to me that the release notes should recommend using `--fake-initial` to migrate from 1.0, and `--run-syncdb` should no longer be used for migrating from 1.0 nor for new databases.

ploxiln:
... and --run-syncdb should no longer be needed.

Initial migrations were created for all models just before 1.1.0
in d1912d4bc0f0. Prior to that, some models did not have any
migrations (and no initial migration), so --run-syncdb was needed
to create tables for them. But now that they have initial migrations,
those migrations run because django knows they have never been
applied, and they fail because the tables already exist. The
--fake-initial option is designed for this situation: it considers
initial migrations already applied if the tables already exist.

fixes #2412 
replaces #2126 

conflicts with #2409 and this is a tiny change so I'd be fine with rebasing this after that lands

### DIFF
--- a/docs/config-database-setup.rst
+++ b/docs/config-database-setup.rst
@@ -20,7 +20,10 @@ To set up a new database and create the initial schema, run:
 
 .. code-block:: none
 
-  PYTHONPATH=$GRAPHITE_ROOT/webapp django-admin.py migrate --settings=graphite.settings --run-syncdb
+  PYTHONPATH=$GRAPHITE_ROOT/webapp django-admin.py migrate --settings=graphite.settings
+
+.. note ::
+  Graphite-Web 1.0 and earlier had some models without migrations, and with Django 1.9 or later, the ``--run-syncdb`` option was needed for migrate to create tables for these models. (Django 1.8 and earlier did not have this option, but always exhibited this behavior.) In Graphite-Web 1.1 and later all models have migrations, so ``--run-syncdb`` is no longer needed. If upgrading a database created by Graphite-Web 1.0 or earlier, you need to use the ``--fake-initial`` option for migrate: it considers an initial migration to already be applied if the tables it creates already exist.
 
 If you are experiencing problems, uncomment the following line in /opt/graphite/webapp/graphite/local_settings.py:
 
--- a/docs/config-local-settings.rst
+++ b/docs/config-local-settings.rst
@@ -373,7 +373,7 @@ The following configures the Django database settings. Graphite uses the databas
 See the `Django documentation <https://docs.djangoproject.com/en/dev/ref/settings/#databases>`_ for full documentation of the DATABASES setting.
 
 .. note ::
-  Remember, setting up a new database requires running ``PYTHONPATH=$GRAPHITE_ROOT/webapp django-admin.py migrate --settings=graphite.settings --run-syncdb`` to create the initial schema.
+  Remember, setting up a new database requires running ``PYTHONPATH=$GRAPHITE_ROOT/webapp django-admin.py migrate --settings=graphite.settings`` to create the initial schema.
 
 .. note ::
   If you are using a custom database backend (other than SQLite) you must first create a $GRAPHITE_ROOT/webapp/graphite/local_settings.py file that overrides the database related settings from settings.py. Use $GRAPHITE_ROOT/webapp/graphite/local_settings.py.example as a template.
--- a/docs/releases/1_1_1.rst
+++ b/docs/releases/1_1_1.rst
@@ -176,7 +176,9 @@ If you're not already running from the *master* branch, Graphite-Web's applicati
   sudo cp /opt/graphite/storage/graphite.db \
         /opt/graphite/storage/graphite.db.backup-`date +%Y%m%d_%H%M%S`
   sudo PYTHONPATH=/opt/graphite/webapp django-admin.py migrate \
-        --noinput --settings=graphite.settings --run-syncdb
+        --noinput --settings=graphite.settings --fake-initial
+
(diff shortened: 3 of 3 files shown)


deniszh:
@ploxiln : will it work for all Django versions? Even if we upgrading from Graphite 0.9.x and/or Django 1.4 ?
Because `--run-syncdb` was added because normal `migrate` command did nothing before...

ploxiln:
`--run-syncdb` was needed before 1.1, but with 1.1+ it should not be needed, even for databases originally created with graphite 0.9.x and older django. While figuring out #2412 also I tested making a new database from nothing, without `--run-syncdb`, and all tables were created (by initial migrations which are now present).

(Now `--fake-initial` may be needed only because `--run-syncdb` was used before, when it was needed.)

deniszh:
@ploxiln : maybe that should be mentioned in docs too? Could you please amend it? Thanks!

ploxiln:
possibly worth mentioning: `--fake-initial` was added in django 1.8 and `--run-syncdb` was added in django 1.9
(according to https://docs.djangoproject.com/en/1.9/ref/django-admin/#migrate)

ploxiln:
looks like, in django 1.8 and earlier, the behavior of `--run-syncdb` was unconditionally provided:

> No arguments: All migrated apps have all of their migrations run, and all unmigrated apps are synchronized with the database

ploxiln:
OK, I'll try adding a bit more to the docs about this

piotr1212:
I'm wondering all the time if we could automate this stuff on app startup? Just like Grafana does.

ploxiln:
Let me know how that looks/sounds.

I'll squash commits before merge.

deniszh:
For me, it looks great. Maybe I could ask @gwaldo to check?

ploxiln:
Good idea - I think I figured this out, but I'm no django expert :)

DanCech:
Is there any harm in always recommending users specify `--fake-initial`?

ploxiln:
I don't think there would be any harm in always using `--fake-initial`.

It should be needed at most once, when migrating from < 1.1 to >= 1.1, and the two places that that I did not include `--fake-initial` in the example were about "setting up a new database" - so I guess I did not want to encourage "just type this even though it does not make sense".

deniszh:
OK, I'm merging this for now.

PR #797, opened 2014-07-29
Ability to serve graphite-web under URL prefix

brutasse:
Fixes #328 #327 #628 #223 #524 #231

This patch introduces a new setting, `URL_PREFIX`, that allows to configure the webapp for serving under a non-root URL.

Docs for apache/nginx aren't there yet so it's not ready for a merge. I'd like people who need that feature to try it and provide feedback.

PR #2433, opened 2019-03-28
Dashboard render urls missing document.body.dataset.baseUrl

cbowman0:
This makes these url settings match the others in dashboard.js like in PR #797

I don't run on a different baseUrl, but I believe these should match the others.


deniszh:
Agreed

Issue #2449, opened 2019-05-24
Calling evaluateTarget in a function loses prefetched data from other targets


cbowman0:
In a request that has multiple targets (or multiple series in the params), that invokes a function that calls evaluateTarget directly (like useSeriesAbove) we lose the prefetched data.

Example:
target=useSeriesAbove(a.b.c.d,0,"d","e")&target=w.x.y.z would do:
1. prefetchData for a.b.c.d and w.x.y.z
2. Invoke useSeriesAbove for a.b.c.d and invoke evaluateTarget on a.b.c.e
3. prefetchData for a.b.c.e
4. Overwrite all prefetched data with the data from a.b.c.e in requestContext
5. When the loop in evaluateTarget gets to w.x.y.z and calls evaluateToken, the prefetched data for it no longer exists and it determines there's no data for that series and doesn't return anything for it.

I attempted to use a copy of requestContext in useSeriesAbove, but that is a shallow copy of the dict, so no change in behavior.  Attempting to use copy.deepcopy() resulted in an error about a str not being callable.  So that didn't work.  If we do:

```
-  newSeries = evaluateTarget(requestContext, 'group(%s)' % ','.join(newNames))
+  newContext = requestContext.copy()
+  newContext['prefetched'] = {}
+
+  newSeries = evaluateTarget(newContext, 'group(%s)' % ','.join(newNames))
```

An alternative solution is to merge the data in prefetch().  It's more generic (and would prevent issues on other uses like this) but could have side effects:
```
-  requestContext['prefetched'][(startTime, endTime, now)] = prefetched
+  if not requestContext['prefetched'].get((startTime, endTime, now)):
+    requestContext['prefetched'][(startTime, endTime, now)] = prefetched
+  else:
+    requestContext['prefetched'][(startTime, endTime, now)].update(prefetched)
```

DanCech:
I suspect that `newContext['prefetched'] = {}` is going to be the simplest solution.

cbowman0:
I thought so, too.  I'll prepare a PR for that.

PR #2450, opened 2019-05-24
Copy requestContext() and empty prefetch

cbowman0:
Copy requestContext() and empty prefetch, so we don't override prior fetched data in functions that invoke evaluateToken directly.

Deals with the issues found in #2449


cbowman0:
I have had this deployed for a few days with no complaints.

PR #2452, opened 2019-06-04
Add Statusengine to list of integrations (Forwarding)

nook24:
More information about Statusengine and Tutorial about how to setup: https://statusengine.org/tutorials/Graphite-Grafana/

### DIFF
--- a/docs/tools.rst
+++ b/docs/tools.rst
@@ -125,7 +125,7 @@ Forwarding
 `statsd`_
   A simple daemon for easy stats aggregation, developed by the folks at Etsy. A list of forks and alternative implementations can be found at <http://joemiller.me/2011/09/21/list-of-statsd-server-implementations/>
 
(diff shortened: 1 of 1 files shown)


docs/tools.rst
```diff
@@ -125,6 +125,9 @@ Forwarding
 `statsd`_
   A simple daemon for easy stats aggregation, developed by the folks at Etsy. A list of forks and alternative implementations can be found at <http://joemiller.me/2011/09/21/list-of-statsd-server-implementations/>
 
+`_Statusengine`_
```
piotr1212: Build fails:

> /home/travis/build/graphite-project/graphite-web/docs/tools.rst:129: ERROR: Unknown target name: "_statusengine".

I think you need to get rid of the prefix underscore

nook24: Thanks for the hint looks better now :)

PR #2462, opened 2019-07-04
Django22 compatibility

piotr1212:
Adds "django.contrib.messages" to INSTALLED_APPS in local_settings.py for Django22 compatibility.

Adds Django 2.2 to testing, removes Django 1.8, 2.0 and Python 3.4 from testing (EOL).


deniszh:
Thanks, @piotr1212 !

PR #2490, opened 2019-09-26
docs: add netdata to 'tools that work with graphite'

sbasgall:
netdata is a project missing from this list. I hope people can benefit from having it linked here.


deniszh:
Sure. Thanks!

Issue #2355, opened 2018-09-12
minimumBelow should use < and not =<


erez-rabih:
I think that `minimumBelow` should be the exact opposite of `maximumAbove` which uses the `>` operator  and not the `>=` 
minimumBelow: https://github.com/graphite-project/graphite-web/blob/c96420c7f80e8400db5364145bdaa3c4a6c40c9e/webapp/graphite/render/functions.py#L2835

maximumAbove:
https://github.com/graphite-project/graphite-web/blob/c96420c7f80e8400db5364145bdaa3c4a6c40c9e/webapp/graphite/render/functions.py#L2766

Currently this is very confusing since one uses > and the other =<

DanCech:
When I implemented `filterSeries` to replace the old implementations I set it up that way to duplicate the previous behavior.  If we want to change the semantics of `minimumBelow` that would be a breaking change to the API, and we'd need to carefully consider the impact it might have on existing users of the function.

As you're no doubt aware, you can use `filterSeries(<seriesList>, 'min', '<', <n>)` to get the result you're looking for.

erez-rabih:
I agree there is a workaround to get the behaviour I am looking for it just seems very inconsistent that maximumAbove uses > and minimumBelow uses <=
Of course it might impact existing queries and there are other considerations to take into account I just wanted to raise the flag since it took me two hours to understand why my query returns unexpected series
I would at least add a comment on the documentation to make users know that although it says `below` it is implemented as `below or equal to`

PR #2493, opened 2019-10-05
Updated minimumBelow() docstring #2355

bigpythonimish:
Issue #2355 suggests that minimumBelow() works on <= rather than <, to avoid confusion I have updated docstring


deniszh:
Thanks!

PR #2499, opened 2019-10-11
Removing  function 'send_email'

deniszh:
It's insecure, not used in the code and undocumented.

Fixing #2008 

Please also note that sending email from Dashboard is implemented in completely different commit 'e2a70d8' and not use 'send_email' function at all.


JLLeitschuh:
You also need to remove the call to this endpoint from inside of the UI.

deniszh:
> You also need to remove the call to this endpoint from inside of the UI, unless that UI component uses some newer logic.

That's a point - there's no call from UI exist. Or any other call. It looks like some abandoned piece of code, which was never used.

cbowman0:
Which UI component?  I find no callers of send_email.

DanCech:
It looks like we can also get rid of the SMTP_SERVER config setting, which isn't used anywhere else.

deniszh:
@DanCech : well... Dashboard have ability to send email, see https://github.com/graphite-project/graphite-web/commit/e2a70d8fe5a37ed52cd8166233607bcc50c15b64

deniszh:
It's probably also worth to backport it to 0.9.x and 1.0.x
Not sure should we make new releases for this branches... Maybe fixed branch should be enough.

DanCech:
Yes but that doesn't use that config setting, it uses the django `EmailMessage` class which is configured via the django `EMAIL_*` settings.

deniszh:
> Yes but that doesn't use that config setting, it uses the django `EmailMessage` class which is configured via the django `EMAIL_*` settings.

Ah, nice find then, @DanCech
Removed together with unused imports.

PR #2502, opened 2019-10-23
Upgrading minimal Django version

deniszh:
According to https://www.djangoproject.com/download/ 1.11 is single 1.x supported version now.
According to https://github.com/graphite-project/graphite-web/network/alert/docs/requirements.txt/django/open versions <1.11.19 vulnerable to CVE-2019-6975 and CVE-2019-3498


deniszh:
I think they're not in use because not in `envlist`, but that's good thing anyway

PR #2507, opened 2019-10-24
Merge prefetched data

liyichao:
prefetchData may be called multiple times in one request, so prefetched data must be merged.

### DIFF
--- a/webapp/graphite/render/datalib.py
+++ b/webapp/graphite/render/datalib.py
@@ -281,7 +281,11 @@ def prefetchData(requestContext, pathExpressions):
 
   if not requestContext.get('prefetched'):
     requestContext['prefetched'] = {}
-
-  requestContext['prefetched'][(startTime, endTime, now)] = prefetched
+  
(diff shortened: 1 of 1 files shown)


webapp/graphite/render/datalib.py
```diff
@@ -281,7 +281,11 @@ def prefetchData(requestContext, pathExpressions):
 
   if not requestContext.get('prefetched'):
     requestContext['prefetched'] = {}
-
-  requestContext['prefetched'][(startTime, endTime, now)] = prefetched
+  
+  if (startTime, endTime, now) in requestContext['prefetched']:
+    for k in prefetched:
+      requestContext['prefetched'][(startTime, endTime, now)][k] = prefetched[k]
```
DanCech: Rather than looping can we just use `update()`?
```suggestion
    requestContext['prefetched'][(startTime, endTime, now)].update(prefetched)
```

liyichao: Done.

Issue #2515, opened 2019-11-14
Migrate out of ISSUE_TEMPLATE.md to new template system


deniszh:
Github doc: https://help.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository#creating-issue-templates

Also, we need to modify template, because we're also support generic support requests in github issues (or do we?)

PR #2517, opened 2019-11-19
Update issue templates

deniszh:
Fixing #2515


deniszh:
Please review @DanCech @piotr1212 @cbowman0 
If looks OK - will copy to whisper, carbon and carbonate.

PR #2521, opened 2019-12-07
add graphite-dl4j and carbon-proxy

jdbranham:
Better late than never? =]


deniszh:
Hi @jbdranham,
Thanks a lot for PR!
Could you please move carbon-proxy to Forwading section and please also note that tools in every section sorted alphabetically.
Thanks!

deniszh:
Looks good (except known things), merging

PR #2526, opened 2020-01-14
Skipping pyhash env when testing python 2.7

deniszh:


deniszh:
I think more and more modules will give up on 2.7 support. 
We should or pin them or drop 2.7 too. Djanjo 1.11 LTS ends support in April 2020 too, later versions are 3.x only.
Thoughts? @piotr1212 @DanCech

deniszh:
Our tests are extremely flaky too :(

deniszh:
Nah, looks like still broken

ploxiln:
the latest pyhash is still pyhash-0.9.3 which was released in march last year ... I don't think its python2 compatibility has changed

(and when it does, if it changes the metadata in setup.py correctly, `pip2 install pyhash` should continue to install the last python2 compatible version)

ploxiln:
... all those python3 syntax errors on python2, are actually due to the version of **pytest** which is used while installing pyhash, for some reason:

```
Collecting pyhash

  Downloading https://files.pythonhosted.org/packages/f0/bf/4db9bed05d10824a17697f65063de19892ca2171a31a9c6854f9bbf55c02/pyhash-0.9.3.tar.gz (602kB)

    Complete output from command python setup.py egg_info:

      File "build/bdist.linux-x86_64/egg/_pytest/doctest.py", line 135
        self, reprlocation_lines: Sequence[Tuple[ReprFileLocation, Sequence[str]]]
                                ^
    SyntaxError: invalid syntax

      File "build/bdist.linux-x86_64/egg/_pytest/recwarn.py", line 59
        expected_warning: Union["Type[Warning]", Tuple["Type[Warning]", ...]],
                        ^
    SyntaxError: invalid syntax

      File "build/bdist.linux-x86_64/egg/_pytest/pytester.py", line 158
        def _pytest(request: FixtureRequest) -> "PytestArg":
                           ^
```

deniszh:
Hi @ploxiln ,

Thanks, I overlooked that! Not sure why this happening, according to [docs](https://docs.pytest.org/en/latest/py27-py34-deprecation.html) it should pick proper version automatically.

ploxiln:
This is probably due to `setup_requires` and pip not being smart enough about recursive setup-only dependencies somehow ... maybe if you explicitly install pytest-runner and pytest-benchmark first, it will have appropriate versions. Now that I look for it, pytest-runner documents some odd warnings about `setup_requires` - https://pypi.org/project/pytest-runner/

ploxiln:
(it looks like `setup_requires` does installs without using pip ... that makes sense now ... installing the pytest things explicitly with pip first is very likely the fix)

ploxiln:
forgot to link to https://github.com/flier/pyfasthash/blob/0.9.3/setup.py#L162

ploxiln:
... guess I was wrong, explicitly installing pytest stuff first with pip didn't help setup.py `setup_requires` ... but I have trouble reproducing on ubuntu-16.04 so I wonder if updating setuptools would help, but I'm not hopeful ...

ploxiln:
... maybe it's because tox creates pristine environments, so the pytest stuff needs to be at the top of the tox deps list ...

deniszh:
@ploxiln : yep, I already tested that. 
I see that proper pytest and more-itertools already installed
```Requirement already satisfied: more-itertools<=5.0.0 in /home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages (4.3.0)

Requirement already satisfied: pytest<=5.0.0 in /home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages (4.3.1)```
but pyhash trying to install them again, for some reason...

deniszh:
OK, will try to fix same issue in carbon, to isolate pyhash issue, then I'll return here.

Issue #2533, opened 2020-01-22
[Q] Graphite events broken in Django 3 because of tagging which is eol


piotr1212:
Tests fail on Django3 for the Graphite-events component:

```
  File "/home/piotr/git/graphite-web/webapp/graphite/events/models.py", line 4, in <module>
    from tagging.models import Tag
  File "/home/piotr/git/graphite-web/.tox/py36-django30-pyparsing2/lib/python3.6/site-packages/tagging/models.py", line 7, in <module>
    from django.utils.encoding import python_2_unicode_compatible
ImportError: cannot import name 'python_2_unicode_compatible
```

`python_2_unicode_compatible` is now only available through six. Problem here is that we do not import that method directly but through `django-tagging`. `django-tagging` is quite dead, the replacement seems to be `django-taggit`. Taggit uses different schemas which makes migrating for users a bit of a mess, see https://blog.birdhouse.org/2011/04/17/migrate-django-tagging-taggit/

I think there are several options to go from here:

1. Bundle `django-tagging` and patch it.
2. Migrate to `django-taggit`, write migration instructions.
3. Drop graphite-events, Grafana now has a built in annotations store.

I don't plan on implementing 2. What do you think?

deniszh:
If it's only for events, I would say 3.

piotr1212:
Yes, it is just for events. I have no clue how many people are actually using it.

Grafana does support the [Graphite events format](https://grafana.com/docs/grafana/latest/http_api/annotations/#create-annotation-in-graphite-format) so if I understand this correctly users will just have to change the endpoint.

deniszh:
Well, data migration is also needed. Maybe we can fix it differently then...

deniszh:
But we need to got to Django 3 anyway, so, in worst case I would say kill events.

piotr1212:
I managed to monkeypatch the imports. We can use this for now and announce that events will be deprecated in the next release, we'll see how much backlash we get and we can eventually create a data migration script.

PR #2534, opened 2020-01-23
Django 3.0 compatibility

piotr1212:
Changes render_to_response() to response().

Changes load staticfiles to load static in templates.

Django 3 sets the private header when setting no-cache headers. Adjusted tests to
reflect this.

Monkeypatched django.utils.enocoding.python2_unicode_compatible so that django-tagging
can continue to function. Django-tagging is dead, it should be replaced with django-taggit
or the graphite events functionality should be removed in favor of Graphana annotations.

Updates travis and tox to test currently maintained Django versions.

### DIFF
--- a/webapp/graphite/events/models.py
+++ b/webapp/graphite/events/models.py
@@ -1,9 +1,15 @@
 import os
 
 from django.db import models
-from tagging.models import Tag
 
-from graphite.events.compat import ModelTaggedItemManager
+# Monkeypatching so that django-tagging can import python_2_unicode_compatible
+import django.utils.encoding
+import six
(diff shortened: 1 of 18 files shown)


piotr1212:
For Django-tagging see: https://github.com/graphite-project/graphite-web/issues/2533

webapp/graphite/events/models.py
```diff
@@ -1,9 +1,15 @@
 import os
 
 from django.db import models
-from tagging.models import Tag
 
-from graphite.events.compat import ModelTaggedItemManager
+# Monkeypatching so that django-tagging can import python_2_unicode_compatible
```
deniszh: 👍

PR #2537, opened 2020-02-02
Adding Pipfile

deniszh:
And excluding Pipfile.lock from repo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+Pipfile.lock
 MANIFEST
 .vagrant/
 .venv/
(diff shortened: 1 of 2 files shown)


piotr1212:
We now have requirements specified in setup.py, setup.cfg, requirements.txt and now we are adding another file? Can pipfile replace the others and could we clean them up?

deniszh:
Well, they are not interchangeable.  So, I'm afraid, no. But Pipfile is optional, `pipenv` can generate it from `requirements.txt`, so, we can skip that.

PR #2548, opened 2020-02-22
Optional replacement value for replaceAboveValue

tmm1:
I'm converting some raw sensor data into a percentage, and want to truncate values above 100 to 100%. This PR lets me do this using a new optional argument to `replaceAboveValue`:

>sensor.soil_capacitance | offset(-244) | scale(0.3813) | removeAboveValue(100, 100)

If this looks reasonable, I could be convinced to add it to the other `replace{Below,Above}{Value,Percentile}` functions too.

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -3271,17 +3271,17 @@ removeAbovePercentile.params = [
 ]
 
 
-def removeAboveValue(requestContext, seriesList, n):
+def removeAboveValue(requestContext, seriesList, n, replacement=None):
   """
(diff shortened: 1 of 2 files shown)


tmm1:
> add it to the other `replace{Below,Above}{Value,Percentile}` functions too

Maybe instead it makes more sense to add two new functions? PromQL has `clamp_max` and `clamp_min`, so maybe we can borrow that terminology.

deniszh:
I'm fine with both variants. Maybe separate functions is better, although modifying `replace` functions also make sense.

PR #2552, opened 2020-02-29
prune flake8 ignore list

ploxiln:
re-enable some flake8 rules which cause the least code churn

### DIFF
--- a/check-dependencies.py
+++ b/check-dependencies.py
@@ -156,3 +156,21 @@ if required:
   sys.exit(1)
 else:
   print("All necessary dependencies are met.")
+
+
+# suppress unused-import warnings
+__all__ = [
+    'whisper',
+    'ceres',
+    'cairo',
+    'django',
+    'pytz',
+    'pyparsing',
+    'tagging',
+    'memcache',
+    'ldap',
+    'txamqp',
+    'rrdtool',
+    'whitenoise',
+    'pyhash',
+]
--- a/webapp/graphite/account/ldapBackend.py
+++ b/webapp/graphite/account/ldapBackend.py
@@ -12,7 +12,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License."""
 
-import ldap, traceback
+import traceback
+
+import ldap
 from django.conf import settings
 from django.contrib.auth.models import User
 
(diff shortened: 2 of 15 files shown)


ploxiln:
DeepSource is finding stuff like improper indent, on the lines I'm fixing other style issues on. It's better to not re-indent a significant part of the file at the same time though.

ploxiln:
The other thing DeepSource is flagging is "redefining name from outer scope" for a bunch of render function kwargs in `webapp/graphite/render/functions.py`, but those will never be fixed because it's the (public api) render function "color()" or "alpha()" that is shadowed by the argument "color" or "alpha" for another (public api) render function.

check-dependencies.py
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# flake8: noqa
```
piotr1212: Could you narrow this one down with the specific rule to ignore? You can do with `# noqa: E123

webapp/graphite/account/ldapBackend.py
```diff
@@ -12,7 +12,9 @@
 See the License for the specific language governing permissions and
 limitations under the License."""
 
-import ldap, traceback
+import traceback
+
```
cbowman0: Just curious why the blank line here?

DanCech:
Thanks for tackling this!

check-dependencies.py
ploxiln: In this file it's just a bunch of "F401 imported but unused" (which is the intention of this script). You can't ignore a specific rule for an entire file, so I'll have to add  `# noqa: F401` to each line. But I just remembered there's another trick using `__all__` which may or may not be cleaner :)

webapp/graphite/account/ldapBackend.py
ploxiln: It is a somewhat less consistent convention, to put standard library imports first, then a blank line, then third-party imports, then a blank line, then project-internal imports.

ploxiln:
I can easily squash down these commits if desired, just let me know.

webapp/graphite/account/ldapBackend.py
cbowman0: Ok.  Thanks for the explanation.

ploxiln:
(I squashed down to 3 commits, which seemed appropriate.)

deniszh:
Thanks a lot again, @ploxiln !
I will merge it today, I want to test backport bot on it.

ploxiln:
Thank you for maintaining graphite :)

PR #2562, opened 2020-03-08
Adding release 1.1.7

deniszh:


deniszh:
I'll move release for later, will do couple of more backports too.

PR #2576, opened 2020-03-31
Suggestion: Use 24-hours time format

Low-power:
I suggest use the 24-hours time format for UI, or make it configurable.

Because I personally has a poor performance on reading 12-hours formatted time string...

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -737,7 +737,7 @@ function initDashboard () {
   var lastRefreshedText = {
     id: 'last-refreshed-text',
     xtype: 'tbtext',
-    text: ( new Date() ).format('g:i:s A')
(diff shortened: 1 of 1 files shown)


deniszh:
Hi @Low-power,
I will gladly accept PR to make this configurable, but would not just blindly replace one hard coded thing with another, sorry.

Low-power:
deniszh, I understand, replacing some hardcoded values to other hardcoded values are probably not a good solution.

piotr1212:
`new Date().toLocaleTimeString()`

Should print it in the locale configured on your system. 

Unfortunately this doesn't understand more complicated set-ups, for example I have LANG=en_US.UTF-8 and LC_TIME=nl_NL.UTF-8 which means I have the EN-us locale except for time formatting which should use the Dutch format. toLocaleTimeString prints it in EN-us format.

Issue #2518, opened 2019-12-01
[BUG] SVG rendering returns bytes object, leading to XML Parsing Error


dueringa:
**Describe the bug**
When attempting to render an SVG file, I get an "XML Parsing Error: not well-formed [...] Line Number 1, Column 2:" in Firefox.

```
b'<?xml version="1.0" encoding="UTF-8"?>\n ...
```
Note the b' in the beginning, it looks like a string representation of a bytes object is returned

**To Reproduce**
Enter the rendering URL https://example.com/render?&target=test.count&format=svg

**Expected behavior**
The requests succeeds with a valid SVG file.

**Screenshots**

**Environment (please complete the following information):**
 - Debian Buster
 - Graphite-web version 1.1.6
 - Python 3.7.3
 - Django version 2.2.7
 - graphite-web and Django installed via pip from pypi (pip install graphite-web) in a venv
 - run/started via uwsgi (also installed in venv), served by nginx

**Additional context**

uwsgi ini file:
```
[uwsgi]
processes = 1
socket = localhost:12222
virtualenv = /opt/graphite
chdir = /opt/graphite/conf
module = wsgi:application
```

```
# conf/wsgi.py 
import sys
import os

sys.path.append('/opt/graphite/webapp')

from graphite.wsgi import application
```

earthgecko:
Experiencing the same thing.

deniszh:
Yes, I reproduced it on latest release in docker. Not sure how to fix it though, looks like it's somewhere in Django. Maybe @piotr1212 knows?

piotr1212:
Yep, already found the issue. Just figuring out what the best way is to fix this.

earthgecko:
@piotr1212 thanks very much, that fix works :+1:

PR #2578, opened 2020-04-03
Fix SVG output for Python3

piotr1212:
fixes graphite-project/graphite-web#2518

### DIFF
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -885,12 +885,12 @@ class Graph:
         metaData = { }
 
       self.surface.finish()
-      svgData = str(self.surfaceData.getvalue())
+      svgData = self.surfaceData.getvalue()
       self.surfaceData.close()
 
-      svgData = svgData.replace('pt"', 'px"', 2) # we expect height/width in pixels, not points
-      svgData = svgData.replace('</svg>\n', '', 1)
-      svgData = svgData.replace('</defs>\n<g', '</defs>\n<g class="graphite"', 1)
+      svgData = svgData.replace(b'pt"', b'px"', 2)  # we expect height/width in pixels, not points
+      svgData = svgData.replace(b'</svg>\n', b'', 1)
+      svgData = svgData.replace(b'</defs>\n<g', b'</defs>\n<g class="graphite"', 1)
 
       if hasData:
         # We encode headers using special paths with d^="M -88 -88"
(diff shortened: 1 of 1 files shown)


piotr1212:
There are some code paths not covered by the test cases. I'll try to create some tests this weekend.

deniszh:
Cool, thanks, Piotr!
If I understand correctly, it will break SVG on Python 2 (but just for my understanding, we're don't care much abouth Python 2 anymore).

piotr1212:
I think it will still work on Python 2. I'm not 100% sure if all those regexes work at all now. Please don't merge yet, I'll try to create some tests.

webapp/graphite/render/glyph.py
```diff
@@ -885,12 +885,12 @@ def output(self, fileObj):
         metaData = { }
 
       self.surface.finish()
-      svgData = str(self.surfaceData.getvalue())
+      svgData = self.surfaceData.getvalue()
```
ploxiln, line 888: instead of adding all the `b` below, you could change this to:

```python
svgData = self.surfaceData.getvalue().decode()
```

That should work on both python2 and python3.

(also in this case, keep the original `.encode('utf-8')` at the end)

ploxiln: to demystify a bit, here's a demo of python2 and python3 behavior:

```
>>> import sys
>>> sys.version
'2.7.17 (default, Jan 14 2020, 03:01:26) \n[GCC 4.2.1 Compatible Apple LLVM 11.0.0 (clang-1100.0.33.17)]'
>>> val = b"<svg>foobar</svg>"
>>> str(val)
'<svg>foobar</svg>'
>>> val.decode()
u'<svg>foobar</svg>'
>>> val.decode().replace("foobar", "baz").encode('utf-8')
'<svg>baz</svg>'
>>> '<svg>baz</svg>' == b"<svg>baz</svg>"
True
```

```
>>> import sys
>>> sys.version
'3.7.7 (default, Mar 12 2020, 20:55:24) \n[Clang 11.0.0 (clang-1100.0.33.17)]'
>>> val = b"<svg>foobar</svg>"
>>> str(val)
"b'<svg>foobar</svg>'"
>>> val.decode()
'<svg>foobar</svg>'
>>> val.decode().replace("foobar", "baz").encode('utf-8')
b'<svg>baz</svg>'
```

piotr1212: If the input is a byte array and the output has to be a byte array. Then what would be the point of transcoding it to a string at the beginning and then transcoding it back to a byte array at the end?

ploxiln: just to avoid adding `b` to all the lines of code below

webapp/graphite/render/glyph.py
```diff
@@ -899,16 +899,16 @@ def onHeaderPath(match):
           name = ''
           for char in re.findall(r'L -(\d+) -\d+', match.group(1)):
```
ploxiln, line 900: might this also need to be a binary string, because match groups are bytes now?

ploxiln: ... also, `name` is a string, two lines above:

```python
          name = ''
```

 which you can't substitute into a bytes below:

```python
          return b'</g><g data-header="true" class="%s">' % name
```

example:

```
>>> name = ''
>>> b"foobar %s" % name
  File "<stdin>", line 1, in <module>
TypeError: %b requires a bytes-like object, or an object that implements __bytes__, not 'str'
```

ploxiln: (so you see why keeping all this string-composing logic in terms of strings rather than bytes might be easier to keep working and readable)

piotr1212:
I tried to do some refactoring so it could be unittested but ended up in a rabbit hole. I think it would be better to switch to matplotlib eventually. I played a bit with it a year ago and it looks promising, thing was that matplotlib is slow, especially the simpler graphs (with just one series) were rendered twice as fast by graphite's glyph.py.

For now just @ploxiln 's suggestion.

PR #2584, opened 2020-04-14
render: only encodeHeader() for svg output

### LINKED ISSUE #1309: There is no need to call encodeHeaders for png images
It looks like there is no need in calling "encodeHeader" for png images.

It was added by https://github.com/graphite-project/graphite-web/commit/08be1ba8a465554443af474a924dd8f668abdf72 this commit.

Function itself: https://github.com/graphite-project/graphite-web/blob/0.9.x/webapp/graphite/render/glyph.py#L407

From my point of view it draws this text outside of the canvas and it's useless for PNG rendering. I also wonder maybe there is a better way to do whatever it does for PNG images? Maybe it was a limitation of old cairo from 2011?

ploxiln:
fixes #1309
which points out that 08be1ba8a465 added this
odd out-of-bounds "header" just for svg metadata

### DIFF
--- a/webapp/graphite/render/glyph.py
+++ b/webapp/graphite/render/glyph.py
@@ -796,6 +796,12 @@ class Graph:
           y += lineHeight
 
   def encodeHeader(self,text):
+    """
+    Puts some metadata in the generated svg xml that is not inside the frame.
+    This can be used to manipulate the svg later on with a framework like d3.
+    """
+    if self.outputFormat != 'svg':
+      return
     self.ctx.save()
     self.setColor( self.backgroundColor )
     self.ctx.move_to(-88,-88) # identifier


ploxiln:
I finally got around to testing with a real setup, seems to work. I also prepared a super-trivial example of png and svg with this change:

https://www.ploxiln.net/adhoc/render.png
https://www.ploxiln.net/adhoc/render.svg

pdf output is a bit garbled for me, but seems to be garbled the same with graphite-web-1.1.5 which I still have running somewhere. seems the main problem may be the overall width/height? anyway I'm pretty sure it is unrelated (and I don't know the contexts in which pdf output is used)
https://www.ploxiln.net/adhoc/render.pdf

Issue #2334, opened 2018-08-22
Graph Target incorrectly encoded with "&quot;" when saving to "My Graphs", renders graph unusable/broken


josiahbryan:
**What I attempted:**
I created my first graph, putting together a couple stats I'd dumped in, and saved it. The Graph Target items looked like this when I entered them:
`alias(summarize(scale(stats.ncs.dispatch.office.count,10),"1min"),"/office/*")`

I then saved it to My Graphs and closed the composer window, then double-clicked the graph from the tree view to view the saved graph.

**Expected Results**
I expected that when I clicked the graph in the tree view from the "My Graphs" tree that it would open a composer window again with a graph similar to what I had created (black background, some lines, etc)

**Witnessed Result**
Screenshot below:
![image](https://user-images.githubusercontent.com/4821548/44442407-03afbf80-a5a0-11e8-9cba-10e283c4df9f.png)

I boiled it down to the fact that the data ACTUALLY stored in the Graph Target section was this:
`alias(summarize(scale(stats.ncs.dispatch.office.count,10),&quot;1min&quot;),&quot;/office/*&quot;)`

Notice the `&quot` encodings? When I edited that string manually and put them back to regular quotes (e.g. `"`), the graph immediately rendered again. (Notice the "broken image" icon in the screenshot above where the graph should be.)

However, when I opened the graph again from the tree view, still the same result 

This is a show-stopper problem for our organization - if we can't save the graphs, it is a waste of time for us to spend time creating the Graph Targets to view our data over and over again.

**System Details**
This is just the basic docker image, NO customizations or changes done to it - at all. Installed exactly as shown on the github page for the docker image: [https://github.com/hopsoft/docker-graphite-statsd](https://github.com/hopsoft/docker-graphite-statsd)

The `/version` endpoint tells me it's running version `1.1.3`. No sure what the OS/Python/Django/pip versions that the docker image is running - if you need those, let me know, and we'll find a way to get them, I'm sure. 

Otherwise, tips on what I'm doing wrong if it's user error are greatly appreciated! Thanks!

deniszh:
Yes, we have the similar issue - https://github.com/graphite-project/graphite-web/issues/1801
I was thinking that breaking only existing graphs, not the new...
Looks like this issue was introduced after https://github.com/graphite-project/graphite-web/pull/1662 and looks like #1662 not fixing all XSS issues anyway.
Currently, only one workaround exist - use some 3rd party dashboard, like Grafana, for saving graphs.

Issue #1801, opened 2017-01-26
Quotes in saved graphs are expanded to HTML entities, which results in invalid query syntax for Graphite


scianos:
We are currently following the development branch of Graphite-Web. After upgrading last week, we noticed that saved graphs with quotes in the entries result in the quotes being expanded to HTML entities. Example:
`alias(scale(summarize(collectd.someserver.someservice.somemetric.time-to-prod,&quot;1d&quot;,&quot;max&quot;),0.0166667),&quot;Daily+Maximum&quot;)`

Should be:
`alias(scale(summarize(collectd.someserver.someservice.somemetric.time-to-prod,"1d","max"),0.0166667),"Daily+Maximum")`

This results in many saved graphs being broken. Unfortunately, I haven't been able to track back which package or change caused the difference in behavior.

obfuscurity:
Perhaps https://github.com/graphite-project/graphite-web/pull/1662?

dsongc:
We, scianos and I, made it working after removing escape in the code.
`graphUrlParams[param[0]].append(escape(param[1]))`
I think the issue is related to #1662

scianos:
It definitely looks like the escaping of the quote to an HTML entity is related to the issue at hand. The question is how we implement a fix which both solves the XSS and allows existing graph URLs containing quotes to work (the internal patch we're using will re-open the XSS problem, so this isn't a solution we can/would want to commit back yet without further work).

ybizeul:
I'm having lots of random issues recently with that kind of errors in the logs, could that be related?

I think it started happening with a recent Graphite update. I don't see any quote in the query at least in Grafana

```
KeyError: ({Suppress:("\") W:((){}...)}, u'netapp.perf.VYN.S18SAN.node.S18SAN-02.vol_summary.total_ops', 6, True, True)
KeyError: ("}", u'aliasByNode(highestAverage(netapp.perf.{CEL,HAN,MRI,NOR,ODH,PRI,SHA,TOK,VYN}.{S10SAN,S11SAN,S13SAN,S16SAN,S18SAN,S25SAN,S32SAN,S34SAN,S37SAN}.node.*.vol_summary.total_ops, 4), 5)', 75, True, True)
KeyError: ({{Suppress:("\") W:((){}...)} | W:(!#%$...)}, u'netapp.perf.VYN.S18SAN.node.S18SAN-02.vol_summary.total_ops', 12, False, True)
[...]

PR #2587, opened 2020-04-18
fix composer saved graphs target escaping

ploxiln:
saved graphs targets were html-escaped in the json response
to fix an XSS vulnerability in graphite-project/graphite-web#1662

... but that was not really the right place to escape the graph targets,
it broke targets using quotes: #1801 #2334

so effectively revert the original fix, and instead html-escape the
targets just before rendering them in the GraphDataWindow Ext.ListView

-----

TODO:
  - [x] ensure works with python-3 (`str(thing)` is never the right way :)
  - [x] check what the "dashboard" interface does (it doesn't have this problem, does it allow html injection though?)

### DIFF
--- a/webapp/content/js/composer_widgets.js
+++ b/webapp/content/js/composer_widgets.js
@@ -515,7 +515,14 @@ var GraphDataWindow = {
       hideHeaders: true,
       width: 385,
       height: 140,
-      columns: [ {header: 'Graph Targets', width: 1.0, dataIndex: 'value'} ],
+      columns: [
+        {
+          header: 'Graph Targets',
+          width: 1.0,
+          dataIndex: 'value',
+          tpl: '{value:htmlEncode}'
+        }
+      ],
       listeners: {
         contextmenu: this.targetContextMenu,
         afterrender: this.targetChanged,
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -1915,6 +1915,7 @@ function graphClicked(graphView, graphIndex, element, evt) {
           header: 'Target',
           dataIndex: 'target',
           width: gridWidth - 90,
+          renderer: 'htmlEncode',
           editor: {xtype: 'textfield'}
         },
         {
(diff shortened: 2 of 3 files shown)


ploxiln:
This is done and tested with both python-2.7 and python-3.7

I tested by pasting `<svg/onload=alert()>` (thanks #1662) at the end of graph target expressions in these graph data lists:

Screen Shot 2020-04-20 at 6 06 13 PM

Screen Shot 2020-04-20 at 6 05 31 PM

(and I also tested target expressions with double quotes like `summarize(..., "sum")`)

Before this branch, pasting the svg-alert expression in either place and hitting enter to finish the edit would immediately cause the popup. In the composer, saving and loading the graph and viewing the data targets would no longer cause the popup, but would mess-up the legit double-quotes in the expression, breaking the graph. In the dashboard, saving and loading the dashboard and viewing the data targets would cause the XSS popup (but expressions with double-quotes worked correctly).

After this branch, pasting the svg-alert expression and hitting enter to finish the edit no longer causes popups, saving/loading/viewing does not cause popups in either place, and graphs with quotes work correctly.

Figuring out the ancient extjs parts were admittedly tricky.
  * The function doing the escaping for both the composer and the dashboard can be found here: https://docs.sencha.com/extjs/3.4.0/#!/api/Ext.util.Format
  * For the composer ListView I found an example here showing how to use these functions in templates: https://docs.sencha.com/extjs/3.4.0/#!/api/Ext.list.ListView
  * For the dashboard the same 'tpl' property didn't work, but I followed the docs for the `ColumnModel` `Column` `renderer` to here: https://docs.sencha.com/extjs/3.4.0/#!/api/Ext.grid.Column

ploxiln:
please take a look @deniszh

deniszh:
Wow, @ploxiln , missed that somehow, sorry! Look good, awesome job!

deniszh:
If @cbowman0 or others have no objections I'll merge it soon

ploxiln:
Thanks :)

PR #2591, opened 2020-04-22
[Settings] Allow pickle protocol to be configurable for carbonlink requests.

alikhtag:
TL/DR: Allow to set pickle protocol in case in future any non python `carbon` cannot parse newer highest protocols in higher Python versions. 

**Description of Issue**
When installing `graphite-web` it will use highest available pickle protocol `pickle.dumps(request, protocol=-1)` in the Python version, which is then used for sending `carbonlink` requests to `Carbon`. 
This works well for any Python based implementation of `Carbon` considering if they will both use similar Python version. 

But when using non Python, e.g. [go-carbon](https://github.com/lomik/go-carbon), which was implemented during/for Python 2.7 `graphite-web`.  It is parsing highest available protocol of `2` in Python 2.7 . Consequently `carbonlink` requests are unparsable by `go-carbon` , if Python 3.0+ is used for `graphite-web`. 

 **Benefit** 
Allows user set their own pickle protocol for `carbonlink` requests, so that they are not fixed to certain Python versions, when using non python Carbon that only support certain pickle protocols. 

**To Reproduce**
Set up and run [go-carbon](https://github.com/lomik/go-carbon) with `carbonlink` enabled and configured. 
Run `graphite-web` in Python 3.0+ (3.7) and try to request data from `go-carbon` `carbonlink`. 
`go-carbon` will throw `warnings` due to higher pickle protocol being used, thus it does not parse it correctly. 

 - OS flavor: Linux Ubuntu 18.04
 - Graphite-web version 1.1.6 
 - Django/Python version: Python 3.7
 - Setup type: running in docker, installed using `pip`
 - Carbon: [go-carbon 0.14](https://github.com/lomik/go-carbon)

SIDENOTE: Python 2.x to 3.x change from ASCII to Unicode has also affected pickle parsing in `go-carbon`, but still considering `graphite-web` is sending protocol 4 it would not have been able to parse it,  I've also Opened [PR in go-carbon](https://github.com/lomik/go-carbon/pull/340) to be able to parse Unicode pickle up to protocol 5. But still would be nice to select desired protocol from `graphite-web`


piotr1212:
We should probably make this configurable in Python Carbon as well

PR #2604, opened 2020-05-16
Revert symlink

piotr1212:
see https://github.com/graphite-project/graphite-web/pull/2326


deniszh:
Thanks for spending time on that, @piotr1212 !

PR #2605, opened 2020-05-17
List Promitor as a collector tool

tomkerkhove:
Relates to https://github.com/tomkerkhove/promitor/issues/938


deniszh:
Ok, it's a bit stretched here, but fine, I think.

tomkerkhove:
Thanks for merging!

Just wondering, what do you mean with a bit stretched? The text or? Otherwise I'm happy to adapt.

deniszh:
No worries, I meant that it has support of statsd and not Graphite directly. But I'm just nitpicking. :P

tomkerkhove:
Oh you have a point - Sorry!

Issue #2609, opened 2020-05-21
/functions endpoint returns invalid JSON


shanson7:
**Describe the bug**
Upgrading from a rather old version of graphite to the latest provided by [graphite-mt](https://github.com/raintank/graphite-mt) (graphite version `7ebde4c60`).

Grafana started to fail to display certain functions like `movingWindow`. I think it's due to `Infinity` being returned as a number. This parses fine in python but a lot of other tools fail.

```
$ curl -s http://graphite-stage-nj.internal.com/functions | json_pp
malformed JSON string, neither array, object, number, string or atom, at character offset 52236 (before "Infinity, "type": "i...")
```

That position seems to be `{"default": Infinity,` for the interpolate function. `Infinity` shows up for keepLastValue as well.

**To Reproduce**
Try to parse the json returned from `/functions`

**Expected behavior**
Should parse as JSON

ploxiln:
Hmm ... `Infinity` is accepted by the python json parser, and the c library json-c, and `Infinity` is valid  javascript ... but indeed, `Infinity` does not seem to be accepted nodejs or firefox using `JSON.parse()`, and it is technically not in the standard https://tools.ietf.org/html/rfc7159#section-6

Well, that's awkward.

DanCech:
We do some "special" handling in the render output json to deal with that:
https://github.com/graphite-project/graphite-web/blob/master/webapp/graphite/render/views.py#L242
```
  output = json.dumps(series_data, indent=(2 if requestOptions.get('pretty') else None)).replace('None,', 'null,').replace('NaN,', 'null,').replace('Infinity,', '1e9999,')
```

schen1152:
hello, any updates/eta on the fix for this issue?

deniszh:
Could someone check is this issue still relevant for latest Grafana and
latest Graphite versions? I’m currently AFK.
If still - maybe it can be fixed separately, without global rework of json
parser?

ploxiln:
The root of the problem, still exists, is that graphite-web sometimes includes Infinity or Nan in data structures which it serializes to json (and the python stdlib json will allow that), but javascript `JSON.parse()` will reject that json, both in browsers and nodejs.

I have a somewhat hacky fix idea, which I didn't have time to test yet, at https://github.com/graphite-project/graphite-web/pull/2612 (and I'm abandoning #2610 as not critical and not the best way).

PR #2610, opened 2020-05-21
more robust NaN/Infinite replacement for renderViewJson

ploxiln:
no longer possible to accidentally replace coincidental text
in the middle of string values, though that was already very
unlikely in this context

Inspired by #2609 although not actually fixing that yet. It was pointed out how the fix was done already for `renderViewJson()`, where that existing fix probably works perfectly well, but the implementation seems not quite generically robust enough for other uses. So I wanted to offer up an alternative idea of how to do this replacement.

This still needs some tests, in fact I have not run it at all, just threw it up here real quick for now ...

### DIFF
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -269,7 +269,7 @@ def renderViewJson(requestOptions, data):
 
   replace_nan(series_data)
   output = json.dumps(series_data, indent=(2 if requestOptions.get('pretty') else None))
-  output = re.sub(r'"(-?)%s"' % inf_placeholder, r'\1' + '1e9999', output)
+  output = re.sub(r'"(-?)%s"' % inf_placeholder, r'\g<1>1e9999', output)
 
   if 'jsonp' in requestOptions:
     response = HttpResponse(
--- a/webapp/graphite/render/views.py
+++ b/webapp/graphite/render/views.py
@@ -203,6 +203,7 @@ def renderViewCsv(requestOptions, data):
 
 inf_placeholder = "Infinity-vsEGZrnIyKLHfteA"
 
+
 def replace_nan(obj):
     "replace NaN and Infinity which are non-standard and not accepted by browsers"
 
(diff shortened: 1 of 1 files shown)


DanCech:
We definitely need to do some benchmarking on this to make sure that it doesn't have any nasty impact on rendering times.  It may make more sense to implement this in a variant of `Series.datapoints()` (which could also handle the NoNullPoints option) to avoid needing to iterate the series multiple times.

ploxiln:
That's a good point, this iteration will not be cheap for long sequences of datapoints ... but I do wonder how it compares to `.replace(...).replace(...).replace(...)`

piotr1212:
Looks like orjson implements it correctly and claims to be fast:
```
In [4]: orjson.dumps({'a': float('NaN')})        
Out[4]: b'{"a":null}'
```

ploxiln:
That's an interesting option ... but it requires python-3.6 or later

ploxiln:
... also it requires rust to build, so if using a musl-libc based distro or container like Alpine Linux which does not support manylinux wheel packages, that's a gigantic dependency.

But I do like the idea. I took a look at good old ujson, it has some new releases just recently, and it is still implemented in C. It will reject NaN and Infinity with "OverflowError" exceptions, but there's no way to hook in and replace the values ...

piotr1212:
They also replace infinity with null, not some high number. Not sure what is desired, I think infinity was abused a lot to encode events in timeseries.

Yes and I now see it is rust. It installed without issues here because they ship prebuilt wheels for a limited number of architectures. Also no support for pypy.

I run into the same issue with ujson.

Don't understand why there is no good solution for this, there must be a lot of people having the same issue.

piotr1212:
Maybe simplejson?

```
In [11]: simplejson.dumps({'a': float('Inf')}, allow_nan=True, ignore_nan=True)                    
Out[11]: '{"a": null}'
```

DanCech:
now you see how we ended up with the current solution, which is ugly but works well ;)

ploxiln:
The current solution does not work for values which occur last in an array or object. It does seem to work well enough for this use case, but it turns out there is another endpoint that could use this treatment, `/functions`, which is more likely to have these as "last value" and more likely to have them appear in strings (descriptions).

(I will admit that graphite should probably also get fixes elsewhere to express infinity differently in this output, or avoid needing to, as it is possible `null` will not be correct in some places. But json always parsing successfully seems like an important first step.)

simplejson looks good for this purpose. Widely compatible, and `replace_nan` will replace NaN and Infinity with null. But it doesn't replace Infinity with 1e9999, which does seem to work to smuggle Infinity into browsers, nodejs, and python. Hmm...

PR #2612, opened 2020-05-28
functions view: replace Infinity in json response

### LINKED ISSUE #2609: /functions endpoint returns invalid JSON

ploxiln:
Infinity is not valid json for browsers or nodejs, but 1e9999 seems to work (already used by render endpoint)

fixes #2609

(in the same hacky way as the render endpoint handles this issue, which is not 100% robust, but may be close enough, see #2610)

### DIFF
--- a/webapp/graphite/functions/views.py
+++ b/webapp/graphite/functions/views.py
@@ -1,14 +1,20 @@
+import json
+
 from graphite.util import jsonResponse, HttpResponse, HttpError
 from graphite.functions import SeriesFunctions, SeriesFunction, PieFunctions, PieFunction, functionInfo
 
 
-def jsonEncoder(obj):
-    if hasattr(obj, 'toJSON'):
-        return obj.toJSON()
-    return obj.__dict__
+class jsonInfinityEncoder(json.JSONEncoder):
+    def encode(self, o):
+        return super(jsonInfinityEncoder, self).encode(o).replace('Infinity,', '1e9999,')
+
+    def default(self, o):
+        if hasattr(o, 'toJSON'):
+            return o.toJSON()
+        return o.__dict__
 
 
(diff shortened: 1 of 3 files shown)


ploxiln:
... I'd like to keep this open, as a plausible stop-gap fix

deniszh:
Oh, I missed that somehow. Thanks for addressing this, @ploxiln !
LGTM. Should help with #2609 and  similar bugs.

deniszh:
Could you please review it, @DanCech @piotr1212 ?

ploxiln:
just give me ~ 8 hours to actually run this and sanity-check that it does fix #2609

ploxiln:
OK, I've confirmed that nodejs can parse the the `/functions` response with this branch, and fails with master.

ploxiln:
(also sanity-tested some dashboard graphs and composer graphs and the function-select menus, all seems OK ...)

deniszh:
OK, merging then. Thanks for adressing this, @ploxiln !

Issue #2520, opened 2019-12-05
[BUG] XSS vulnerability over dashboarding save as API request


ytturi:
**Describe the bug**
It's possible to execute JS on application context by modifying the API query values when saving a dashboard.

Even if you do not use the dashboard features, they are enabled by default in Graphite-web.
If you have it as datasource in your grafana dashboards, then you can trace it to Graphite-web and perform malicious actions.

**To Reproduce**
Access to a new dashboard in graphite-web instance (i.e. http://localhost/dashboard). You don't really need data in it.
Use the "save as" feature (In the context menu Dashboard > Save As). Give it a name. 

I have detected two ways to perform this. 

- One which is an all-user available is to introduce JS to the dashboard name when saving it. (i.e. name it "`loremipsum<img src=x onerror=alert(document.domain)>`")
- The more refined way would be introducing the JS to the "state" parameter in the HTTP request payload to save the dashboard as I explain below:

Capture the request sent to send the save action to Graphite.

You can then alter the payload sent to Graphite in the POST request and send the request again with malicious JS.

_Normal payload_
```
state=%7B%22name%22%3A%22loremipsum%22%2C%22timeConfig...
```

_Altered payload_
```
state=%7B%22name%22%3A%22loremipsum<img src=x onerror=alert(document.domain)>%22%2C%22timeConfig...
```

> I placed it with the dashboard name, but you can actually place it wherever you want where a value is parsed (not between escaped characters).

**Expected behavior**
Ignore (or return an error) on XSS attempting requests to Graphite-web.
OR escape/remove those characters from the query.
This can be solved by removing or ignoring requests containing the characters "<" ">" and/or other escaping/scripting characters. -> Sanitize the value before using it.

**Screenshots**

**Environment (please complete the following information):**
 - OS flavor: N/A but confirmed on Ubuntu, RedHat, CentOS
 - Graphite-web version: 0.9.X (since I did not see any issue mentioning this, I assume it's not fixed)
 - Django/Python version: N/A but confirmed on 1.08-1.11/2.7, 2.1/3.6
 - Setup type: N/A but confirmed with pip, sources and OS package (apt)

**Additional context**
This issue was found by a colleague of mine. I've confirmed it by running the same actions over a docker image with Graphite-web.

PR #2620, opened 2020-06-14
Prevent xss

StephenDsouza90:
This PR is in reference to the issue https://github.com/graphite-project/graphite-web/issues/2520 for preventing an Cross-site scripting (XSS) attack when saving the dashboard.

It uses the Django's `django.utils.html.strip_tags()` function to sanitize html input elements, in this case, the name of the dashboard, to prevent the XSS attack.

Currently my approach is to apply the sanitization on the fields that are inputted from the html template and rendered on the graphite web page compared to doing it via a middleware that would inspect `HttpRequest` objects - this IMO would mostly likely be a little inefficient. Sanitizing on the fields that require it is more faster.

The Dashboard save functionality has a possible security issue allowing for an XSS attack, this PR fixes it. In the future, the sanitize utility function can be used for other input fields that require it.

### DIFF
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -3015,8 +3015,8 @@ function setDashboardName(name) {
 
     document.title = name + ' - Graphite Dashboard';
     changeHash(name);
-    navBar.setTitle(htmlEncode(name) + ' - (' + dashboardURL + ')');
-    saveButton.setText('Save "' + htmlEncode(name) + '"');
+    navBar.setTitle(htmlEncode(name + ' - (' + dashboardURL + ')'));
+    saveButton.setText(htmlEncode('Save "' + name + '"'));
     saveButton.enable();
   }
 }
--- a/webapp/content/js/dashboard.js
+++ b/webapp/content/js/dashboard.js
@@ -3015,8 +3015,8 @@ function setDashboardName(name) {
 
     document.title = name + ' - Graphite Dashboard';
     changeHash(name);
-    navBar.setTitle(name + ' - (' + dashboardURL + ')');
-    saveButton.setText('Save "' + name + '"');
+    navBar.setTitle(htmlEncode(name) + ' - (' + dashboardURL + ')');
+    saveButton.setText('Save "' + htmlEncode(name) + '"');
     saveButton.enable();
   }
 }
@@ -3153,7 +3153,7 @@ function showDashboardFinder() {
 
   dashboardsList = new Ext.list.ListView({
     columns: [
-      {header: 'Dashboard', width: 1.0, dataIndex: 'name', sortable: false}
+      {header: 'Dashboard', width: 1.0, dataIndex: 'name', sortable: false, tpl:'{name:htmlEncode}'}
     ],
     columnSort: false,
     emptyText: 'No dashboards found',
--- a/webapp/graphite/dashboard/views.py
+++ b/webapp/graphite/dashboard/views.py
@@ -14,7 +14,7 @@ from graphite.compat import HttpResponse
 from graphite.dashboard.models import Dashboard, Template
 from graphite.dashboard.send_graph import send_graph_email
 from graphite.render.views import renderView
-from graphite.util import json
+from graphite.util import json, sanitize
 from graphite.user_util import isAuthenticated
 
 fieldRegex = re.compile(r'<([^>]+)>')
@@ -224,9 +224,15 @@ def getPermissions(user):
 def save(request, name):
   if 'change' not in getPermissions(request.user):
     return json_response( dict(error="Must be logged in with appropriate permissions to save") )
-  # Deserialize and reserialize as a validation step
-  state = str( json.dumps( json.loads( request.POST['state'] ) ) )
 
+  # Deserialize and reserialize as a validation step
+  state = json.loads(request.POST['state'])
+  if state.get("name"):
+    state["name"] = sanitize(state["name"])
+  state = str(json.dumps(state))
+
+  name = sanitize(name)
+  
   try:
     dashboard = Dashboard.objects.get(name=name)
   except Dashboard.DoesNotExist:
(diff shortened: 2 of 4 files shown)


deniszh:
Cool! Thanks a lot, @StephenDsouza90 , for fixing this.
Ignore DeepSource, looks like it went crazy again, analyzing whole file and not your changes.

DanCech:
The problem isn't in the dashboard save per-se, it is caused by the dashboard name being displayed in an unsafe manner.  Stripping potential "bad" sequences from the dashboard name isn't a bad idea, but to properly fix the XSS we'll also need to make sure that we html-escape the name everywhere it's displayed.

StephenDsouza90:
> The problem isn't in the dashboard save per-se, it is caused by the dashboard name being displayed in an unsafe manner. Stripping potential "bad" sequences from the dashboard name isn't a bad idea, but to properly fix the XSS we'll also need to make sure that we html-escape the name everywhere it's displayed.

Hi @DanCech thanks for the quick feedback!

I'm still a bit unclear on what more needs fixing, if you can point me in the right direction and guide me.

I understand the problem isn't in the dashboard save functionality, but rather the "name" that's being rendered that allows for an XSS attack, but this is exactly what i've tried to tackle - I strip out potential "bad" sequences in the name and in the dashboard's state before it gets saved to the database. This should render the name everywhere correctly since the name will be fetched from the database which is the source of truth and it's saved in a safe manner.

For an attacker trying to inject XSS into the name will be the only person who will be affected by XSS for that current state and session, unless he/she refreshes the page - for other end-users/clients, it shouldn't affect them as what gets saved into the database will be "safe" and that's what will be rendered to them.

So still wondering if this solution is good enough? Or maybe if you could give me a scenario where this attack can still be exploited, I could try to work around that.

StephenDsouza90:
Hi,

I'm still awaiting feedback and next steps on how to move forward with this PR based on my last comment. Would be great if one of you could point me in the right direction.

DanCech:
Basically whenever we display the name it should be html-encoded so that any "special" characters are encoded as html entities rather than interpreted as html.

The docs for the strip_tags function explicitly state that it does not produce strings that are html-safe https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.html.strip_tags, and that html entities should be escaped when displaying the string.

StephenDsouza90:
Hi @DanCech 

Thanks again for the feedback, you're right, my bad! 

I just checked there are two ways to solve this. We can go with [Mozilla Bleach](https://github.com/mozilla/bleach) that does it in a clean way:

```python

>>> import bleach

>>> bleach.clean('an <script>evil()</script> example')
u'an &lt;script&gt;evil()&lt;/script&gt; example'
```

Or we can go with Django's [escape()](https://docs.djangoproject.com/en/3.0/ref/utils/#django.utils.html.escape) function:

> Returns the given text with ampersands, quotes and angle brackets encoded for use in HTML. The input is first coerced to a string and the output has mark_safe() applied.

```python

>>> from django.utils.html import escape

>>> escape('an <script>evil()</script> example')
'an &lt;script&gt;evil()&lt;/script&gt; example'
```

With bleach, the only disadvantage is that we'll be adding yet another library, which might not be a good thing.

I'll go ahead with the escape() for now unless there is any objection.

DanCech:
The escaping needs to be done when we're actually inserting the name into the html on the client side, so you'd want to use the `htmlEncode` function in dashboard.js to do that for dynamic content and a python library when we're building html server-side.

ploxiln:
a possibly helpful example of escaping in the wrong place vs escaping in the right place: https://github.com/graphite-project/graphite-web/pull/2587

StephenDsouza90:
Thanks @ploxiln and @DanCech for pointing me in the right direction! Appreciate it. I'll be working on this later today. More commit fixes coming soon in the PR!

StephenDsouza90:
> The escaping needs to be done when we're actually inserting the name into the html on the client side, so you'd want to use the `htmlEncode` function in dashboard.js to do that for dynamic content and a python library when we're building html server-side.

If I understand @DanCech correctly, would the following be a good enough solution? i.e. to `htmlEncode` the dashboard name at the time of input in dashboard.js just before it gets saved in the backend - So what gets saved is what gets rendered? Something like this:

```javascript
function saveDashboard() {
       <...>
       <...>
       if (button == 'ok') {
         // html-encode dashboard name from input
         text = htmlEncode(text);

         // hmtl-encode dashboard name
         setDashboardName(text);

         // save html-encoded dashboard name
         sendSaveRequest(text);
       }
       <...>
       <...>
}
```

https://github.com/graphite-project/graphite-web/blob/0ed7dd7500521fd3ef7acd011d615dbccec3ad2a/webapp/content/js/dashboard.js#L2723-L2737

DanCech:
You don't want to store the html-encoded dashboard name in the database, because then when you load it and html-encode it again you'll mangle it.  eg if you have a dashboard name like `Testing > TestDash` it'll html-encode to `Testing &gt; TestDash` and that will be rendered correctly.  If you store the encoded string in the database then next time you load that dashboard it'll be html-encoded as `Testing &amp;gt; TestDash` (because `&` html-encodes to `&amp;` and will literally display in the browser as `Testing &gt; TestDash`.  You want to store the actual name in the database and just html-encode it at the point it's inserted into an html document (either on the server side or in the client).

webapp/content/js/dashboard.js
```diff
@@ -3015,8 +3015,8 @@ function setDashboardName(name) {
 
     document.title = name + ' - Graphite Dashboard';
     changeHash(name);
-    navBar.setTitle(name + ' - (' + dashboardURL + ')');
-    saveButton.setText('Save "' + name + '"');
+    navBar.setTitle(htmlEncode(name) + ' - (' + dashboardURL + ')');
+    saveButton.setText('Save "' + htmlEncode(name) + '"');
```
DanCech, line 3019:
```suggestion
    navBar.setTitle(htmlEncode(name + ' - (' + dashboardURL + ')'));
    saveButton.setText(htmlEncode('Save "' + name + '"'));
```

StephenDsouza90:
Thanks @DanCech and @ploxiln for pointing me in the right direction. Really appreciate it and thanks for answering my questions to get a better understanding of what solution is required.

I added a commit for the fix, let me know if it's inline to the solution you had in mind.

PR #2634, opened 2020-09-07
Fix weighted average unequal series count

fkaleo:


deniszh:
Why closed, @fkaleo ?

Issue #2641, opened 2020-10-18
[BUG] metrics with utf8 strings in path are not displayed


mcikosos:
**Describe the bug**
metrics with utf8 strings in path are not displayed in graphs

**To Reproduce**
just send any metric with national chars in it

ex. 
echo "local.random.gówno 5 `date +%s`" | nc ${SERVER} ${PORT}

**Expected behavior**
metrics show in graphs

**Screenshots**
no point

**Environment (please complete the following information):**
 - OS flavor: Ubuntu
 - Graphite-web version: latest docker - graphiteapp/graphite-statsd
 - Django/Python version: latest docker - graphiteapp/graphite-statsd
 - Setup type latest docker - graphiteapp/graphite-statsd

mcikosos:
When I change metric path to local.random.g?wno - the data are shown

deniszh:
Hi @mcikosos ,

IIRC UTF-8 characters can work in tags names/values (because tags are hashed internally), but UTF-8 support in metrics itself is depending on your OS / filesystem setup.

piotr1212:
I found a few issues with unicode chars in Graphite-web, carbon on the other hand does work.

This fixes two https://github.com/piotr1212/graphite-web/commit/17e23efecb6cb40059264fabd3ea36718065176e but I never got around of making a PR. I think there was also one in the carbonlink code. And last, the images rendering code cannot handle unicode (metricnames in the legenda). The last one is the hardest to fix but probably the least important one. There is probably some more, potential places to look are `str()` calls

mcikosos:
> IIRC UTF-8 characters can work in tags names/values (because tags are hashed internally), but UTF-8 support in metrics itself is depending on your OS / filesystem setup. I mean, I'm not saying that's not a bug - and if someone can solve it in generic and backward compatible way - I would be very glad.

Well I thought it is a BUG because carbon/whisper etc. works ok but if it was not supported from the start I can change it to "feature" ;)

mcikosos:
@piotr1212 can You make a PR? So some day it will be merged to master.

piotr1212:
Sorry, I don't expect to anytime soon.

deniszh:
Oh, cool, I didn't know that, @piotr1212 !
PR is welcome ofc, I will pick this up if nobody else.

deniszh:
@mcikosos : Fix is merged in master. `UTF8_METRICS` should be enabled in config (disabled by default because of possible negative  performance consequences).

[...]

PR #2643, opened 2020-10-19
Backport unicode fix from piotr1212@17e23ef 

deniszh:
Backport unicode fix from https://github.com/piotr1212/graphite-web/c…ommit/17e23efecb6cb40059264fabd3ea36718065176e
Fixing #2641

### DIFF
--- a/webapp/graphite/render/grammar_unsafe.py
+++ b/webapp/graphite/render/grammar_unsafe.py
@@ -5,18 +5,7 @@ from pyparsing import (
     __version__, Suppress, Empty
 )
 
-import sys
 from django.conf import settings
-
-try:
-    unichr
-except NameError:
-    unichr = chr
-
-try:
-    xrange
-except NameError:
-    xrange = range
 
 grammar = Forward()
 
--- a/webapp/graphite/render/grammar_unsafe.py
+++ b/webapp/graphite/render/grammar_unsafe.py
@@ -1,7 +1,7 @@
 from pyparsing import (
     Forward, Combine, Optional, Word, Literal, CaselessKeyword,
     CaselessLiteral, Group, FollowedBy, LineEnd, OneOrMore, ZeroOrMore,
-    alphas, alphanums, printables, delimitedList, quotedString, Regex,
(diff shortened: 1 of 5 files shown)


webapp/graphite/render/grammar_unsafe.py
```diff
@@ -93,7 +105,8 @@ def setRaw(s, loc, toks):
 ).setParseAction(setRaw)('call')
 
 # Metric pattern (aka. pathExpression)
-validMetricChars = ''.join((set(printables) - set(symbols)))
+unicodePrintables = u''.join(unichr(c) for c in xrange(sys.maxunicode) if not unichr(c).isspace())
```
DanCech, line 108: 2 concerns here:
1. This is a huge range
2. Should we be excluding more characters than just spaces, there are many unicode categories https://en.wikipedia.org/wiki/Unicode_character_property#General_Category

deniszh: 1. Yes, I missed that point, thanks! I'll probably put it behind switch then - if user want UTF-8 in metrics, it can enable that. Would it be better?
2. I think only spaces is good for start, not sure should we add anything later

piotr1212: Agree with Dan, this is basically the reason why I didn't create the PR back then, I didn't think this was an elegant solution and didn't know how to do it properly.

deniszh: Oh, TIL pyparsing 2.3.0 and newer giving you `pyparsing_unicode.printables` !
But it's better to hide it behind switch anyway

piotr1212: Ah `pyparsing_unicode.printables` is much better! Still not sure if this is the right way to do this but as long it is behind the switch it is good enough for me.

piotr1212 (approved):

deniszh:
Still need to test is it really fix #2641

deniszh:
Yes, it's helping with #2641, if UTF8_METRICS is enabled ofc. Merging.

PR #2649, opened 2020-11-05
Update whisper.rst with Python 3 fix

cdeil:


piotr1212 (approved):

cdeil:
Why isn't https://graphite.readthedocs.io/en/latest/whisper.html updating?

There's no active docs build for latest?
https://readthedocs.org/projects/graphite/builds/

PR #2654, opened 2020-12-07
fix expand_braces

unitmatrix:
Fixes https://github.com/grafana/grafana/issues/5936 and https://github.com/grafana/grafana/issues/11274

graphite web v0.9.16 has a bug where curly braces {} in metrics name do not expand correctly. This update fixes it.

### DIFF
--- a/webapp/graphite/storage.py
+++ b/webapp/graphite/storage.py
@@ -26,7 +26,7 @@ except ImportError:
 
 
 DATASOURCE_DELIMETER = '::RRD_DATASOURCE::'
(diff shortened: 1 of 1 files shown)


DanCech:
Interesting, if this does solve the problem it should be applied to master as well since that uses the same expand_braces function (just moved to `finders/__init__.py`)

deniszh:
Hi @unitmatrix ,

Thanks for your contribution, but you created patch for unmainteained `0.9.x` branch. Could you please do the same for master branch instead. As @DanCech stated code was moved to `finders/__init__.py`

deniszh:
Closed ion favor of #2655

Issue #2154, opened 2017-12-14
0.9.16 brace expansion bug


mlausch:
Metrics with braces like at the following will not "expand" anymore.
foo.{bar}.baz
queries like  foo.{bar1,bar2}.baz  works as intented.

Until Version 0.9.15 The braces around a value without commas are removed and all was fine. But now only if in the braces are two or more comma spparated values this will work correctly.

Since grafana sends queries with single values in braces this functionality is neccessary to work. 

I figured out, that the faulty function is in graphite/storage.py  (function "expand_braces")
I had a look in the 1.0.x branch. There is the functionality moved to graphite/finders/__init__.py In this branch the described bug is fixed.

Is it possible to backport this function to 0.9.x?  
I cannot use Version 1.0.x in my environment caused by the django dependencies.

DanCech:
See #2156

deniszh:
Hello @mlausch,
Please note that we're not going to make any 0.9.x releases. But you can use 0.9.x branch, of course, after merging fix in #2156

deniszh:
@mlausch: Fix is merged, please test.

mlausch:
@deniszh: The Fix works for me.

PR #2655, opened 2020-12-08
fix expand_braces

unitmatrix:
Fixes https://github.com/grafana/grafana/issues/5936, https://github.com/grafana/grafana/issues/11274, and https://github.com/graphite-project/graphite-web/issues/2154

graphite web has a bug where curly braces {} in metrics names do not expand correctly. This update fixes it.

### DIFF
--- a/webapp/graphite/finders/__init__.py
+++ b/webapp/graphite/finders/__init__.py
@@ -2,7 +2,7 @@ import fnmatch
 import os.path
 import re
 
-EXPAND_BRACES_RE = re.compile(r'(\{([^\}]*[^\\])\})')
+EXPAND_BRACES_RE = re.compile(r'(\{([^\{\}]*)\})')
 
 
 def get_real_metric_path(absolute_path, metric_path):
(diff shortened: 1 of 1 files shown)


deniszh:
Hi @unitmatrix,

I think fix is not correct :(
Tests are broken now -
https://github.com/graphite-project/graphite-web/blob/fac5a1d98ceb2ee099556169877a801fb0107d01/webapp/tests/test_finders.py#L191-L194
and
https://github.com/graphite-project/graphite-web/blob/fac5a1d98ceb2ee099556169877a801fb0107d01/webapp/tests/test_finders.py#L241-L244

It's also would be great if you add tests for that fix too.

Thanks!

unitmatrix:
Thanks for looking. Here are some improvements.

deniszh:
Thans for looking into this, @unitmatrix ! 
LGTM! One test broken but that's a flacky one.
Could someone else also review please?

deniszh (approved):

DanCech (approved):
This looks good to me, it took a minute to understand that it now works by expanding the braces depth-first where in the past it always processed them in order, but it makes things much simpler this way.  This code is relatively well exercised by the finder tests, but I do agree it would be great to have some direct tests for expand_braces since the old code did also pass those tests.

ploxiln:
Looks good to me. (And agree it would be nice to make a test out of an example that wasn't processed correctly before.)

deniszh:
Closed in favor of https://github.com/graphite-project/graphite-web/pull/2661

PR #2658, opened 2020-12-22
return empty result list for empty series

unitmatrix:
Backport from master. Fixes https://github.com/graphite-project/graphite-web/issues/788. 
Return empty result list for empty series in movingMedian and movingAverage functions.


DanCech (approved):
Seems reasonable

PR #2659, opened 2020-12-22
Optionally resolve right hand dip in sums because of lack of current minute in caches

cbowman0:
When not all data has been loaded in yet for the current minute, the right hand side of a sum can dip which causes some observers heartburn.

Optionally allow a feature that will trim the most recent datapoints if not all of them exist within aggregate().


deniszh:
Well. Not really optimal, but doing its job. At least we have choice how to handle last point now. Thoughts? /cc @DanCech @piotr1212 @ploxiln

deniszh (approved):

PR #2671, opened 2021-01-22
Fix paths not matching description

thedoc31:
Paths in commands specified /var/lib but description specified /usr/share. Makes the command consistent with the description, unless I'm missing something special about /var/lib and the installation procedure.


deniszh:
Looks like typo. Thanks!

PR #2673, opened 2021-01-24
Migrating tests to Github Actions

deniszh:
Will keep Travis for now. I saw flapping test ig GHA before, but now they're working...


deniszh:
Merging, but keeping Travis for now - I'm still not sure on tests stability on GHA and in general.

PR #2693, opened 2021-04-09
Resolves #2692 can't unpickle Interval/IntervalSet

### LINKED ISSUE #2692: [BUG] pickled Interval and IntervalSet objects returned from python 2 graphite web cannot be unpickled in python3
**Describe the bug**
The pickled data returned from the `/metrics/find` endpoint of graphite-web running on python2 cannot be unpickled by graphite-web running on python3.

**To Reproduce**
While this can be observed by configuring an instance of graphite-web running on python 3 to point to any graphite-web hosts running on python2 as their "cluster servers". The minimal reproducible bug is easily arrived at in the python repl
```
$ python2.7
Python 2.7.16 (default, Aug  7 2019, 19:39:03)
[GCC 4.4.7 20120313 (Red Hat 4.4.7-23)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import graphite.intervals
>>> i = graphite.intervals.Interval(1,10)
>>> s = graphite.intervals.IntervalSet([i])
>>> import pickle
>>> with open("py2.pickle", "wb") as f:
...     pickle.dump(s, f)
...
>>>
$ python3
Python 3.6.8 (default, Nov 16 2020, 16:55:22)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-44)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pickle
>>> with open('py2.pickle', 'rb') as f:
...     contents=pickle.load(f)
...
  File "<stdin>", line 2, in <module>
AttributeError: 'Interval' object has no attribute '__dict__'
```

**Expected behavior**
I would expect that regardless of python version being run on graphite-web hosts that the wire protocols used to between them would be compatible.

 - OS flavor: Centos
 - Graphite-web version 1.1.7

**Additional context**
The cause of this seems to be a difference in how slotted objects are stored in python2 and python3. If you modify the `Interval` and `IntervalSet` class definitions to not use slots the objects unpickle just fine on python3. A side effect of this problem is that graphite-web on python3 is incompatible with other graphite-web implementations (carbonapi for example) which use pickled objects that copy the official graphite-web.

While it seems unlikely that there are many mixed python2/3 graphite-web clusters in the wild there definitely are people using graphite-web as a front end to carbonapi that will be surprised to find that graphite-web running on python3 will not work for them.

drawks:
* removes `__slots__` definition from `Interval` and `IntervalSet`

It took me a minute to figure this one out. So, the original code uses "old style" classes which means that the `__slots__` definition is a noop on python2, but on python3 you automatically get new style classes. Really this should be updated to use "new style" classes, but that would suddenly create compatibility issues between older versions of graphite-web AND also software such as carbonapi which works by mimicking the existing graphite-web behavior. By removing the `__slots__` we get identical behavior on python2 and python3 which is a vanilla unslotted class instance that unpickles fine in both pyhton2 and python3 and works with carbonapi too.

### DIFF
--- a/webapp/graphite/intervals.py
+++ b/webapp/graphite/intervals.py
@@ -3,7 +3,6 @@ NEGATIVE_INFINITY = -INFINITY
 
 
 class IntervalSet:
-    __slots__ = ('intervals', 'size')
 
     def __init__(self, intervals, disjoint=False):
         self.intervals = intervals
@@ -74,7 +73,6 @@ class IntervalSet:
 
 
 class Interval:
-    __slots__ = ('start', 'end', 'tuple', 'size')
 
     def __init__(self, start, end):
         if end - start < 0:


deniszh:
Geez, thanks, @drawks !
LGTM directly reports that as error 🤦

deniszh (approved):

PR #2720, opened 2021-09-28
compressPeriodicGaps function

deniszh:
This PR contains function named `compressPeriodicGaps()` which can help with removing periodic gaps (i.e. `None` values) from series, (recalculating start, stop and step), which could be useful for users with incorrect resolution setup for some metrics. Especially, that can be useful for derivative functions (like `derivative()`, `nonNegativeDerivative()` and `perSecond()`) which are not working properly with series with gaps.
Please check test `test_compressPeriodicGaps` for illustration how `compressPeriodicGaps()` works.
You can use `summarize(..., '<desired interval>', 'last')` function to have same effect (thanks to @DanCech for that suggestion) but `compressPeriodicGaps()` will try to guess proper step by first three non-null values, and it just do nothing with series if guessing failed.

Also, this patch contains support for quite experimantal variable `AUTOCOMPRESS_GAPS_IN_DERIVATIVE_FUNCTIONS` which if set to True will automatically apply compressGaps() to all argument of derivative functions above. It's a hack, but it's really useful for customers who want to have working derivative functions over series with gaps, without changing many hundreds of dashboards.

I also removing pyhash from Python 3.x tests temporarily because of issue in [pyhash](https://github.com/flier/pyfasthash/issues/49) which I couldn't fix for now (pipping setuptools<58 for some reason not working). I'll put it back when pyhash release fixed version.


webapp/tests/test_functions.py
```diff
@@ -5796,6 +5796,34 @@ def test_hitcount_1minute_alignToFrom_false(self):
...
+        seriesList = self._gen_series_list_with_data(
+            key=['collectd.test-db0.load.value', 'collectd.test-db1.load.value', 'collectd.test-db2.load.value', 'collectd.test-db3.load.value'],
+            start=100,
+            end=250,
+            step=10,
+            data=[
+                [None, 1, None, None, 2, None, None, 3, None, None, 4, None, None, 5, None],
+                [None, None, 1, None, None, 2, None, None, 3, None, None, 4, None, None, 5],
+                [None, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, None],
+                [None, 1, None, 3, None, 5, None, 7, None, 9, None, 11, None, 13, None],
```
DanCech, line 5810: Should we add some tests here with less-regular patterns of None values? I'm trying to figure out what will happen if you run this on a series that isn't quite so regular and it seems like you'll end up with values potentially moved to unexpected timestamps.

deniszh: Good point, @DanCech !
Current implementation is indeed very naive - irregularities can cause misalaigments, long series of None completely mangling results.  I added warning to description and added 2 more wrong use case to test.
Should I rename it to  `compressRegularGaps()` or it's sounds more misleading then?
Current implementation works quite OK for specific usecase of regular gaps, not sure if it's reasonable to make it more complex, though...

DanCech: Yeah it's a tough one, especially since the more complex it becomes the harder it is to reason about.

The use-case seems to be very close to `summarize(seriesList, '<desired step>', 'last')`, though you're trying to determine the target step automatically.  That in itself can be problematic though because there's no way to be sure of what it will decide the step should be or that it'll be consistent between different series in the same list.

deniszh: It was tested with similar use case - `divideSeries(nonNegativeDerivative(compressGaps(lowfrq.with.gaps)),hifreq.without.gaps)` producing exact same result as `divideSeries(nonNegativeDerivative(lowfreq.without.gaps),hifreq.without.gaps)`.  Not sure if `summarize(seriesList, '<desired step>', 'last')` will do the same...

DanCech: It should be very similar, since it'll just take the last non-null value in each interval.  The advantage will be that it can handle situations where there is a genuine None for a given interval while compressGaps would end up shifting all subsequent values.  It will also produce a predictable and consistent output step.

deniszh: Yes, `ummarize(seriesList, '<desired step>', 'last')` works same way, I tested that. It has only one downside - you need to provide step manually. I'll probably rework compressGaps - it will autodetect step and just use `def _summarizeValues()` call instead of `_compressGaps()`

DanCech (commented):
This is looking good, added a couple of minor suggestions

DanCech (approved):
LGTM

deniszh:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------:|
|❌|1.1.x|Cherrypick failed because the selected commit was a merge commit. Please try again by specifying the parent with the `mainline` argument:

> backport --mainline

or:

> backport --mainline <parent-number>

Or refer to the git documentation for more information: https://git-scm.com/docs/git-cherry-pick#Documentation/git-cherry-pick.txt---mainlineparent-number|

To backport manually run: `node scripts/backport --pr 2720`.
For more info read the [Backport documentation](https://github.com/sqren/backport#backport)

PR #2724, opened 2021-11-03
maxStep in requestContext for Finder

deniszh:
This PR adding URL parameter named `maxStep` which will be added to requestContext (if defined in URL parameters) and then finder plugin can use it.

Some context and explanation.
----------------------------------
We have requestContext parameter, which used in Graphite-web code. It can be used for transferring some data between functions (i.e. providing context). This context parameter also accepting by Finder plugins, but by default Finder ignores it.
Both changes trying to help fix long standing issue in Graphite, when datapoint period becoming less then windowSize parameter in some functions these functions return empty or wrong result. That can happen with `hitcount()`, `summarize()`, `smartSummarize()`, `movingAverage/Min/Max/Median/Sum/Window()` and `exponentialMovingAverage()`.
Unfortunately, that issue is not possible to fix in case of Whisper - if we have only aggregated data in low resolution archive we just can't get high resolution data from it. But issue can be solvable with storages that (at least potentially) can do live aggregation - then we can or transfer desired step directly (using `maxStep` parameter described above) - or Finder plugin can calculate desired step using target list which is already provided in context starting from v1.1.8.


deniszh:
Rebased, looks better now.

DanCech:
> We can't use normal target parameter for it, so, I creating second one, named targets_serialized.

I'm confused by what we gain by having a pickled version of the same data that's already in targets, can you explain what you mean here?

deniszh:
Easy - it’s not there. I couldn’t trace what removes it and why but target
just not present in requestContext in Finder.

deniszh:
But you probably right, it's worth to at least find why current setup is not working.

deniszh:
@DanCech : yes, looks like I started from different version and just missed @replay 's commit which bringing targets into context. 😵‍💫 
Removed it and amended description.

DanCech (approved):

deniszh:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------:|
|✅|1.1.x|[](https://github.com/graphite-project/graphite-web/pull/2728)|

Issue #2726, opened 2021-11-12
[BUG] AttributeError: 'ParseResults' object has no attribute 'raw'


tekbot:
**Describe the bug**
grammar-unsafe.py is attempting to assign the attribute 'raw' without using indexed notation as required by pyparsing >= 3.0.4

**To Reproduce**
With the versions of graphite-web and pyparsing listed below, attempt to use a basic function in either graphite composer, grafana, the API, etc.

**Expected behavior**
Function is applied, data is rendered

**Environment (please complete the following information):**
 - OS flavor: AWS AL2 (effectively CentOS 7)
 - Graphite-web version 1.1.8
 - Django/Python version 2.1.2/3.7
 - Setup type from sources, libraries pip3 installed, no venv.

**Additional context**
This issue was raised a few days ago [here](https://stackoverflow.com/questions/69886792/graphite-error-parseresults-object-has-no-attribute-raw-when-call-render).  The accepted answer was to downgrade the pyparsing version, which likely won't work for everybody.  Reading the pyparsing release notes [here](https://pyup.io/changelogs/pyparsing/#3.0.4) this was an intended change (though clearly not backwards compatible).  Updating the referenced line (83) in grammar_unsafe.py to use the indexed notation allows me to create and display graphs using aggregate functions.  My concern is that there are other places where the incorrect assignment type is being used.

marcphilipp:
Should this issue be closed given that #2727 has been merged?

marcphilipp:
Is there a timeline for the 1.2.0 release? 😇

deniszh:
1.1.9
Usually we have no timeline, cutting releases when it's enough changes happened.
I'll try to find time to do release this week.

marcphilipp:
@deniszh That would be awesome! I got 1.2.0 from the [changelog](https://github.com/graphite-project/graphite-web/blob/5f35fc2f1567dcc3ae7572e71ea1255bb1406d60/CHANGELOG.md).

deniszh:
1.2.0 is current master, 1.1.x is current prod branch

deniszh:
Fix will be released in 1.1.9 soon

lilydjwg:
Please keep open. A fixed version is not released yet.

marcphilipp:
Was this released as part of 1.1.10?

deniszh:
Yes, it was

marcphilipp:
Thanks for the confirmation! In that case this issue can be closed, right?

[...]

PR #2727, opened 2021-11-13
Fix pyparsing > 3.0 compatibility issue.

parrotpock:
Addresses issue raised in https://github.com/graphite-project/graphite-web/issues/2726 plus one other compatibility issue I came across with a failing piped query test.

I've checked that tests pass but haven't tested this in a live environment. Is it worth adding something to the tox matrix to check multiple pyparsing versions? It's been a while since I've used tox, so I left this part out as I'm not setup to test it properly.

### DIFF
--- a/webapp/graphite/render/evaluator.py
+++ b/webapp/graphite/render/evaluator.py
@@ -58,7 +58,7 @@ def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
     return evaluateTokens(requestContext, tokens.template, arglist)
 
   if tokens.expression:
-    if tokens.expression.pipedCalls:
+    if tokens.expression.pipedCalls.asList():
       # when the expression has piped calls, we pop the right-most call and pass the remaining
       # expression into it via pipedArg, to get the same result as a nested call
       rightMost = tokens.expression.pipedCalls.pop()
--- a/webapp/graphite/render/grammar_unsafe.py
+++ b/webapp/graphite/render/grammar_unsafe.py
@@ -80,7 +80,11 @@ kwargs = delimitedList(kwarg)
 
 
 def setRaw(s, loc, toks):
-    toks[0].raw = s[toks[0].start:toks[0].end]
+    try:
+        toks[0]['raw'] = s[toks[0].start:toks[0].end]
+    except AttributeError:
+        # Accommodate version < 3 of the PyParsing API.
+        toks[0].raw = s[toks[0].start:toks[0].end]
 
 
 call = Group(
--- a/webapp/graphite/render/grammar_unsafe.py
+++ b/webapp/graphite/render/grammar_unsafe.py
@@ -80,11 +80,7 @@ kwargs = delimitedList(kwarg)
 
 
 def setRaw(s, loc, toks):
-    try:
-        toks[0]['raw'] = s[toks[0].start:toks[0].end]
-    except AttributeError:
-        # Accommodate version < 3 of the PyParsing API.
-        toks[0].raw = s[toks[0].start:toks[0].end]
+    toks[0]['raw'] = s[toks[0].start:toks[0].end]
 
 
 call = Group(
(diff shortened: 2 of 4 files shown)


deniszh:
Hi @parrotpock ,

Thanks for your contribution!

> Is it worth adding something to the tox matrix to check multiple pyparsing versions? 

As far as I understand tox it's already checking multiple pyprsing versions implicitly. But maybe I'm wrong.

Please also take note, that currently pyparsing version is pinned, so, your change doesn't do much:
https://github.com/graphite-project/graphite-web/blob/master/tox.ini#L39
https://github.com/graphite-project/graphite-web/blob/master/requirements.txt#L46

You need to relax constraits to proper check your changes.

parrotpock:
Yeah, I only tested this manually with the updated dependencies. Should I drop the pin in both places or just the tox config for now?

deniszh:
Yes, please. Minimal pin has sense, but drop < 3.0.0 part

deniszh:
Tests are green, but looks like I was wrong - now it's testing only latest pyparsing, which could be OK, but including also pyparsing2 tests could be good idea too.

ptmcg:
As the maintainer of pyparsing, I suggest that at minimum you test against pyparsing 2.4.7 and 3.0.6. 2.4.7 has been the stable release since April, 2020, and 3.0.6 is the most recent released version including all regression and compatibility bug-fixes encountered in 3.0.x.

webapp/graphite/render/grammar_unsafe.py
```diff
@@ -80,7 +80,11 @@
 
 
 def setRaw(s, loc, toks):
-    toks[0].raw = s[toks[0].start:toks[0].end]
+    try:
```
ptmcg, line 83: The code you have in the try-block should be compatible with pyparsing pre-3.0 and post-3.0. It's the code in the except block that only works in pyparsing pre 3.0.

I don't think the code in the try block will *ever* raise `AttributeError`, in either version of pyparsing.

webapp/graphite/render/evaluator.py
```diff
@@ -58,7 +58,7 @@ def evaluateTokens(requestContext, tokens, replacements=None, pipedArg=None):
     return evaluateTokens(requestContext, tokens.template, arglist)
 
   if tokens.expression:
-    if tokens.expression.pipedCalls:
```
ptmcg, line 61: Good catch on this! It is a subtle change, and was not in the pyparsing release notes (I have added it).

webapp/graphite/render/grammar_unsafe.py
parrotpock: My bad, you're right. I initially had the try/except blocks the other way around, but swapped them to make the post-3.0 compatible change the default and forgot to update the exception handling. I hadn't checked that they both would work post and pre 3.0, and so I fixed this. Thanks for catching this.

webapp/graphite/render/evaluator.py
parrotpock: Great. Thanks for the confirmation :)

parrotpock:
I added a new target for the suggested pyparsing version and above. Since we removed the upper pin this seems vaguely sensible? 

Thanks for the quick feedback, btw.

deniszh:
Tests are green. Thanks a lot, @ptmcg , for suggestions!
And @parrotpock ofc for fixing that.

DanCech (approved):
Nice!

parrotpock:
No problem @deniszh. Let me know if you need anything else from me before you can merge this.

deniszh:
I think it's good enough to merge. Thanks!

parrotpock:
Thanks @deniszh!

deniszh:
## 💔 Some backports could not be created

| Status | Branch | Result |
|:------:|:------:|:------:|
|❌|1.1.x|Aborted|

To backport manually run: `node scripts/backport --pr 2727`.
For more info read the [Backport documentation](https://github.com/sqren/backport#backport)

PR #2829, opened 2023-12-16
Fix some deprecations

niol:


brd:
These work for me, @deniszh can you look at merging this PR?

PR #2845, opened 2024-08-16
Update carbon-daemons.rst - rewrite-rules

earthgecko:
Explicitly state and relate rewrite-rules.conf to carbon-aggregator in the same way that relay-rules.conf is only applied to carbon-relay making it clearer as to what evaluates and executes rewrite-rules.conf


earthgecko:
Related carbon/conf/rewrite-rules.conf-example update PR - https://github.com/graphite-project/carbon/pull/955

PR #2858, opened 2024-08-25
More upgrades

deniszh:
Some from dependabot, doing that more explicitly


deniszh:
Ok, that would require more work

PR #2873, opened 2025-02-01
Sync Django version with requirements.txt

UweMenges:
This fixes a dependency issue when installing with pip from a tarball. Found using Python 3.13, which removed the 'cgi' module, required for Django 3.2.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,7 +107,7 @@ try:
       package_data={'graphite': ['templates/*', 'local_settings.py.example']},
       scripts=glob('bin/*'),
       data_files=list(webapp_content.items()) + storage_dirs + conf_files + examples,
-      install_requires=['Django>=3.2,<4', 'django-tagging==0.5.0', 'pytz',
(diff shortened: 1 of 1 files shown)


deniszh:
Well, thanks for your contribution but 3.13 is not single supported versions, I'm afraid. 3.8 can be removed, but even 3.9 is still supported by some distros.

deniszh:
Ah, sorry, I changed requirements already. Ok, let me fix tests then. Master would not be compatible with old versions, indeed

deniszh:
@UweMenges : sorry for confusion. Unfortunately, I was forced to include your fix to https://github.com/graphite-project/graphite-web/pull/2874 (because tests using setup.py, d-oh) and I did it without proper attribution of your code. You can try to rebase but I suspect it will be noop now.  I would like to include your change to CHANGELOG anyway.
Thanks!

UweMenges:
Closing pull request as the relevant code change went in with https://github.com/graphite-project/graphite-web/pull/2874.
Feel free to include or omit to CHANGELOG.md what you like. Thanks for caring! :)

PR #2912, opened 2026-03-30
Fix TypeError in _nonNegativeDelta when val is None and maxValue is set

npazosmendez:
`_nonNegativeDelta` compared `val` against `maxValue`/`minValue` before checking whether `val` is `None`, causing a `TypeError` when a series contains `None` values and `maxValue` (or `minValue`) is provided.

Fix: move the `None in (prev, val)` early-return before the comparisons. This also makes semantic sense — on the first reading or a `None` value there is no delta to compute, regardless of bounds.

The added test case reproduces the issue, and fails without the fix:

```python
     ERROR: test_perSecond_nones (tests.test_functions.FunctionsTest.test_perSecond_nones)
       File "/home/nicolas/code/graphite-web/webapp/tests/test_functions.py", line 1802, in test_perSecond_nones
         result = functions.perSecond({}, seriesList, maxValue=1000)
       File "/home/nicolas/code/graphite-web/webapp/graphite/render/functions.py", line 2230, in perSecond
         delta, prev = _nonNegativeDelta(val, prev, maxValue, minValue)
                       ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       File "/home/nicolas/code/graphite-web/webapp/graphite/render/functions.py", line 2431, in _nonNegativeDelta
         if maxValue is not None and val > maxValue:
                                     ^^^^^^^^^^^^^^
     TypeError: '>' not supported between instances of 'NoneType' and 'int'

     Ran 1 test in 0.003s
```

### DIFF
--- a/webapp/graphite/render/functions.py
+++ b/webapp/graphite/render/functions.py
@@ -2427,15 +2427,15 @@ nonNegativeDerivative.params = [
 
 
 def _nonNegativeDelta(val, prev, maxValue, minValue):
+  # first reading or None value
+  if None in (prev, val):
+    return None, val
+
   # ignore values larger than maxValue
   if maxValue is not None and val > maxValue:
     return None, None
   if minValue is not None and val < minValue:
     return None, None
-
-  # first reading
-  if None in (prev, val):
-    return None, val
 
   # counter increased, use the difference
   if val >= prev:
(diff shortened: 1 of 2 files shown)


deniszh:
Not sure about this... it's definitely changing function behaviour - see broken tests.

npazosmendez:
@deniszh sorry I had missed those failures locally. I did spot the behavior change but it seemed harmless to me. I guess it's not. Reworked, behavior should be preserved now. All tests pass locally

